PR #9, opened 2018-01-24
Cavesystem

duncanspumpkin:
I'm not convinced the cave system is actually used.
Thought i should pr my work as it's getting stale and don't want you to waste time reworking


OmniBlade (commented):
Looks like you forgot to pass the xfer pointer to the Call_Method calls in the Xfer_Snapshot functions, do you want to fix that before I merge or are you happy for me to fix it?

duncanspumpkin:
Feel free to fix it. OpenLoco is taking up my free time right now so this is on the back burner

PR #11, opened 2018-02-01
Adding issue template

mrtnptrs:
This can be used for people to make (better) issue reports, especially for in the future. This can stimulate development as it will attract more developers to help with this project. I added it to a hidden .Github folder, so it doesn't clutter anything. What do you think of it @OmniBlade ? Any suggestions?

My only question is where to find the stackdump after Thyme crashes. I couldn't find out where it gets stored. Then I could add it to the issue template too.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,27 @@
+<!--
+  Fill in the placeholders below. Delete any headings and placeholders that you do not fill in. 
+-->
+####Operating System: [e.g. Windows 10]
+####Commit/build: [e.g. 30376e1]
+####Original game source: [e.g. Origin or game disc]
+
+- [ ] Reproducible in the original Generals Zero Hour?
+- [ ] Multiplayer affected?
+- [ ] Are you using any mods? (like Gentool, Rise of the Reds or Contra)
+<!-- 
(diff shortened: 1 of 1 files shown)


OmniBlade:
Currently crash logs are saved to C:\Users\%username%\Documents\Command and Conquer Generals Zero Hour Data\ReleaseCrashInfo.txt which is where the original game saved them. At some point I will probably change this to C:\Users\%username%\AppData\Roaming\Thyme\%gamename%\ReleaseCrashInfo.txt to be consistent with where later sage games saved their data and to avoid conflicts with running the original binary on its own.

There is also ThymeDebugLogFile.txt in the same directory which does logging for reimplemented functions and dumps some info about the build and the players system.

mrtnptrs:
Thank you @OmniBlade Will put that in the template. Do you have any more suggestions?

mrtnptrs:
This also makes [this wiki page](https://github.com/TheAssemblyArmada/Thyme/wiki/What-To-Include) unnecessary/suprfluous. Also, to be honest, during my time managing the CorsixTH bug-hunting team and looking at the OpenRCT2 projects issue reports for a few years, I found out that some people even don't bother to read the issue template sometimes. So, I'm pretty sure that sadly almost no one will read that wiki page. But what do you think? :)

Edit: but if you want to keep that wiki-page, I would be happy to improve it.

PR #23, opened 2018-03-06
Standalone

OmniBlade:
Allows building a standalone executable (not very functional currently) and adds .travis.yml to build standalone linux builds with GCC and Clang to help ensure standalone can be compiled.


.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+env:
+  - 'BUILD_TYPE=Release'
+  - 'BUILD_TYPE=Debug'
+  
+before_script:
+  - mkdir build
+  - cd build
+  - cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DSTANDALONE=TRUE ..
+  
+script: make
```
janisozaur: Travis offers 2 CPUs, so pass `-j2` here.

src/game/client/optionpreferences.cpp
```diff
@@ -490,36 +485,52 @@ bool OptionPreferences::Get_Firewall_Need_Refresh()
...
+#else
+    return 0;
+#endif
 }
 
 void OptionPreferences::Set_Online_IPAddress(AsciiString address)
 {
+#ifndef THYME_STANDALONE
     // TODO needs IPEnumeration
     return Call_Method<void, OptionPreferences, AsciiString>(0x004636F0, this, address);
```
janisozaur, line 526: Wouldn't it make more sense to simply have dummy `Call_Method` when `THYME_STANDALONE` is defined?

.travis.yml
OmniBlade: Thanks, I missed that.

src/game/client/optionpreferences.cpp
OmniBlade: Would I not still need an ifdef for the standalone return on none void functions though?

janisozaur: Only if you return the non-default value.

OmniBlade: Good point. I'd have to rejig the includes as currently I conditionally include the header for Call_Method based on if the build is standalone or not as well.

janisozaur: Up to you. In case of OpenRCT2 it proved useful, as I was able to repurpose such (similar) macro to only use the data section while `.test` was not used at all.

PR #28, opened 2018-04-04
Calculate CRC lookup table at compile time

feliwir:
The table is still identical to the table before


src/game/common/array.h
```diff
@@ -0,0 +1,51 @@
...
+ * @copyright Thyme is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#pragma once
+
+#ifndef ARRAY_H
```
janisozaur, line 17: superfluous in presence of `#pragma once` just above?

feliwir:
@janisozaur
i know, but the other headers have it aswell

feliwir:
Seems to be some style. Maybe @Omniblade can comment on this

OmniBlade:
Originally I wasn't sure what compilers and level of C++ I was intending to support and pragma once isn't actually standard although pretty much all C++11 compilers do support it so I've been hedging my bets as it were and couldn't face changing the style after the fact.

feliwir:
i don't know of any compiler that doesn't have #pragma once these days

OmniBlade:
Superceeded by #30

PR #30, opened 2018-04-04
Calculate fast tables at compiletime

feliwir:
This builds on top of #28


feliwir:
@OmniBlade PR should be ready

PR #42, opened 2018-07-18
Particle System

OmniBlade:
Implements most of the none graphic related parts of the SAGE software particle system.


ghost:
Great progress ... I'm still curious though if you're implementing this code in a way so it's expandable-flexible for future while not affecting the accuracy of replicating the original, possibly avoiding future complete rewrites.

PR #43, opened 2018-07-24
Videoplayer

OmniBlade:
Base classes for video player subsystem.

### DIFF
--- a/src/game/client/videoplayer/videoplayer.cpp
+++ b/src/game/client/videoplayer/videoplayer.cpp
@@ -1,0 +1,181 @@
+/**
+ * @file
+ *
+ * @author OmniBlade
+ *
+ * @brief Virtual interface for video players.
+ *
+ * @copyright Thyme is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#include "videoplayer.h"
+#include "videostream.h"
+#include <cstddef>
+
+#ifndef THYME_STANDALONE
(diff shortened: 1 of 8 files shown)


src/game/client/videoplayer/videoplayer.h
```diff
@@ -0,0 +1,72 @@
...
+    virtual void Init() override;
+    virtual void Reset() override;
+    virtual void Update() override;
+
+    virtual void Deinit() {}
+    virtual void Lose_Focus() {}
+    virtual void Regain_Focus() {}
+    virtual int Open(AsciiString title) { return 0; }
+    virtual int Load(AsciiString title) { return 0; }
+    virtual VideoStream *First_Stream() { return m_firstStream; }
```
janisozaur, line 47: You could make some methods here `const`

src/game/client/videoplayer/videoplayer.cpp
```diff
@@ -0,0 +1,181 @@
...
+ *            LICENSE
+ */
+#include "videoplayer.h"
+#include "videostream.h"
+#include <cstddef>
+
+#ifndef THYME_STANDALONE
+VideoPlayer *&g_theVideoPlayer = Make_Global<VideoPlayer *>(0x00A2C0AC);
+#else
+VideoPlayer *g_theVideoPlayer = nullptr;
```
janisozaur, line 22: Is it being initialised anywhere?

OmniBlade: Its initialised in GameEngine::Init which is unfinished so the original code currently initialises it.

src/game/client/videoplayer/videoplayer.h
OmniBlade: I'll double check the original prototypes, its possible some are intended to be const, First_Stream should almost certainly be const from the look of it. The others will have more complex final implementations in a child class though, the original has an implementation built around the bink library which we'll almost certainly want to replace with the ffmpeg libs.

PR #49, opened 2018-08-21
DDS File loading

OmniBlade:
Implements loading code for the DDS format as used by Zero Hour.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -190,7 +190,10 @@ set(GAMEENGINE_SRC
     w3d/math/matrix3d.cpp
     w3d/math/matrix4.cpp
     w3d/math/quaternion.cpp
+    w3d/renderer/colorspace.cpp
+    w3d/renderer/ddsfile.cpp
     w3d/renderer/dx8wrapper.cpp
+    w3d/renderer/w3dformat.cpp
 )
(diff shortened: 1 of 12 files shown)


ghost:
Future support of BC7 maybe? Would be nice to add placeholders at least so it's noted in the code.

Here's how Dolphin-Emu did it, not that big of a deal https://github.com/dolphin-emu/dolphin/pull/5834 (not sure if that example is a valid comparison at all tho)

http://i.imgur.com/9KZNdpS.jpg

OmniBlade:
Looks like they were able to add support because the 3D frameworks they are using natively support textures compressed in that format. While we are stuck with D3D8 we would need a software decoder to uncompress it before loading to the GPU. I can certainly add stub enums and detection code though to note it for future support.

ghost:
Yeah, and note their presence in the console/logs (but truncate the message so it doesn't spam for each file, maybe a report, X total files found, Y dds DXT5 textures, Z dds BC7 textures, Q non-DDS files) 

As long as with D3D8, it would be a special feature indeed, because that would take much more VRAM/RAM, ofcourse for most modern PCs it shouldn't be that of an issue.

https://github.com/dolphin-emu/dolphin/pull/5279

PR #55, opened 2018-10-11
Build standalone version by default when not targetting Windows

janisozaur:


OmniBlade:
Thanks for all these cleanup patches, your contributions are appreciated.

PR #56, opened 2018-10-11
Remove C++14 feature

janisozaur:
auto return types are a C++14 feature, but the project is C++11


janisozaur:
Can you also please merge https://gitlab.com/OmniBlade/gamemath/merge_requests/1 ? I will need to apply the same c++14 change in there as well, after which I'd like to request an update to the submodule.

PR #59, opened 2018-10-19
Atomic builtin checking code.

OmniBlade:
Adds a CMake module to detect GCC style atomic operation intrinsics and updates the code to make use of them.

### DIFF
--- a/src/game/common/system/asciistring.h
+++ b/src/game/common/system/asciistring.h
@@ -19,6 +19,11 @@
 #include <cstdarg>
 #include <cstring>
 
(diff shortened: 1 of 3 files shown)


janisozaur:
Is it really needed? You're already using c++11, so why not just use its built-in atomics?

OmniBlade:
That is a good point, I'll check if using it maintains the same abi and switch to using it if it does. If it doesn't I can still use it for standalone builds.

OmniBlade:
Seems STLPort interferes with using the atomics for the dll build, so I'll have to make so with using it for standalone.

PR #66, opened 2018-12-03
Refactors debugging and library detected defines.

OmniBlade:
Debug asserts now generate a debug break and can
optionally be ignored.
Macros defined for detecting which libraries we are
building against are now BUILD_WITH_X.
Logging an asserts can be disabled independantly.


CMakeLists.txt
```diff
@@ -190,6 +198,13 @@ if(CMAKE_SIZEOF_VOID_P EQUAL 4 AND STANDALONE)
 endif()
 
 find_package(ICU COMPONENTS data i18n io tu uc)
+
+if(NOT WIN32 OR NOT "${CMAKE_SYSTEM}" MATCHES "Windows")
+    if(NOT ICU_FOUND)
+        message(FATAL_ERROR "ICU is required on none windows platforms and was not found.")
```
janisozaur, line 204: Typo? Should've been `non`

PR #80, opened 2019-01-25
Drop const on trivial return types

janisozaur:


OmniBlade:
Thanks for catching these, they date back to early RedAlert++ code that the file class stuff is based on where they were mistakenly added and I clearly wasn't thorough enough in cleaning them out when I ported it

PR #121, opened 2020-01-24
Implements SurfaceClass.

OmniBlade:


src/w3d/renderer/dx8wrapper.cpp
```diff
@@ -254,6 +259,38 @@ w3dsurface_t DX8Wrapper::Create_Surface(unsigned width, unsigned height, WW3DFor
...
 }
 
+w3dsurface_t DX8Wrapper::Create_Surface(const char *name)
+{
+    char buf[PATH_MAX];
+
+    { // auto_file_ptr scope.
+        auto_file_ptr fptr(g_theFileFactory, name);
+
+        // If the initial file name isn't available try the dds version.
```
duncanspumpkin: If the initial file is a tga and is unavailable

src/w3d/renderer/dx8wrapper.cpp
```diff
@@ -254,6 +259,38 @@ w3dsurface_t DX8Wrapper::Create_Surface(unsigned width, unsigned height, WW3DFor
...
+        if (!fptr->Is_Available()) {
+            strlcpy(buf, name, sizeof(buf));
+            char *ext = strchr(buf, '.');
+
+            if (strlen(ext) == 4) {
+                if ((ext[1] == 't' || ext[1] == 'T') && (ext[2] == 'g' || ext[2] == 'G')
```
duncanspumpkin: Could replace this with something a bit nicer to read which uses strncmp

src/w3d/renderer/dx8wrapper.cpp
```diff
@@ -254,6 +259,38 @@ w3dsurface_t DX8Wrapper::Create_Surface(unsigned width, unsigned height, WW3DFor
...
+                    && (ext[3] == 'a' || ext[3] == 'A')) {
+                    ext[1] = 'd';
```
duncanspumpkin: A strcpy would be easier to read

PR #124, opened 2020-02-11
Implements TextureClass.

OmniBlade:


src/w3d/renderer/w3dformat.h
```diff
@@ -171,6 +171,25 @@ inline uint8_t Alpha_Bits(WW3DFormat format)
     return 0;
 }
 
+/**
+ * Determine if the format is a compressed format.
+ */
+inline bool Is_Compressed(WW3DFormat format)
+{
+    switch (format) {
```
altimumdelta: Couldn't we just switch to BCn semantics and to support future texture mods with BC7 compression? Or it has to be strictly what C&C supports here?

src/w3d/renderer/w3dformat.h
```diff
@@ -171,6 +171,25 @@ inline uint8_t Alpha_Bits(WW3DFormat format)
...
+        // All cases fall through.
+        case WW3D_FORMAT_DXT1:
+        case WW3D_FORMAT_DXT2:
+        case WW3D_FORMAT_DXT3:
+        case WW3D_FORMAT_DXT4:
+        case WW3D_FORMAT_DXT5:
```
ghost, line 185: Couldn't we just switch to BCn semantics and to support future texture mods with BC7 compression? Or it has to be strictly what C&C supports here?

OmniBlade: I don't believe the D3D8 backend which is what the game currently uses would even support such textures at the moment. Once we fully control the rendering engine and can upgrade it we can look at supporting newer texture formats then.

PR #127, opened 2020-03-02
Crashpad integration.

OmniBlade:
Integrates with Google's Crashpad library to add industry standard minidump crash log format and an automated uploading facility for crash reports.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "src/libs/captnlog"]
 	path = src/libs/captnlog
 	url = https://github.com/TheAssemblyArmada/CaptainsLog.git
+[submodule "src/libs/crashpad"]
+	path = src/libs/crashpad
+	url = https://github.com/TheAssemblyArmada/crashpad-cmake.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,6 @@ before_install:
 before_script:
   - mkdir build
   - cd build
-  - cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DSTANDALONE=TRUE ..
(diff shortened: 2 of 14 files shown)


janisozaur:
This is quite interesting. You mention ability to upload crash dumps, but looking at the code the URL is empty? So you plan on running your own Socorro or similar? In OpenRCT2 we use the open source offering of backtrace.io and are very happy with it.

ghost:
Wait that means I have to be connected to uncle google when playing ZH with this?

janisozaur:
@altimumdelta where did you get that kind of an idea?

OmniBlade:
@janisozaur I don't think we are going to hard code anything at this point, but I've implememted a configuration file that can set the URL and enable the automated reporting that could be set on a first run or via a configuration program similar to the OpenRCT2 one. Once we have enough implemented to add new UI elements we will probably want to add some kind of configuration in game. I've done some testing with sentry.io's offering and need to work out how to automate symbolification and how to generate a symbol file for the original game binary.

janisozaur:
Nice. I liked sentry's stuff as well when I was evaluating it, but they didn't seem to have to proper (or automated enough?) C++ symbolication et al. in place back then. I know it has changed in the last few months, maybe even earlier.

Some tips:
1. include the git commit as attribute when uploading
2. if possible/relevant, attach a save/log/etc.
3. same, but with screenshot
4. make sure user consents

It's also nice to have backtrace authorized to our project so it creates tickets for each _unique_ crash: as we have fairly large enough base, we have set it to collect dumps for about an hour or so before the issue is created, then we get it with aggregated count.

OmniBlade:
Seems like attaching files only works on linux with the current upstream source. There are forks that add the same feature to the handler for windows and macos as well. I'll probably revisit attachments later when we have something that end users might actually use on a regular basis.

janisozaur:
Really? That's quite surprising, we've been using breakpad (crashpad's predecessor) with attachments reliably for quite some time now only on Windows

PR #131, opened 2020-04-26
Add various definitions as well as vertex and index buffer stuff

jonwil:
Add various definitions as well as vertex and index buffer stuff

### DIFF
--- a/src/platform/w3dengine/client/w3droadbuffer.h
+++ b/src/platform/w3dengine/client/w3droadbuffer.h
@@ -68,8 +68,8 @@ struct TRoadPt
     int last;
     char multi;
     char count;
-    char isAngled;
-    char isJoin;
+    bool isAngled;
+    bool isJoin;
 };
 
 enum TCorner
--- a/src/platform/w3dengine/client/w3droadbuffer.h
+++ b/src/platform/w3dengine/client/w3droadbuffer.h
@@ -1,0 +1,213 @@
+/**
+ * @file
+ *
+ * @author Jonathan Wilson
+ *
+ * @brief Road Drawing Code
+ *
+ * @copyright Thyme is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#pragma once
+#include "always.h"
+#include "vector2.h"
+#include "asciistring.h"
+#include "sphere.h"
+#include "multilist.h"
+class TextureClass;
+class DX8VertexBufferClass;
+class DX8IndexBufferClass;
+class WorldHeightMap;
+class RenderObjClass;
+struct VertexFormatXYZDUV1;
+class CameraClass;
+struct IRegion2D;
+
+class RoadType
+{
+public:
+    RoadType();
+    ~RoadType();
+    void Load_Texture(Utf8String path, int ID);
+    void Apply_Texture();
+    int Get_Stacking() { return m_stackingOrder; }
+    void Set_Stacking(int stacking) { m_stackingOrder = stacking; }
+    int Get_Unique_ID() { return m_uniqueID; }
+    DX8VertexBufferClass *Get_VB() { return m_vertexRoad; }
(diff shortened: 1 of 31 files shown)


src/w3d/lib/simplevec.h
```diff
@@ -0,0 +1,210 @@
...
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#pragma once
+#include "always.h"
+
+template<class T>
```
OmniBlade: Can we use `typename` in preference to `class` for templates? Another thing I'll add to the style guide :D

src/w3d/lib/simplevec.h
```diff
@@ -0,0 +1,210 @@
...
+        if (m_vector != nullptr) {
+            delete[] m_vector;
+            m_vector = nullptr;
+            m_vectorMax = 0;
+        }
+    }
+    T &operator[](int index)
+    {
+        captainslog_assert(index < m_vectorMax);
+        return (m_vector[index]);
```
OmniBlade: Can we be consistent on how values are returned? Specifically without `()`?

src/platform/w3dengine/client/w3droadbuffer.h
```diff
@@ -0,0 +1,213 @@
...
+
+struct TRoadPt
+{
+    Vector2 loc;
+    Vector2 top;
+    Vector2 bottom;
+    int last;
+    char multi;
+    char count;
+    char isAngled;
```
OmniBlade: From the name, these last two look like bools rather than char?

src/w3d/lib/simplevec.h
```diff
@@ -0,0 +1,210 @@
...
+};
+
+template<class T>
+class SimpleDynVecClass : public SimpleVecClass<T>
+{
+public:
+    SimpleDynVecClass(int size = 0) : SimpleVecClass<T>(size), m_activeCount(0) {}
+    virtual ~SimpleDynVecClass(void)
+    {
+        if (m_vector != NULL) {
```
OmniBlade: Prefer nullptr over NULL.

src/w3d/lib/simplevec.h
```diff
@@ -0,0 +1,210 @@
...
+        m_activeCount = 0;
+        if (allow_shrink) {
+            Shrink();
+        }
+    }
+
+protected:
+    bool Grow(int new_size_hint)
+    {
+        int new_size = MAX(Length() + Length() / 4, Length() + 4);
```
OmniBlade: Prefer std::max and std::min from <algorithms> header.

src/w3d/renderer/dx8fvf.cpp
```diff
@@ -0,0 +1,74 @@
...
+#else
+    return 0;
+#endif
+}
+
+FVFInfoClass::FVFInfoClass(unsigned int FVF_, unsigned int fvf_size_)
+{
+#ifdef BUILD_WITH_D3D8
+    m_FVF = FVF_;
+    if (FVF_) {
```
OmniBlade, line 34: We use explicit value comparisons for conditions, so `if (FVF_ != 0)` in this case. Only bools and bitmask tests are treated as true/false statements for the project style.

src/platform/w3dengine/client/w3droadbuffer.h
```diff
@@ -0,0 +1,213 @@
...
+    float vOffset;
+    float scale;
+};
+
+class RoadSegment
+{
+public:
+    RoadSegment();
+    ~RoadSegment();
+    void Set_Vertex_Buffer(VertexFormatXYZDUV1 *vb, int numVertex);
```
OmniBlade, line 104: Prefer variable_name style for parameters, in this case num_vertex in place of num_vertex.

OmniBlade (changes requested):
Bunch of coding style feedback here, mostly nit picks. If I've picked up on anything that isn't in the style guide let me know so I can ensure I'm not complaining about stuff that isn't actually documented that we want it that way.

OmniBlade:
Superceeded by #132

PR #134, opened 2020-05-11
Cleanup a bunch of math classes to make them like the stock code.

jonwil:


src/w3d/math/matrix3.cpp
```diff
@@ -152,18 +144,100 @@ void Matrix3::Multiply(const Matrix3 &A, const Matrix3 &B, Matrix3 *set_res)
...
+
     tmp1 = B[0][2];
     tmp2 = B[1][2];
     tmp3 = B[2][2];
     (*set_res)[0][2] = (float)((*Aptr)[0][0] * tmp1 + (*Aptr)[0][1] * tmp2 + (*Aptr)[0][2] * tmp3);
     (*set_res)[1][2] = (float)((*Aptr)[1][0] * tmp1 + (*Aptr)[1][1] * tmp2 + (*Aptr)[1][2] * tmp3);
     (*set_res)[2][2] = (float)((*Aptr)[2][0] * tmp1 + (*Aptr)[2][1] * tmp2 + (*Aptr)[2][2] * tmp3);
 }
+
+int Matrix3::Is_Orthogonal() const
```
OmniBlade, line 163: This looks like it should be a bool return?

src/w3d/math/matrix3.h
```diff
@@ -303,30 +281,35 @@ class Matrix3
     friend void Swap(Matrix3 &a, Matrix3 &b);
     friend Vector3 operator*(const Matrix3 &a, const Vector3 &v);
 
-    __forceinline static void Add(const Matrix3 &a, const Matrix3 &b, Matrix3 *res)
+    __forceinline static void Add(const Matrix3 &a, const Matrix3 &b, Matrix3 *c)
     {
-        Vector3::Add(a.Row[0], b.Row[0], &(res->Row[0]));
-        Vector3::Add(a.Row[1], b.Row[1], &(res->Row[1]));
-        Vector3::Add(a.Row[2], b.Row[2], &(res->Row[2]));
+        captainslog_assert(c);
```
OmniBlade: Should be explicit with c != nullptr?

src/w3d/math/matrix4.h
```diff
@@ -145,69 +135,133 @@ class Matrix4
...
-        
+
         return det;
     }
 
     __forceinline Matrix4 Inverse() const
     {
         float s = Determinant();
-        if ( s == 0.0 ) return Matrix4();
+        if (s == 0.0)
```
OmniBlade: Needs braces.

src/w3d/math/quaternion.cpp
```diff
@@ -163,39 +170,125 @@ Quaternion Quaternion::Trackball(float x0, float y0, float x1, float y1, float s
...
-
+    static int _nxt[3] = {1, 2, 0};
     float tr, s;
     int i, j, k;
     Quaternion q;
+    tr = mat[0][0] + mat[1][1] + mat[2][2];
 
+    if (tr > 0.0f) {
+        s = GameMath::Sqrt(tr + 1.0);
+        q[3] = s * 0.5;
```
OmniBlade, line 181: Are we sure all these floating point literals are intended to be double? Doesn't look that way in Mac binary. It looks like the + 1.0 might be intentional though so I may need to look at implementing the double precision versions of the gamemath wrappers in the future.

src/w3d/math/vector3.h
```diff
@@ -209,195 +176,191 @@ class Vector3
...
     }
 
-    __forceinline Vector3 Abs() const
+    __forceinline static float Distance(const Vector3 &p1, const Vector3 &p2)
     {
-        return Vector3(GameMath::Fabs(X), GameMath::Fabs(Y), GameMath::Fabs(Z));
+        Vector3 temp;
+        temp = p1 - p2;
+
+        return (temp.Length());
```
OmniBlade: Don't need brackets for return.

src/w3d/math/vector3.h
```diff
@@ -444,10 +408,20 @@ __forceinline bool operator!=(const Vector3 &a, const Vector3 &b)
...
+    return ((GameMath::Fabs(a.X - b.X) < epsilon) && (GameMath::Fabs(a.Y - b.Y) < epsilon)
+        && (GameMath::Fabs(a.Z - b.Z) < epsilon));
+}
+
 __forceinline Vector3 Normalize(const Vector3 &vec)
 {
-    float len2 = GAMEMATH_FLOAT_TINY + vec.Length2();
-    return vec * GameMath::Inv_Sqrt(len2);
+    float len2 = vec.Length2();
+    if (len2 != 0.0f) {
```
OmniBlade, line 422: New lines around scopes please.

src/w3d/math/vector4.h
```diff
@@ -24,50 +24,106 @@ class Vector4
...
+    __forceinline void Normalize()
     {
-        float len2 = GAMEMATH_FLOAT_TINY + Length2();
-        float oolen = GameMath::Inv_Sqrt(len2);
-        X *= oolen;
-        Y *= oolen;
-        Z *= oolen;
-        W *= oolen;
+        float len2 = Length2();
+        if (len2 != 0.0f) {
```
OmniBlade, line 78: New lines around scopes please.

OmniBlade (changes requested):

src/w3d/math/matrix3.cpp
jonwil: Is_Orthogonal is int in the original code so we should do the same.

OmniBlade (approved):

PR #140, opened 2020-07-12
[AsciiString/Utf8String] Adds hooks to all methods of the Utf8String …

Smileynator:
…class, Adds method documentation, Fixes potential size issue with ToLower buffer (macOS), Fixes issue with Format methods that did not call va_end

This was my first attempt at working in C++ in a long time. And first real reverse engineering attempt. As exercise i validated the functionality of all the methods before i hooked them. And playtested the game on a basic level to verify nothing significantly broke.

### DIFF
--- a/src/game/common/system/asciistring.cpp
+++ b/src/game/common/system/asciistring.cpp
@@ -21,35 +21,41 @@
 
 Utf8String const Utf8String::s_emptyString(nullptr);
 
-Utf8String::Utf8String() :
-    m_data(nullptr)
+/**
+ * Default constructor added for convenience
+ */
+Utf8String::Utf8String() : m_data(nullptr)
 {
 }
 
-Utf8String::Utf8String(const char *s) :
-    m_data(nullptr)
+/**
+ * Inits this string with an existing string (copy) and increments reference count
+ */
+Utf8String::Utf8String(Utf8String const &string) : m_data(string.m_data)
+{
+    if (m_data != nullptr) {
+        m_data->Inc_Ref_Count();
+    }
+}
+
+/**
+ * Inits this string with a reference to the start of a char array.
+ */
+Utf8String::Utf8String(const char *s) : m_data(nullptr)
 {
     if ( s != nullptr ) {
+		//Get length of the string that was passed
         int len = (int)strlen(s);
         if ( len > 0 ) {
+			//Allocate a buffer
             Ensure_Unique_Buffer_Of_Size(len + 1, false, s);
         }
     }
 }
 
-Utf8String::Utf8String(Utf8String const &string) :
-    m_data(string.m_data)
-{
-    if ( m_data != nullptr ) {
-        m_data->Inc_Ref_Count();
-    }
-}
-
-void Utf8String::Validate()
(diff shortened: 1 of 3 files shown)


src/game/common/system/asciistring.cpp
```diff
@@ -21,56 +21,73 @@
...
+    }
+}
+
+/**
+ * Inits this string with a reference to the start of a char array.
+ */
+Utf8String::Utf8String(const char *s) : m_data(nullptr)
 {
     if ( s != nullptr ) {
+		//Get length of the string that was passed
```
OmniBlade, line 47: Do we have tabs here instead of spaces? Alignment looks odd in the github diff. The whole file looks like needs the current formatting rules applied as its predates the clang-format file and is inconsistent with more recent code.

src/game/common/system/asciistring.cpp
```diff
@@ -21,56 +21,73 @@
...
     captainslog_dbgassert(m_data != nullptr, "null string ptr");
     
     return m_data->Peek();
 }
 
+/**
+ * //Internal shorthand to decrement the reference count and release buffer if 0
 
-void Utf8String::Release_Buffer()
+ */void Utf8String::Release_Buffer()
```
OmniBlade: Missing newline?

src/game/common/system/asciistring.cpp
```diff
@@ -139,6 +159,9 @@ char Utf8String::Get_Char(int index) const
     return Peek()[index];
 }
 
+/**
+ * This is effectively ToString() It is in original code, but never called.
```
OmniBlade, line 163: Its used a lot but is inlined everywhere its called I think, probably in the header originally based on that, but lets not worry about that for now.

src/game/common/system/asciistring.cpp
```diff
@@ -150,28 +173,34 @@ const char *Utf8String::Str() const
...
+        if ( c && (len = strlen(c) + 1, len != 1) ) {
+            Ensure_Unique_Buffer_Of_Size(len, false, c, nullptr);
         } else {
             Release_Buffer();
         }
     }
 }
 
+/**
+ * A convenience version of Set. Not found in original code.
```
OmniBlade, line 202: Probably inlined in the copy constructor.

OmniBlade (changes requested):
Its been ages since I looked at this code properly, its some of the earliest code in the project. Nice to get a fresh set of eyes looking over it and noticing stuff I missed.

src/game/common/system/asciistring.cpp
Smileynator: As discussed i will add a commit to do all the Clang stuff.

src/game/common/system/asciistring.cpp
Smileynator: The message can be edited by whoever knows better. I just go with what i found through what i had.

Smileynator:
@OmniBlade Resolved the issues, and added an extra commit with all Clang changes.

OmniBlade (approved):

Issue #100, opened 2019-08-30
Bug: AI is not working correctly


duncanspumpkin:
#### Original game source: game disc

- [ ] Reproducible in the original Generals Zero Hour? No.
- [ ] Multiplayer affected? No.
- [ ] Are you using any mods? No.

### Issue explanation
It looks like the script reading isn't working correctly for the AI. I did a bit of debugging myself just commenting out hooks and narrowed it down to somewhere in either: script.h, scriptaction.h, scriptgroup or scriptlist.

https://github.com/TheAssemblyArmada/Thyme/blob/develop/src/hooker/setuphooks_zh.cpp#L467-L491

I'll try narrow it down when I get the time but it might not be for a while. So thought I'd make an issue to record it.

### Steps to reproduce the problem
1. Start a 1vs1 skirmish with a Hard AI
2. Build a few items then go attack the Hard AI
3. Notice that the Hard AI has not constructed very much at all.

PR #142, opened 2020-08-19
Fixes AI not working correctly.

OmniBlade:
Fixes bug in the Thyme implementation of the games scripting code.

### DIFF
--- a/src/game/logic/scriptengine/scriptaction.cpp
+++ b/src/game/logic/scriptengine/scriptaction.cpp
@@ -57,7 +57,9 @@ ScriptAction *ScriptAction::Duplicate()
     ScriptAction *new_action = new ScriptAction(m_actionType);
 
     for (int i = 0; i < m_numParams; ++i) {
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
Is this fixing #100?

OmniBlade:
> 
> 
> Is this fixing #100?

It is :D, had an `if` that was `==` instead of `!=`.

duncanspumpkin:
Excellent. I spent ages trying to work out the source of the issue but didn't know enough to crack it. 

Btw you probably know this but if you add "Fix: #isuuenumber" then GitHub will auto close the referenced issue when the pr is merged. There a could other keywords that do it as well. I find it useful to avoid additional clean up work.

OmniBlade:
The second commit already has `Fixes #issuenumber` in its message which does the same thing, not sure whether its best practice to put it in the commit message or the PR description though. Probably the PR now that I think about it since its specific to the repository.

PR #150, opened 2020-08-30
Image class

Smileynator:
This should be all for the image class. I have found no issues in testing myself.
2 methods are in there, but i have not found their location in the WIN binary. I believe they are properly implemented though.

Sorry about clang there, seems it got upset with my latest changes to Urf8String and decided to do something about that just now. Not sure how or when that is supposed to happen but it looks like it is pure formatting either way. I have not applied any changes to the file myself.


src/CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ set(GAMEENGINE_SRC
     game/client/view.cpp
     game/client/water.cpp
     game/client/gui/gamefont.cpp
+	game/client/image.cpp
```
OmniBlade: Tab rather than spaces here? Clang format doesn't handle the cmake I don't believe.

src/game/client/image.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+Image::~Image() 
+{
+    return;
+}
+
+void Image::Set_Status(uint32_t bit)
+{
+    m_status = m_status | bit;
```
OmniBlade: can simplify this to `m_status |= bit;`

src/game/client/image.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+void Image::Clear_Status(uint32_t bit)
+{
+    m_status = m_status & ~bit;
```
OmniBlade: Can simplify to `m_status &= ~bit;`

src/game/client/image.h
```diff
@@ -0,0 +1,54 @@
...
+
+private:
+    Utf8String m_name;
+    Utf8String m_filename;
+    ICoord2D m_textureSize;
+    Region2D m_UVCoords;
+    ICoord2D m_imageSize;
+    TextureClass *m_rawTextureData;
+    uint32_t m_status;
+};
```
OmniBlade: Missing newline at end of file.

OmniBlade (changes requested):
Just a few minor style fixes to do, plus if you want to do the simplified bitwise assignments.

src/game/client/image.cpp
```diff
@@ -0,0 +1,90 @@
...
+	m_name(), 
+	m_filename(), 
+	m_UVCoords(), 
+	m_textureSize(), 
+	m_imageSize(), 
+	m_rawTextureData(), 
+	m_status()
+{
+    m_UVCoords.hi.x = 1.0;
+    m_UVCoords.hi.x = 1.0;
```
tomsons26: should be setting y here

src/game/client/image.cpp
```diff
@@ -0,0 +1,90 @@
...
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#pragma once
+
+#include "always.h"
+#include "image.h"
+
+const char *imageStatusNames[3] = {"ROTATED_90_CLOCKWISE", "RAW_TEXTURE", nullptr};
```
tomsons26: was likely a static in the function that uses it, also s_imageStatusNames if you make a global, _imageStatusNames if you put it in the function

src/game/client/image.cpp
```diff
@@ -0,0 +1,90 @@
...
+    image->m_imageSize.x = right - left;
+    image->m_imageSize.y = bottom - top;
+}
+
+void Image::Parse_Image_Status(INI *ini, void *formal, void *store, const void *user_data)
+{
+    Image *image = static_cast<Image *>(formal);
+    ini->Parse_Bitstring32(ini, formal, store, imageStatusNames);
+    uint8_t *byte = static_cast<uint8_t *>(store);
+    if (byte != 0) {
```
tomsons26: isn't correct,
 it checks a bit here,
bitstrings correspond to enum's showing what bit is what, in this case
enum ImageStatus {
	IMAGE_STATUS_NONE,
	IMAGE_STATUS_ROTATED_90_CLOCKWISE,
	IMAGE_STATUS_RAW_TEXTURE
};
its checking for IMAGE_STATUS_ROTATED_90_CLOCKWISE
so (*store & IMAGE_STATUS_ROTATED_90_CLOCKWISE)

src/game/client/image.h
```diff
@@ -0,0 +1,54 @@
...
+public:
+    Image();
+    virtual ~Image();
+
+    // Methods
+    void Clear_Status(uint32_t bit);//TODO not hooked or verified
+    void Set_Status(uint32_t bit);
+
+    // initFromINIMulti variants for Field Parsing Functions.
+    static void Parse_Image_Coords(INI *ini, void *formal, void *store, const void *user_data);
```
tomsons26, line 38: formal is what MSVC uses when there's no name, the actual name for these is instance

Smileynator:
All issues should be resolved as suggested or discussed.

OmniBlade (approved):

PR #154, opened 2020-09-03
 Implements 3 Collision Test classes.

tomsons26:
RayCollisionTestClass remains but haven't found it yet.
Code based on WB, checked with ZH to make sure there aren't ifdef DEBUG stuff going on. 
Afaik this is mostly inlined in the game itself.
Code was very clean so i have no doubts these are correct.

### DIFF
--- a/src/w3d/renderer/coltest.cpp
+++ b/src/w3d/renderer/coltest.cpp
@@ -1,0 +1,264 @@
+/**
+ * @file
+ *
+ * @author tomsons26
+ *
+ * @brief Classes for testing collision.
+ *
+ * @copyright Thyme is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#include "coltest.h"
+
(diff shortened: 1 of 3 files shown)


src/game/common/commandline.cpp
```diff
@@ -511,7 +511,7 @@ int Parse_Wire_Frame(char **argv, int argc)
 {
 #ifdef GAME_DEBUG_STRUCTS
     if (g_theWriteableGlobalData != nullptr) {
-        g_theWriteableGlobalData->m_wireFrame = true;
```
OmniBlade: This is fixed in another PR and doesn't belong in this one, can you rebase to remove the need for this? Thanks.

src/w3d/renderer/coltest.cpp
```diff
@@ -0,0 +1,274 @@
...
+        m_sweepMin.Y = endmin.Y;
+    }
+    if (endmin.Z < m_sweepMin.Z) {
+        m_sweepMin.Z = endmin.Z;
+    }
+}
+
+bool AABoxCollisionTestClass::Cull(const AABoxClass &box)
+{
+    // Original did this, what's more efficient?
```
OmniBlade: Same difference really, you can drop the comment though, no need to keep that around.

OmniBlade (changes requested):

PR #156, opened 2020-09-12
Game Font changes and additions

tomsons26:
Would be good if someone else would take a look at the functions, make sure i didn't miss anything
Also not sure on how GameFont pool was implemented, i think what i have is right, all i can tell for sure is that it has a pool which we missed before.

### DIFF
--- a/src/game/client/gui/gamefont.cpp
+++ b/src/game/client/gui/gamefont.cpp
@@ -1,9 +1,10 @@
-/**
+﻿/**
  * @file
  *
  * @author OmniBlade
+ * @author tomsons26
  *
- * @brief Information relating to a font to use in game.
+ * @brief Font information and handling classes.
  *
  * @copyright Thyme is free software: you can redistribute it and/or
  *            modify it under the terms of the GNU General Public License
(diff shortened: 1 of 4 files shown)


src/game/client/gui/gamefont.cpp
```diff
@@ -12,4 +13,105 @@
...
+}
+
+void FontLibrary::Reset()
+{
+    Delete_All_Fonts();
+}
+
+void FontLibrary::Link_Font(GameFont *font)
+{
+    if (font) {
```
OmniBlade: Can we make this an explicit font != nullptr check as done elsewhere?

src/hooker/setuphooks_zh.cpp
```diff
@@ -1009,4 +1009,8 @@ void Setup_Hooks()
     Hook_Method(0x00613AB0, &RayEffectSystem::Add_Ray_Effect);
     Hook_Method(0x00613B30, &RayEffectSystem::Delete_Ray_Effect);
     Hook_Method(0x00613B70, &RayEffectSystem::Get_Ray_Effect_Data);
+
+    // gamefont.h
+    Hook_Method(0x0041FE60, &FontLibrary::Unlink_Font);
```
OmniBlade, line 1014: Do we not know where the other implemented functions are in the binary?

OmniBlade (changes requested):

src/hooker/setuphooks_zh.cpp
tomsons26: Inlined

OmniBlade (approved):

PR #161, opened 2020-09-25
Adds github workflows for building with msvc on windows and gcc on linux.

OmniBlade:
Also changes the issue template slightly to stop issues with checkboxes being created and updated the readme to be in line with EA's new C&C modding policies.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,0 +1,24 @@
+name: linux
+
+on:
+  pull_request:
+  push:
+  release:
+    types: published
+
+jobs:
+  build_thyme_ubuntu_gcc:
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: 'recursive'
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get install -y libgtk-3-dev ninja-build
+
+    - name: Build thyme
+      run: |
+        cmake -DSTANDALONE=TRUE -DCMAKE_BUILD_TYPE=Release -DUSE_CRASHPAD=ON -B build -G Ninja
+        cmake --build build
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,0 +1,53 @@
+name: windows
+
+on:
+  pull_request:
+  push:
+  release:
+    types: published
+
+jobs:
+  build_thyme_msvc:
+    runs-on: windows-2019
+    strategy:
+      matrix:
+        standalone: [ON, OFF]
+        platform: [x86, amd64]
+        exclude:
+          - {standalone: "OFF", platform: amd64}
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: 'recursive'     
+    - uses: ilammy/msvc-dev-cmd@v1.2.0
+      with:
+        arch: ${{ matrix.platform }}
+      
+    - name: Install dependencies
+      shell: bash
+      run: |
+        curl -L -O https://github.com/ninja-build/ninja/releases/download/v1.10.1/ninja-win.zip
+        7z x ninja-win.zip -o`pwd`/deps/ninja
+        export PATH=`pwd`/deps/ninja:$PATH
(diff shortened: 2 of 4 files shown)


.github/workflows/windows.yml
```diff
@@ -0,0 +1,44 @@
...
+  push:
+  release:
+    types: published
+
+jobs:
+  build:
+    runs-on: windows-2019
+    strategy:
+      matrix:
+        buildtype: [Debug, Release]
```
janisozaur: What's the need for both those types? I think that only wastes CPU time

.github/workflows/linux.yml
```diff
@@ -0,0 +1,27 @@
...
+  push:
+  release:
+    types: published
+
+jobs:
+  common:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        buildtype: [Debug, Release]
```
janisozaur: What's the point for using both here? Usually just the release is fine for CI

.github/workflows/linux.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: 'recursive'
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get install -y libgtk-3-dev
```
janisozaur:
```suggestion
        sudo apt-get install -y libgtk-3-dev ninja-build
```

.github/workflows/linux.yml
```diff
@@ -0,0 +1,27 @@
...
+
+    - name: Build thyme
+      run: |
+        cmake -DSTANDALONE=TRUE -DCMAKE_BUILD_TYPE=${{ matrix.buildtype }} -DUSE_CRASHPAD=ON -B build
```
janisozaur:
```suggestion
        cmake -DSTANDALONE=TRUE -DCMAKE_BUILD_TYPE=${{ matrix.buildtype }} -DUSE_CRASHPAD=ON -B build -G Ninja
```

.github/workflows/linux.yml
```diff
@@ -0,0 +1,27 @@
...
+        cmake --build build -- -j 4
```
janisozaur:
```suggestion
        cmake --build build
```

.github/workflows/windows.yml
```diff
@@ -0,0 +1,44 @@
...
+          - {buildtype: "Debug", standalone: "OFF", platform: x64}
+          - {buildtype: "Release", standalone: "OFF", platform: x64}
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: 'recursive'
+        
+    - name: Build Thyme
+      run: |
+        cmake -A ${{ matrix.platform }} -DSTANDALONE=${{ matrix.standalone }} -DUSE_CRASHPAD=${{ matrix.standalone }} -B build
```
janisozaur: You can use ninja here as well, but that requires changing how the arch is passed. You get sane error messages and better build speed in turn. I will try providing a sample.

janisozaur: Here's one for appveyor, I know it's not quite compatible, but it's good for starters: https://github.com/openblack/openblack/pull/68/files

.github/workflows/linux.yml
```diff
@@ -0,0 +1,24 @@
...
+
+on:
+  pull_request:
+  push:
+  release:
+    types: published
+
+jobs:
+  build_thyme_ubuntu_gcc:
+    runs-on: ubuntu-18.04
```
ghost, line 11: Couldn't we just jump to 20.04.1 LTS while we're at it, I just reinstalled a week ago, linux chat rooms refuse to give support for old version pretty quick. But I guess you have your env setup for 18.04 and it may take more work to move up?

OmniBlade: 18.04 is an LTS and still in support. The linux build is really just a sanity check to make sure windows only code is correctly walled off at the moment anyhow so it building on any linux version is kind of good enough. When playable standalone builds are a reality then what is built for can be reviewed.

PR #168, opened 2020-10-28
Implements Display::Delete_Views()

duncanspumpkin:
A relatively simple function that annoyingly can't use a more standard for loop as the next view needs to be found before deleteing the current view.

### DIFF
--- a/src/game/client/display.cpp
+++ b/src/game/client/display.cpp
@@ -14,8 +14,11 @@
  *            LICENSE
  */
 #include "display.h"
+#include "displaystringmanager.h"
 #include "gamefont.h"
 #include "mouse.h"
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
So turned out this was already implemented in another branch https://github.com/tomsons26/Thyme/commit/5502d282007f4e589828de3d85776d5e48e9ddd5 I've modified my code so that it matches that code with one change. I set unkDisplayString to nullptr after freeing this follows the assembly closer and is generally a good idea to perform.

OmniBlade (approved):

duncanspumpkin:
I've rebased and added the Hook_Any's. Appears to work fine from my testing.

PR #169, opened 2020-10-30
Implement Win32Mouse

duncanspumpkin:
All functions implemented and ready to go. Not tested as its a pain to get these hooked up nicely.

### DIFF
--- a/src/platform/input/win32mouse.cpp
+++ b/src/platform/input/win32mouse.cpp
@@ -22,9 +22,9 @@
 #include <winuser.h>
 #endif
 
-#ifdef PLATFORM_WINDOWS
-// 0x00A31F00 (note first entry not used as it is CURSOR_NONE) 8 directions
-static HCURSOR g_LoadedCursors[CURSOR_COUNT][8];
+#ifndef GAME_DLL
+HCURSOR Win32Mouse::s_loadedCursors[CURSOR_COUNT][8];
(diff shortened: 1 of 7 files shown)


duncanspumpkin:
This has now been tested in my W3DMouse branch and appears to work fine.

src/platform/input/win32mouse.cpp
```diff
@@ -0,0 +1,238 @@
...
+    // First cursor is not loaded. As this is CURSOR_NONE
+    for (auto i = 1; i < CURSOR_COUNT; ++i) {
+        auto &cursor_info = m_cursorInfo[i];
+        if (cursor_info.texture_name.Is_Empty()) {
+            continue;
+        }
+
+        if (cursor_info.directions <= 1) {
+            char buffer[256]{};
+            snprintf(buffer, 256, "data\\cursors\\%s.ANI", cursor_info.texture_name.Str());
```
OmniBlade: While I realise that the original binary probably has the strings like this, it would probably be best for future cross compatibility to use a correctly case sensitive path based on the installed data. "Data/Cursors/%s.ani" to be specific. That assumes that the winapi function handles the posix path separators correctly too, but most winapi functions do. Thoughts?

duncanspumpkin: Yes that makes sense. I've made that change. TBH should probably use the filesystem library to handle all of this eventually.

OmniBlade (approved):

PR #171, opened 2020-11-02
Implements RenderInfoClass.

tomsons26:


src/w3d/renderer/rinfo.h
```diff
@@ -0,0 +1,91 @@
...
+ * @copyright Thyme is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#pragma once
+#include "always.h"
+#include <captainslog.h>
```
OmniBlade: You don't appear to be logging anything from the header, so include this in the cpp file, not here.

OmniBlade (changes requested):

src/w3d/renderer/rinfo.cpp
```diff
@@ -0,0 +1,104 @@
...
+void RenderInfoClass::Push_Material_Pass(MaterialPassClass *matpass)
+{
+    if (m_additionalMaterialPassCount <= ADDITIONAL_MATERIAL_PASSES) {
+        if (matpass != nullptr) {
+            // TODO temp until MaterialPassClass is in
+            ((RefCountClass *)matpass)->Add_Ref();
+        }
+        m_additionalMaterialPassArray[m_additionalMaterialPassCount++] = matpass;
+    } else {
+        ++field_A8;
```
OmniBlade, line 58: Is this used anywhere else? It looks to me like is a counter for additional passes that are requested, but there is no room for, perhaps a debug counter for logging if additional passes are being requested?

tomsons26: Haven't been able to find

src/w3d/renderer/rinfo.h
```diff
@@ -0,0 +1,90 @@
...
+    int Additional_Pass_Count() { return m_additionalMaterialPassCount; }
+    MaterialPassClass *Peek_Additional_Pass(int i) const;
+
+    void Push_Override_Flags(RINFO_OVERRIDE_FLAGS flags);
+    void Pop_Override_Flags();
+    RINFO_OVERRIDE_FLAGS &Current_Override_Flags();
+
+public:
+    CameraClass &m_camera;
+    int field_4; // float - fog_scale, fog_start, fog_end or according to BFME2 Vector3 FogColor
```
OmniBlade: Can we make these float if we know they are float?

src/w3d/renderer/rinfo.cpp
```diff
@@ -0,0 +1,104 @@
...
+    for (int i = 0; i < ADDITIONAL_MATERIAL_PASSES; ++i) {
+        m_additionalMaterialPassArray[i] = nullptr;
+    }
+
+    for (int i = 0; i < MAX_OVERRIDE_FLAG_LEVEL; ++i) {
+        m_overrideFlag[i] = RINFO_OVERRIDE_DEFAULT;
+    }
+}
+
+inline RenderInfoClass::~RenderInfoClass()
```
OmniBlade: No point marking things inline in the cpp file.

OmniBlade (changes requested):

src/w3d/renderer/rinfo.h
tomsons26: we don't know anything, got no info on it, its just speculation based on known things
whenever/if ever we encounter it we figure out what to do

OmniBlade (approved):

PR #173, opened 2020-11-03
Finishes Display class.

tomsons26:


src/game/client/display.cpp
```diff
@@ -74,9 +78,110 @@ void Display::Reset()
  */
 void Display::Update()
```
tomsons26, line 79: This function needs a checkup by someone else, it seems to work right but im not sure if i didn't miss anything

src/game/client/display.cpp
```diff
@@ -74,9 +78,110 @@ void Display::Reset()
...
+        } else if (m_someLogoMovieInt2 < 0 && m_someLogoMovieInt1 < 0) {
+            Stop_Movie();
+        } else {
+            // some sort of text timing, we never see the text cause its color is black
+            if (!m_unkInt) {
+                if (m_unkDisplayString != nullptr) {
+                    Delete_Instance(m_unkDisplayString);
+                }
+
+                m_unkDisplayString = g_theDisplayStringManager->New_Display_String();
```
OmniBlade, line 96: Based on this the unknown string is something like `m_copyrightString` or some such?

src/game/client/display.cpp
```diff
@@ -74,9 +78,110 @@ void Display::Reset()
  */
 void Display::Update()
 {
-#ifdef GAME_DLL
-    Call_Method<void, Display>(PICK_ADDRESS(0x00421870, 0x007BDB67), this);
-#endif
+    /*
```
OmniBlade: No need to leave this comment block here, I'll check the function over before I give it final approval.

OmniBlade (changes requested):

src/game/client/display.cpp
tomsons26: To my best ability to tell this is a general purpose slot used in sort of emergency cases(in that this is not the intended way to handle this at all based on code i've seen so far).
W3DDisplay::Draw for instance, seems would use it for some debug prints, not clear what.

OmniBlade (approved):

PR #180, opened 2020-11-07
Implements CameraClass.

tomsons26:


src/w3d/renderer/camera.cpp
```diff
@@ -0,0 +1,414 @@
...
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+
+#include "camera.h"
+#include "dx8wrapper.h"
+#include "matrix4.h"
+#include "w3d.h"
+#include <math.h>
```
OmniBlade: Use GameMaths for the math functions?

OmniBlade (changes requested):
GameMaths is used for atan2 but not tan?

PR #183, opened 2020-11-07
Implements Render2DSentenceClass.

OmniBlade:


src/w3d/renderer/render2dsentence.cpp
```diff
@@ -370,3 +374,559 @@ void FontCharsClass::Free_Character_Arrays()
...
+        screen_rect += Location;
+        RectClass uv_rect = data.UVRect;
+        bool add_quad = true;
+
+        if (IsClippedEnabled) {
+            if (screen_rect.right <= ClipRect.left || screen_rect.bottom <= ClipRect.top) {
+                add_quad = false;
+            } else {
+                RectClass clipped_rect;
+                clipped_rect.left = std::max(screen_rect.left, ClipRect.left);
```
duncanspumpkin: This exact function is used in a number of places (w3ddisplay draw image for example). We should make a function for it.

OmniBlade: A function for RectClass perhaps? Any idea what it might be?

duncanspumpkin: It's to simplify https://github.com/TheAssemblyArmada/Thyme/blob/200cf1a6178ff0d72a1ef54203f9ac5b22e41a33/src/platform/w3dengine/client/w3ddisplay.cpp#L780
I think you would call this working out an intersection. In other projects I've had an intersection function for my rect classes.

OmniBlade: Yeah, looks like its intersection, I've finished the sentence class pretty much now, I'll add the intersect function to rectclass as another commit to simplify this bit of code then I think its ready for review and merge.

tomsons26:
Build_Textures, Draw_Sentence, Allocate_New_Surface, Build_Sentence_Centered, Build_Sentence_Not_Centered
Im a bit out of my depth now with ZH to verify these, rest look good

PR #185, opened 2020-11-08
W3dassetmgr

duncanspumpkin:
A number of the virtual functions were found to not be in use and have therefore been turned into assert(false) functions.
Free_Assets_With_Exclusion_List has not been implemented at this time. It requires an additional class most likely and it is not clear if it performs any useful function. Constructor of W3DAssetManager has not been implemented at this time due to the requirement for multiple PrototypeLoader classes. Non virtual member functions of GameAssetManager require further work on MeshClass to implement.

I suggest I remove the changes to SList as it appears there is no point in implementing it due to no classes requireing it.

SList has now been removed. Seperate PR's have been created for HAnimManager and HTreeManager that will need to be merged in first.

### DIFF
--- a/src/game/common/system/asciistring.h
+++ b/src/game/common/system/asciistring.h
@@ -238,3 +238,8 @@ inline Utf8String &Utf8String::operator+=(Utf8String const &s)
 
     return *this;
 }
+
+// Windows to POSIX names
+// strcmpi = strcasecmp
+// strnicmp = strncasecmp
+// sprintf_s = snprintf
--- a/src/platform/w3dengine/client/w3ddisplay.h
+++ b/src/platform/w3dengine/client/w3ddisplay.h
@@ -20,7 +20,7 @@
 #include "render2d.h"
 #include "scene.h"
 
-class W3DAssetManager;
+class GameAssetManager;
 
 class W3DDisplay final : public Display
 {
@@ -72,7 +72,7 @@ public:
     virtual float Get_Average_FPS() override { return m_averageFps; }
     virtual int Get_Last_Frame_Draw_Calls() override;
 
-    static W3DAssetManager *Get_AssetManager() { return s_assetManager; }
+    static GameAssetManager *Get_AssetManager() { return s_assetManager; }
     static SceneClass *Get_3DInterfaceScene() { return s_3DInterfaceScene; }
 
 private:
@@ -90,12 +90,12 @@ private:
     uint32_t m_nativeDebugDisplay;
 
 #ifdef GAME_DLL
-    static W3DAssetManager *&s_assetManager; // TODO: Actual type is GameAssetManager
+    static GameAssetManager *&s_assetManager;
     static SceneClass *&s_3DScene; // TODO: Actual type is RTS2DScene
     static SceneClass *&s_2DScene; // TODO: Actual type is RTS2DScene
     static SceneClass *&s_3DInterfaceScene; // TODO: Actual type is RTS3DInterfaceScene
 #else
-    static W3DAssetManager *s_assetManager; // TODO: Actual type is GameAssetManager
+    static GameAssetManager *s_assetManager;
     static SceneClass *s_3DScene; // TODO: Actual type is RTS2DScene
     static SceneClass *s_2DScene; // TODO: Actual type is RTS2DScene
     static SceneClass *s_3DInterfaceScene; // TODO: Actual type is RTS3DInterfaceScene
--- a/src/w3d/renderer/assetmgr.cpp
+++ b/src/w3d/renderer/assetmgr.cpp
@@ -13,6 +13,794 @@
  *            LICENSE
  */
 #include "assetmgr.h"
+#include "chunkio.h"
+#include "crc.h"
+#include "ffactory.h"
+#include "fileclass.h"
+#include "mesh.h"
+#include "proto.h"
+#include "render2dsentence.h"
+#include "string.h"
+#include "w3d_file.h"
+#include <cmath>
+#include <stdio.h>
+
 #ifndef GAME_DLL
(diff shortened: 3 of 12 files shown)


src/platform/w3dengine/client/w3ddisplay.h
```diff
@@ -90,12 +90,12 @@ class W3DDisplay final : public Display
     uint32_t m_nativeDebugDisplay;
 
 #ifdef GAME_DLL
-    static W3DAssetManager *&s_assetManager; // TODO: Actual type is GameAssetManager
+    static GameAssetManager *&s_assetManager; // TODO: Actual type is GameAssetManager
```
duncanspumpkin: Remove TODO message

src/w3d/renderer/assetmgr.h
```diff
@@ -12,33 +12,56 @@
...
 class PrototypeClass;
 class TextureFileCache;
 
+#ifdef GAME_DLL
+extern NullPrototypeClass &s_nullPrototype;
+#else
+extern NullPrototypeClass s_nullPrototype;
+#endif
+
 class W3DAssetManager
```
duncanspumpkin, line 57: Add note that `W3DAssetManager` is the renamed `WW3DAssetManager`

src/game/common/system/asciistring.h
```diff
@@ -238,3 +238,8 @@ inline Utf8String &Utf8String::operator+=(Utf8String const &s)
 
     return *this;
 }
+
+// Windows to POSIX names
```
OmniBlade, line 242: This change probably shouldn't be part of the PR I'm guessing?

OmniBlade (changes requested):
Looks like some stray local notes seeped into the PR?

src/game/common/system/asciistring.h
duncanspumpkin: I mentioned this a long time ago it's something that would be useful to have documented somewhere in the codebase so doing a search for either will hit these.

src/w3d/renderer/assetmgr.cpp
```diff
@@ -13,6 +13,794 @@
...
+        m_prototypes.Delete(i);
+        if (prototype != nullptr) {
+            prototype->Delete_Self();
+        }
+    }
+    memset(m_prototypeHashTable, 0, s_prototypeHashTableSize * sizeof(uintptr_t));
+    m_hAnimManager.Free_All_Anims();
+    m_hTreeManager.Free_All_Trees();
+    Release_All_Textures();
+    Release_All_FontChars();
```
tomsons26: should have a commented out line for Release_All_Font3DDatas call to keep a record that this was here

src/w3d/renderer/assetmgr.cpp
```diff
@@ -13,6 +13,794 @@
...
+            case W3D_CHUNK_COMPRESSED_ANIMATION:
+            case W3D_CHUNK_ANIMATION:
+            case W3D_CHUNK_MORPH_ANIMATION:
+                m_hAnimManager.Load_Anim(chunk);
+                break;
+            case W3D_CHUNK_HIERARCHY:
+                m_hTreeManager.Load_Tree(chunk);
+                break;
+            default: {
+                for (auto i = 0; i < m_prototypeLoaders.Count(); ++i) {
```
tomsons26: this is actually 3 inlines, Load_Prototype that calls Find_Prototype_Loader, Add_Prototype

src/w3d/renderer/assetmgr.cpp
```diff
@@ -13,6 +13,794 @@
...
+    Call_Method<void, W3DAssetManager>(0x00814090, this);
+#endif
+}
+
+// 0x008143C0
+W3DAssetManager::~W3DAssetManager()
+{
+    Free_Assets();
+    s_theInstance = nullptr;
+    delete[] m_prototypeHashTable;
```
tomsons26, line 48: Add comment that FontCharsList and Font3DDatas were deleted here

src/w3d/renderer/assetmgr.cpp
```diff
@@ -13,6 +13,794 @@
...
+        proto = Find_Prototype(name);
+    }
+    if (proto == nullptr) {
+        return nullptr;
+    }
+    return proto->Create();
+}
+
+// 0x008152C0
+bool W3DAssetManager::Render_Obj_Exists(const char *name)
```
tomsons26, line 178: this is return WW3DAssetManager::Find_Prototype(this, Str1) != 0;

src/w3d/renderer/assetmgr.cpp
duncanspumpkin: FontCharsList destruction is called from its destructor. Likewise Font3DDatas would have been called from its destructor if it was implemented. I can add a comment for the Font3DDatas if you want but I dont think FontCharsList makes sense.

duncanspumpkin:
Okay thats most of your comments addressed. I'm not 100% sure how those inlined functions are meant to work having looked at only the game.dat.

src/w3d/renderer/assetmgr.h
```diff
@@ -12,33 +12,57 @@
...
 {
 public:
+    W3DAssetManager();
     virtual ~W3DAssetManager();
     virtual bool Load_3D_Assets(const char *filename);
     virtual bool Load_3D_Assets(FileClass &asset_file);
     virtual void Free_Assets();
-    virtual void Release_Unused_Assets();
+    // 0x00814850
+    virtual void Release_Unused_Assets() { Release_Unused_Textures(); }
```
tomsons26: note that this called Release_Unused_Font3DDatas

tomsons26 (approved):
Nothing blows up, no actually missing assets. 👍

OmniBlade (approved):

PR #194, opened 2020-11-14
Implements SList and SLNode.

tomsons26:
Dunno yet if it matches whats in ZH fully, also not sure if SLNode needs its own header, we know it had it but does anything in SAGE use SLNode on its own?

### DIFF
--- a/src/w3d/lib/slist.h
+++ b/src/w3d/lib/slist.h
@@ -28,6 +28,7 @@ public:
     SLNode<T> *Tail() const { return m_tailNode; }
 
     SLNode<T> *Find_Node(T *data) const;
+    SLNode<T> *Find_Prior_Node(T *data) const;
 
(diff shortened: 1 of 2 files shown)


src/w3d/lib/slist.h
```diff
@@ -12,73 +13,262 @@
...
+    if (m_tailNode == nullptr) {
+        m_tailNode = temp;
+    }
+
+    return true;
+}
+
+template<class T> inline bool SList<T>::Add_Head(SList<T> &list)
+{
+    if (list.m_headNode == nullptr) {
```
duncanspumpkin: I suspect all of these `if (list.m_headNode == nullptr) ` are actually `if (Is_Empty())`

tomsons26:
Doesn't seem of use to even have it afaik so closing for now, branch will remain in case if needed so the pr can be reopened

PR #197, opened 2020-11-16
Implement all the CollisionMath functions that ZH and ZH WB actually use

jonwil:


src/hooker/setuphooks_zh.cpp
```diff
@@ -1148,4 +1149,50 @@ void Setup_Hooks()
     Hook_Any(0x008229B0, LightEnvironmentClass::Pre_Render_Update);
     Hook_Any(0x00822B60, LightEnvironmentClass::Add_Fill_Light);
     Hook_Any(0x00822CF0, LightEnvironmentClass::Calculate_Fill_Light);
+
+    // colmath.h
+    Hook_Method(
```
duncanspumpkin: These are all static functions shouldn't this be `Hook_Function`?

jonwil: Fixed.

PR #200, opened 2020-11-20
SegLineRendererClass

jonwil:


src/w3d/renderer/seglinerenderer.cpp
```diff
@@ -0,0 +1,893 @@
...
+    m_subdivisionLevel(0),
+    m_noiseAmplitude(0),
+    m_mergeAbortFactor(1.5f),
+    m_lastUsedSyncTime(that.m_lastUsedSyncTime),
+    m_currentUVOffset(0, 0),
+    m_UVOffsetDeltaPerMS(0, 0),
+    m_bits(DEFAULT_BITS),
+    m_vertexCount(0),
+    m_vertexBuffer(nullptr)
+{
```
tomsons26, line 62: should invoke op = here

src/w3d/renderer/seglinerenderer.cpp
```diff
@@ -0,0 +1,893 @@
...
+}
+
+void SegLineRendererClass::Set_Texture(TextureClass *texture)
+{
+    Ref_Ptr_Set(m_texture, texture);
+}
+
+TextureClass *SegLineRendererClass::Get_Texture() const
+{
+    m_texture->Add_Ref();
```
tomsons26: needs to be conditional on if the texture isn't nullptr

src/w3d/renderer/seglinerenderer.cpp
```diff
@@ -0,0 +1,893 @@
...
+}
+
+void SegLineRendererClass::Set_Current_UV_Offset(const Vector2 &offset)
+{
+    m_currentUVOffset = offset;
+}
+
+void SegLineRendererClass::Set_Texture_Tile_Factor(float factor)
+{
+    if (factor <= 50.0f) {
```
tomsons26: flip the case and add for over 50 the print WB has

src/w3d/renderer/seglinerenderer.cpp
```diff
@@ -0,0 +1,893 @@
...
+        Ref_Ptr_Release(material);
+    }
+
+    DX8Wrapper::Set_Transform(D3DTS_VIEW, view);
+#endif
+}
+
+void SegLineRendererClass::Scale(float scale)
+{
+    m_width = scale * m_width;
```
tomsons26: could just be *= then?

PR #207, opened 2020-11-23
W3dmouse 

duncanspumpkin:
Continuation from the W3DAssetManager PR. This implements the functions that couldn't be done due to W3DAssetManager in the W3DMouse class.

### DIFF
--- a/src/platform/w3dengine/client/w3dmouse.cpp
+++ b/src/platform/w3dengine/client/w3dmouse.cpp
@@ -22,7 +22,7 @@
 #include "scene.h"
 #include "texture.h"
 #include "w3ddisplay.h"
-#include <stdio.h>
+#include <cstdio>
 #ifdef PLATFORM_WINDOWS
 #include <winuser.h>
 #endif
@@ -34,7 +34,7 @@ MouseThreadClass W3DMouse::s_mouseThread("Thyme Mouse thread");
 bool W3DMouse::s_mouseThreadIsDrawing;
 HAnimClass *W3DMouse::s_W3DMouseAssets1[CURSOR_COUNT];
 RenderObjClass *W3DMouse::s_W3DMouseAssets2[CURSOR_COUNT];
-TextureBaseClass *W3DMouse::s_D3DMouseAssets[CURSOR_COUNT][21]; // TODO unsure on type
+TextureBaseClass *W3DMouse::s_D3DMouseAssets[CURSOR_COUNT][MAX_FRAMES]; // TODO unsure on type
 uint32_t W3DMouse::s_PolyMouseAssets[CURSOR_COUNT];
 CriticalSectionClass g_mouseCriticalSection;
 #endif
@@ -144,7 +144,7 @@ void W3DMouse::Set_Cursor(MouseCursor cursor)
 
             if (m_currentD3DCursor != cursor && s_mouseThreadIsDrawing == false) {
                 Release_D3D_Cursor_Texture(m_currentD3DCursor);
-                Load_D3D_Cursor_Texture(cursor);
+                Load_D3D_Cursor_Textures(cursor);
             }
 
             if (m_D3DCursorSurfaces[0] == nullptr) {
@@ -228,6 +228,7 @@ void W3DMouse::Set_Redraw_Mode(RedrawMode mode)
 {
     const auto current_cursor = m_currentCursor;
     Set_Cursor(MouseCursor::CURSOR_NONE);
+    m_currentRedrawMode = mode;
 
     switch (mode) {
         case RedrawMode::RM_WINDOWS:
@@ -266,11 +267,11 @@ void W3DMouse::Set_Redraw_Mode(RedrawMode mode)
             Init_D3D_Assets();
             Free_W3D_Assets();
             memset(s_PolyMouseAssets, 0, sizeof(s_PolyMouseAssets));
-            m_currentW3DCursor = MouseCursor::CURSOR_NONE;
-            m_currentPolyCursor = MouseCursor::CURSOR_NONE;
             if (s_mouseThread.Is_Running()) {
                 s_mouseThread.Execute();
             }
+            m_currentW3DCursor = MouseCursor::CURSOR_NONE;
+            m_currentPolyCursor = MouseCursor::CURSOR_NONE;
             break;
         default:
             break;
@@ -293,7 +294,7 @@ void W3DMouse::Free_D3D_Assets()
     }
 
     for (auto i = 0; i < CURSOR_COUNT; ++i) {
-        for (auto j = 0; j < 21; ++j) {
+        for (auto j = 0; j < MAX_FRAMES; ++j) {
             auto *d3d_surface = s_D3DMouseAssets[i][j];
             if (d3d_surface == nullptr) {
                 continue;
@@ -326,7 +327,7 @@ void W3DMouse::Init_D3D_Assets()
     }
 
     for (auto i = 0; i < CURSOR_COUNT; ++i) {
-        for (auto j = 0; j < 21; ++j) {
+        for (auto j = 0; j < MAX_FRAMES; ++j) {
             s_D3DMouseAssets[i][j] = nullptr;
             m_D3DCursorSurfaces[j] = nullptr;
         }
@@ -340,24 +341,26 @@ void W3DMouse::Free_W3D_Assets()
         auto *interface_scene = W3DDisplay::Get_3DInterfaceScene();
         auto &hanim = s_W3DMouseAssets1[i];
         auto &robj = s_W3DMouseAssets2[i];
-        if (interface_scene != nullptr) {
+
+        if (interface_scene != nullptr && robj != nullptr) {
             interface_scene->Remove_Render_Object(robj);
         }
-        if (robj != nullptr) {
 
+        if (robj != nullptr) {
             robj->Release_Ref();
+            robj = nullptr;
         }
-        robj = nullptr;
+
         if (hanim != nullptr) {
             hanim->Release_Ref();
+            hanim = nullptr;
         }
-        hanim = nullptr;
     }
 
     if (m_camera != nullptr) {
         m_camera->Release_Ref();
+        m_camera = nullptr;
     }
-    m_camera = nullptr;
 }
 
 // 0x007AD330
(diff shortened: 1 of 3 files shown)


xezon:
Is this ready for review?

duncanspumpkin:
Yes, not sure why it cant find sprintf_s though.

src/platform/w3dengine/client/w3dmouse.cpp
```diff
@@ -341,21 +427,52 @@ void W3DMouse::Release_D3D_Cursor_Texture(MouseCursor cursor)
...
+        auto *texture = asset_manager->Get_Texture(file_name);
+        s_D3DMouseAssets[cursor][0] = texture;
+        m_D3DCursorSurfaces[0] = texture->Get_Surface_Level(0);
+        return true;
+    }
+
+    for (auto i = 0; i < frames; ++i) {
+        auto &asset = s_D3DMouseAssets[cursor][i];
+        char file_name[64]{};
+        sprintf_s(file_name, "%s%04d.tga", texture_name, i);
```
xezon: Maybe this function version is not standard?

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/sprintf-s-sprintf-s-l-swprintf-s-swprintf-s-l?view=msvc-160
```
template <size_t size>
int sprintf_s(
   char (&buffer)[size],
   const char *format,
   ...
); // C++ only
```

Alternative would be to pass array element count of `file_name` in 2nd argument.

src/platform/w3dengine/client/w3dmouse.cpp
```diff
@@ -341,21 +427,52 @@ void W3DMouse::Release_D3D_Cursor_Texture(MouseCursor cursor)
...
+        sprintf_s(file_name, "%s.tga", texture_name);
+        auto *texture = asset_manager->Get_Texture(file_name);
+        s_D3DMouseAssets[cursor][0] = texture;
+        m_D3DCursorSurfaces[0] = texture->Get_Surface_Level(0);
+        return true;
+    }
+
+    for (auto i = 0; i < frames; ++i) {
+        auto &asset = s_D3DMouseAssets[cursor][i];
+        char file_name[64]{};
```
xezon, line 472: This will zero initialize entire buffer. Is this necessary here? We do `sprintf_s` next line so it looks not necessary.

src/platform/w3dengine/client/w3dmouse.cpp
duncanspumpkin: Yeah its non standard i forgot i had already written a note stating that it should be replaced with `snprintf`

src/platform/w3dengine/client/w3dmouse.cpp
duncanspumpkin: Its always safest to zero initialise everything especially when not in a hot loop like this.

xezon: Yeah if it is not performance critical then it certainly does not hurt. Just an option to remove it here because not necessary at all.

src/platform/w3dengine/client/w3dmouse.cpp
```diff
@@ -341,21 +428,55 @@ void W3DMouse::Release_D3D_Cursor_Texture(MouseCursor cursor)
...
+
+        s_D3DMouseAssets[cursor][0] = texture;
+        m_D3DCursorSurfaces[0] = texture->Get_Surface_Level(0);
+        m_D3DCursorSurfaceCount = 1; // Note: This just assumes it loaded fine
+        return true;
+    }
+
+    for (auto i = 0; i < frames; ++i) {
+        char file_name[64]{};
+        snprintf(file_name, 64, "%s%04d.tga", texture_name, i);
```
xezon: Instead of typing 64 again, you can also type `ARRAY_SIZE(file_name)`. This macro is from macros.h.

tomsons26:
Seems to work

src/platform/w3dengine/client/w3dmouse.cpp
```diff
@@ -304,28 +308,111 @@ void W3DMouse::Free_D3D_Assets()
...
+        return;
+    }
+
+    auto *asset_manager = W3DAssetManager::Get_Instance();
+    if (asset_manager == nullptr) {
+        return;
+    }
+
+    for (auto i = 0; i < CURSOR_COUNT; ++i) {
+        for (auto j = 0; j < 21; ++j) {
```
xezon: I suggest to add a `constexpr` or `static` variable for this 21 value and replace all related 21 numbers with that.

tomsons26: it appears to be a max frame count for the texture based cursor, should just be enum and all 21s replaced with that

src/platform/w3dengine/client/w3dmouse.cpp
```diff
@@ -304,28 +308,111 @@ void W3DMouse::Free_D3D_Assets()
...
+        }
+        if (robj != nullptr) {
+
+            robj->Release_Ref();
+        }
+        robj = nullptr;
+        if (hanim != nullptr) {
+            hanim->Release_Ref();
+        }
+        hanim = nullptr;
```
xezon: `hanim = nullptr;` can go into the scope above. Otherwise it is nullptr already.

src/platform/w3dengine/client/w3dmouse.cpp
```diff
@@ -304,28 +308,111 @@ void W3DMouse::Free_D3D_Assets()
...
+    }
+
+    if (m_camera != nullptr) {
+        m_camera->Release_Ref();
+    }
+    m_camera = nullptr;
```
xezon: `m_camera = nullptr;` can go into the scope above. Otherwise it is nullptr already.

src/platform/w3dengine/client/w3dmouse.cpp
```diff
@@ -304,28 +308,111 @@ void W3DMouse::Free_D3D_Assets()
...
+        auto &hanim = s_W3DMouseAssets1[i];
+        auto &robj = s_W3DMouseAssets2[i];
+        if (interface_scene != nullptr && robj != nullptr) {
+            interface_scene->Remove_Render_Object(robj);
+        }
+        if (robj != nullptr) {
+
+            robj->Release_Ref();
+        }
+        robj = nullptr;
```
xezon: `robj = nullptr;` can go into the scope above. Otherwise it is nullptr already.

src/platform/w3dengine/client/w3dmouse.cpp
```diff
@@ -304,28 +308,111 @@ void W3DMouse::Free_D3D_Assets()
...
-#endif
+    for (auto i = 0; i < CURSOR_COUNT; ++i) {
+        auto *interface_scene = W3DDisplay::Get_3DInterfaceScene();
+        auto &hanim = s_W3DMouseAssets1[i];
+        auto &robj = s_W3DMouseAssets2[i];
+        if (interface_scene != nullptr && robj != nullptr) {
+            interface_scene->Remove_Render_Object(robj);
+        }
+        if (robj != nullptr) {
+
```
xezon, line 348: Obsolete blank line is here.

src/platform/w3dengine/client/w3dmouse.cpp
```diff
@@ -13,12 +13,16 @@
...
+#include "assetmgr.h"
+#include "camera.h"
 #include "critsection.h"
 #include "dx8wrapper.h"
+#include "hanim.h"
 #include "rtsutils.h"
 #include "scene.h"
 #include "texture.h"
 #include "w3ddisplay.h"
+#include <stdio.h>
```
xezon: Better:
```c++
#include <cstdio>
```
stdio.h is for C.
cstdio includes stdio.h and is for C++. It has some extra code in it.

See
https://www.cplusplus.com/reference/cstdio/

src/platform/w3dengine/client/w3dmouse.cpp
```diff
@@ -341,21 +428,55 @@ void W3DMouse::Release_D3D_Cursor_Texture(MouseCursor cursor)
...
+        return true;
+    }
+
+    if (m_cursorInfo[cursor].frames == 0) {
+        return false;
+    }
+    auto *asset_manager = W3DAssetManager::Get_Instance();
+
+    auto *texture_name = m_cursorInfo[cursor].texture_name.Str();
+    auto frames = std::min(m_cursorInfo[cursor].frames, 21);
```
xezon: `CursorInfo::frames` is `int`. Shouldn't it be `uint32_t`? Can it ever be negative?

src/platform/w3dengine/client/w3dmouse.cpp
OmniBlade: Bare in mind when doing that, you need to use std:: prefix for the functions from that header as well, since cstdio is not required to expose them in the global namespace as well.

src/platform/w3dengine/client/w3dmouse.cpp
OmniBlade: sizeof(file_name) will also work nicely here.

src/platform/w3dengine/client/w3dmouse.cpp
duncanspumpkin: Due to STL port you cant `std::snprintf` so not sure what is best here. For now it looks like snprintf is being included in global scope so doesn't matter.

src/platform/w3dengine/client/w3dmouse.cpp
duncanspumpkin: Although probably true I'd prefer not to modify a structure without knowing all consumers.

Also in general people should tend to never use unsigned types most of the time it is not worth it and it can cause situations people dont expect.

tomsons26 (approved):
looks good

src/platform/w3dengine/client/w3dmouse.cpp
xezon: I don't agree with that assessment. An unsigned type makes its intent clear that it is not meant to be negative. This is a valid and reasonable contract.

duncanspumpkin: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es106-dont-try-to-avoid-negative-values-by-using-unsigned

From the cpp core guidelines.

xezon: Hmm yeah it makes valuable points. One needs to be fully aware about the unsigned nature when using it. Maybe we can inhibit the gsl::index concept into Thyme code to help robustness.

PR #208, opened 2020-11-23
Implements Render2DClass::Set_Texture(const char*)

duncanspumpkin:
Saw this whilst having a look at all functions related to W3DAssetManager. Nice and simple.

### DIFF
--- a/src/w3d/renderer/render2d.cpp
+++ b/src/w3d/renderer/render2d.cpp
@@ -14,6 +14,7 @@
  *            LICENSE
  */
 #include "render2d.h"
+#include "assetmgr.h"
 #include "texture.h"
 #include "w3d.h"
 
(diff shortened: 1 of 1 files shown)


src/w3d/renderer/render2d.cpp
```diff
@@ -93,10 +94,9 @@ void Render2DClass::Set_Coordinate_Range(const RectClass &range)
  */
 void Render2DClass::Set_Texture(const char *filename)
 {
-    // TODO Needs W3DAssetManager
-#ifdef GAME_DLL
-    Call_Method<void, Render2DClass, const char *>(PICK_ADDRESS(0x00809590, 0x004F3390), this, filename);
-#endif
+    auto *tex = W3DAssetManager::Get_Instance()->Get_Texture(filename);
```
tomsons26: passes MIP_LEVELS_1 actually too

duncanspumpkin: doh!

PR #209, opened 2020-11-26
SceneClass and SimpleSceneClass

duncanspumpkin:
Was wanting to get these done so that i can eventually do the 3 derived scenes 2d, 3d and interface.

### DIFF
--- a/src/w3d/renderer/scene.cpp
+++ b/src/w3d/renderer/scene.cpp
@@ -15,6 +15,7 @@
 #include "scene.h"
 #include "camera.h"
 #include "chunkio.h"
+#include "coltest.h"
 #include "dx8wrapper.h"
 #include "light.h"
 #include "lightenv.h"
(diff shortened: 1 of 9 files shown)


src/w3d/renderer/scene.cpp
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+#ifdef BUILD_WITH_D3D8
+    for (auto i = 0; i < LightEnvironmentClass::MAX_LIGHTS; ++i) {
+        DX8Wrapper::Set_Light(i, nullptr);
+    }
+#endif
+
+    auto *light_environment = rinfo.m_lightEnvironment;
+    if (light_environment != nullptr) {
```
tomsons26: should assert rinfo.light_environment==NULL before this and this if case should be  if ( !light_environment  ) as it makes a fallback light if one isn't set

src/w3d/renderer/scene.cpp
```diff
@@ -0,0 +1,264 @@
...
+            break;
+    }
+}
+
+// 0x00832DF0
+float SimpleSceneClass::Compute_Point_Visibility(RenderInfoClass &rinfo, const Vector3 &point)
+{
+    CastResultStruct res;
+    auto camera_position = rinfo.m_camera.Get_Position();
+    LineSegClass ray{ camera_position, point };
```
tomsons26: should be LineSegClass ray(camera_position, point);, if the ctor isn't there ask, ill provide the code

src/w3d/renderer/scene.cpp
```diff
@@ -0,0 +1,264 @@
...
+        }
+    }
+    m_releaseList.Reset_List();
+}
+
+// 0x00832B30
+void SimpleSceneClass::Remove_All_Render_Objects()
+{
+    for (auto *robj = m_renderList.Remove_Head(); robj != nullptr; robj = m_renderList.Remove_Head()) {
+        robj->Notify_Removed(this);
```
tomsons26: original calls SceneClass::Remove_Render_Object()

src/w3d/renderer/scene.cpp
duncanspumpkin: It is that? do you mean `LineSegClass ray{rinfo.m_camera, point};`

tomsons26: no, () there's a Vector3, Vector3 ctor

OmniBlade:
@tomsons26 Are there any outstanding changes you were wanting for this PR or does it just need rebasing for merge?

tomsons26:
> @tomsons26 Are there any outstanding changes you were wanting for this PR or does it just need rebasing for merge?

can't remember at this point, only that i don't like that line seg ctor init, its supposed to call the actual ctor not jam the values in.

duncanspumpkin:
It is calling the ctor. It's not an aggregate initialiser if that's what you are getting at. As soon as you define a constructor you can't use aggregate initialisation. Braces and parens do the exact same thing in this context.

src/w3d/renderer/scene.cpp
```diff
@@ -0,0 +1,265 @@
...
+    auto *light_environment = rinfo.m_lightEnvironment;
+    captainslog_dbgassert(light_environment == nullptr, "Light Environment already exists");
+    if (light_environment == nullptr) {
+        static LightEnvironmentClass s_lightEnv;
+        s_lightEnv.Reset({ 0, 0, 0 }, m_ambientLight);
+        for (auto iter = m_lightList.Iterator(); iter; ++iter) {
+            s_lightEnv.Add_Light(*(static_cast<LightClass *>(iter.Get_Obj())));
+        }
+        rinfo.m_camera.Validate_Transform();
+        s_lightEnv.Pre_Render_Update(rinfo.m_camera.Get_Transform_No_Validity_Check());
```
tomsons26: this would be Get_Transform() since Validate_Transform is called above

duncanspumpkin:
I've rebased and switched back from {} to () and made that other change.

src/w3d/renderer/scene.cpp
```diff
@@ -0,0 +1,264 @@
...
+#ifdef BUILD_WITH_D3D8
+    for (auto i = 0; i < LightEnvironmentClass::MAX_LIGHTS; ++i) {
+        DX8Wrapper::Set_Light(i, nullptr);
+    }
+#endif
+
+    auto *light_environment = rinfo.m_lightEnvironment;
+    captainslog_dbgassert(light_environment == nullptr, "Light Environment already exists");
+    if (light_environment == nullptr) {
+        static LightEnvironmentClass s_lightEnv;
```
OmniBlade: local statics should be prefixed with _ but otherwise named the same as local variables.

OmniBlade (changes requested):
Just need the local static renaming but otherwise I think this is good to merge.

OmniBlade (approved):

PR #210, opened 2020-11-29
Fix DX8Renderer

jonwil:


src/w3d/renderer/dx8renderer.cpp
```diff
@@ -724,17 +724,41 @@ unsigned int DX8TextureCategoryClass::Add_Mesh(Vertex_Split_Table &split_table,
                     if (Compare_Materials(split_table.Peek_Material(i, pass), m_material)) {
                         if (split_table.Peek_Shader(i, pass) == m_shader) {
                             unsigned short index = vertex_offset + polys[i].I;
-                            std::clamp(index, vmin, vmax);
+
+                            if (vmin >= index) {
```
duncanspumpkin: This is updating the min and max you would normally do this with a function to make it clearer:
```
vmin = std::min(index, vmin);
vmax = std::max(index, vmax);
```

duncanspumpkin:
Can confirm that this fixes the crash and also fixes the change in colour of the particle cannon.

src/w3d/renderer/dx8renderer.cpp
```diff
@@ -724,17 +724,26 @@ unsigned int DX8TextureCategoryClass::Add_Mesh(Vertex_Split_Table &split_table,
                     if (Compare_Materials(split_table.Peek_Material(i, pass), m_material)) {
                         if (split_table.Peek_Shader(i, pass) == m_shader) {
                             unsigned short index = vertex_offset + polys[i].I;
-                            std::clamp(index, vmin, vmax);
+
```
duncanspumpkin, line 727: Just a thought but couldn't this be turned into a for loop?
Not sure if that makes it more readable.
```
for (auto j = 0; j < 3; ++j){
    uint16_t index = vertex_offset + polys[i][j];
    vmin = std::min(index, vmin);
    vmax = std::max(index, vmax);
    *indices++ = index;
}
```

PR #211, opened 2020-11-29
Fix AI not building units

duncanspumpkin:
Missing concat from `0x0051D0CA`


tomsons26:
Should fix #100 completely now

PR #212, opened 2020-11-29
Switch Ref_Ptr_Set arguments

duncanspumpkin:
Switches the Ref_Ptr_Set arguments and introduces a bit more const. Unsure if the const part is a good idea still humming and harring over that.


src/w3d/renderer/dx8wrapper.cpp
```diff
@@ -1244,7 +1244,7 @@ void DX8Wrapper::Set_Vertex_Buffer(const VertexBufferClass *vb, int number)
     }
 
     VertexBufferClass *v = const_cast<VertexBufferClass *>(vb);
```
duncanspumpkin, line 1246: Imho I think we should not touch the v variable after we do Ref_Ptr_Set i see it as being invalid to use after this. Its not but I feel we should avoid using it.

PR #215, opened 2020-12-06
Fixes various issues in Mouse class and implements more functions of it.

tomsons26:


src/game/client/input/mouse.cpp
```diff
@@ -715,3 +741,30 @@ void Mouse::Parse_Cursor_Definitions(INI *ini)
         ini->Init_From_INI(&g_theMouse->m_cursorInfo[g_theMouse->Get_Cursor_Index(tok)], _cursor_parsers);
     }
 }
+
+bool Mouse::Is_Click(ICoord2D *coord1, ICoord2D *coord2, unsigned val1, unsigned val2) const
```
duncanspumpkin: Val 1 and 2 are probably time 1 and time 2

src/platform/input/win32mouse.cpp
```diff
@@ -108,7 +108,7 @@ void Win32Mouse::Set_Visibility(bool visibility)
 }
 
 // 0x0073BE20
-bool Win32Mouse::Get_Mouse_Event(MouseIO *io, int8_t unk)
+uint8_t Win32Mouse::Get_Mouse_Event(MouseIO *io, int8_t unk)
 {
```
duncanspumpkin, line 112: Could you do a check and see if the false should be changed to 0xff

src/game/client/input/mouse.cpp
```diff
@@ -74,10 +77,29 @@ const char *Mouse::s_cursorNames[] = {
...
+    image_name.Clear();
+    w3d_model_name.Clear();
+    w3d_anim_name.Clear();
+}
+
 /**
  * 0x004029D0
  */
 Mouse::Mouse() :
+    m_cursorInfo(),
```
OmniBlade: No need to have this in the initialiser list, the array will always be default initialised by calling the default constructor for each item in the array.

OmniBlade (approved):
You can remove the array initialisation from the Mouse constructor, but since it seems to compile as is I suppose it doesn't really matter. Looks good otherwise, subject to the other suggestions being implemented as well.

OmniBlade (approved):

PR #216, opened 2020-12-07
Fixes various issues in Keyboard class.

tomsons26:


src/game/client/input/keyboard.cpp
```diff
@@ -134,17 +134,27 @@ void Keyboard::Create_Stream_Messages()
     if (g_theMessageStream != nullptr) {
         GameMessage *msg = nullptr;
 
-        for (int i = 0; m_keys[i].key != 0; ++i) {
-            if (m_keys[i].state & KEY_DOWN) {
+        KeyboardIO *ptr = Get_First_Key();
+        while (ptr->key != 0) {
```
duncanspumpkin: You should keep this as a for loop keyboard ptr is never used outside of this loop. I'd suggest changing the name to `keyIO` or something more descriptive.

src/game/client/input/keyboard.cpp
```diff
@@ -194,25 +204,25 @@ void Keyboard::Reset_Keys()
 /**
  * 0x0040A460
  */
-bool Keyboard::Is_Shift()
+bool Keyboard::Is_Shift() const
 {
-    return m_modifiers & MODIFIER_SHIFT || m_modifiers & MODIFIER_SHIFTEX;
+    return m_modifiers & MODIFIER_LSHIFT || m_modifiers & MODIFIER_RSHIFT || m_modifiers & MODIFIER_SHIFTEX;
```
duncanspumpkin, line 205: I always find it best to add brackets to these to prevent any issues with operator precedence. In this instance there is no issues though.

src/game/client/input/keyboard.cpp
```diff
@@ -284,7 +294,7 @@ void Keyboard::Update_Keys()
  */
 wchar_t Keyboard::Translate_Key(wchar_t key)
 {
-    if (key > 256) {
+    if (key > 255) {
```
duncanspumpkin: `>= KEY_COUNT`

OmniBlade:
I'm willing to approve this once @duncanspumpkin suggestions are addressed.

src/game/client/input/keyboard.cpp
OmniBlade: I agree with @duncanspumpkin on this one, it would be much better as a for loop.
`for (KeyboarIO *keyio = Get_First_Key; keyio->key != 0; ++keyio)`

src/game/client/input/keyboard.cpp
```diff
@@ -134,17 +134,27 @@ void Keyboard::Create_Stream_Messages()
...
 
-        for (int i = 0; m_keys[i].key != 0; ++i) {
-            if (m_keys[i].state & KEY_DOWN) {
+        KeyboardIO *ptr = Get_First_Key();
+        while (ptr->key != 0) {
+            if (ptr->state & KEY_DOWN) {
                 msg = g_theMessageStream->Append_Message(GameMessage::MSG_RAW_KEY_DOWN);
-            } else if (m_keys[i].state & KEY_UP) {
+                if (msg == nullptr) {
+                    captainslog_debug("Unable to append key down message to stream");
```
OmniBlade: These debug logs look like they should be debug asserts instead.

src/game/client/input/keyboard.cpp
OmniBlade: @duncanspumpkin is right, use the constant we already defined, `>= 256` is the same as `> 255` logically.

OmniBlade (changes requested):

OmniBlade (approved):

PR #219, opened 2020-12-07
Adds missing DisplayString virtual.

tomsons26:


OmniBlade (approved):
Oops, looks like that one was my fault :D

PR #230, opened 2020-12-15
Definition of Object class

jonwil:


src/game/logic/object/object.h
```diff
@@ -0,0 +1,620 @@
...
+    bool Get_Apply_Battle_Plan_Bonuses() const { return m_applyBattlePlanBonuses; }
+    PathfindLayerEnum Get_Destination_Layer() const { return m_destinationLayer; }
+    void Get_Formation_Offset(Coord2D *c) { *c = m_formationOffset; }
+    bool Get_Single_Use_Command() const { return m_singleUseCommand; }
+
+    void Clear_Status(BitFlags<OBJECT_STATUS_COUNT> bits) { return Set_Status(bits, false); }
+    bool Clear_Script_Status(ObjectScriptStatusBit bit) { return Set_Script_Status(bit, false); }
+
+    Module *Find_Module(NameKeyType type) const;
+    DamageModule *Find_Damage_Module(NameKeyType type) const { return (DamageModule *)Find_Module(type); }
```
duncanspumpkin, line 286: Don't use a c style cast for this it could be subtly broken if damage module uses multiple inheritance. Use a static cast instead

jonwil: The original code did the equivalent of a C style cast so its perfectly OK for our clone to do the same.

duncanspumpkin: I've already had one bug in this project that was due to this just wanting to stop it from happening again. If you can guarantee that DamageModule doesn't / will never use multiple inheritance then yes its fine.

PR #247, opened 2021-06-10
Pre script engine fixes/implementations.

OmniBlade:


tomsons26:
Could this be rebased?

src/game/common/system/gametype.h
```diff
@@ -184,6 +184,24 @@ enum StealthLookType
     STEALTH_LOOK_0,
 };
 
+enum GameMode
+{
+    GAME_SINGLE_PLAYER,
```
xezon: Or `GAME_CAMPAIGN`

src/game/common/system/gametype.h
```diff
@@ -184,6 +184,24 @@ enum StealthLookType
...
+    GAME_LAN,
+    GAME_SKIRMISH,
+    GAME_REPLAY,
+    GAME_SHELL,
+    GAME_INTERNET,
```
xezon: Or `GAME_ONLINE`

src/game/common/system/gametype.h
```diff
@@ -184,6 +184,24 @@ enum StealthLookType
...
+    GAME_NONE,
```
xezon: For clarification, can add a comment here that this applies to menu without shell map.

src/game/logic/scriptengine/script.h
```diff
@@ -42,10 +42,14 @@ class Script : public MemoryPoolObject, public SnapShot
     OrCondition *Get_Condition() { return m_condition; }
     ScriptAction *Get_Action() { return m_action; }
     ScriptAction *Get_False_Action() { return m_actionFalse; }
+    float Get_Total_Exec_Time() const { return m_totalExecTime; }
+    int Get_Evalulation_Count() const { return m_evalCount; }
+    Utf8String Get_Name() const { return m_scriptName; }
```
xezon: `const Utf8String&`

src/game/logic/system/gamelogic.cpp
```diff
@@ -64,7 +64,7 @@ bool GameLogic::Is_Intro_Movie_Playing()
 
 int GameLogic::Rebalance_Parent_Sleepy_Update(int update)
 {
-    captainslog_dbgassert(update >= 0 && update < m_sleepingUpdateModules.size(), "bad sleepy idx");
+    captainslog_dbgassert(update >= 0 && (unsigned)update < m_sleepingUpdateModules.size(), "bad sleepy idx");
```
xezon: `static_cast<size_t>(update)`

src/game/logic/system/gamelogic.h
```diff
@@ -126,7 +108,7 @@ class GameLogic : public SubsystemInterface, public SnapShot
             return nullptr;
         }
 
-        if (id >= m_objectLookupTable.size()) {
+        if ((unsigned)id >= m_objectLookupTable.size()) {
```
xezon: `static_cast<size_t>(id)`

src/game/common/system/gametype.h
OmniBlade: I don't intend to rename these items in this PR as it affects additional files that aren't really related.

src/game/logic/scriptengine/script.h
```diff
@@ -44,7 +44,7 @@ class Script : public MemoryPoolObject, public SnapShot
     ScriptAction *Get_False_Action() { return m_actionFalse; }
     float Get_Total_Exec_Time() const { return m_totalExecTime; }
     int Get_Evalulation_Count() const { return m_evalCount; }
-    Utf8String Get_Name() const { return m_scriptName; }
+    const Utf8String &Get_Name() const { return m_scriptName; }
```
xezon: On second thought... does the original return a copy? Should we keep the copy then? Returning reference is different - breaks the contract - and hooking that function won't be the same I think.

duncanspumpkin: We wouldn't hook a getter like this. Most of them are inlined anyway. I think it would be fine to return a const ref as it's not a virtual.

OmniBlade: If you guys think it all looks good I'll do a final rebase and merge this as soon as I'm able.

xezon (approved):

PR #258, opened 2021-10-09
Add CMake and build scripts for Windows and Visual Studio

xezon:
Thyme repository has no build scripts and tools ready for pickup, other than the CMakeLists.txt files. This change adds a stripped down CMake + build scripts for Windows 10 + Visual Studio 2017 + Visual Studio 2019. With this developer can generate solution comfortably without having to figure out build steps on Windows first.

This is better for on-boarding new programmers (like myself).

PR #259, opened 2021-10-10
Add cmake build scripts for Windows and Visual Studio

xezon:
Similar to
* #258
but without cmake and 7zip added directly. It will be downloaded by script.


tools/build/win_x86/create_vs2019_solution.bat
```diff
@@ -0,0 +1,24 @@
...
+@call "%ThisDir%\scripts\get_tools.bat"
+
+@if not exist "%SolutionVS2019Dir%" mkdir "%SolutionVS2019Dir%"
+cd /d "%SolutionVS2019Dir%"
+:: Generate solution files
+"%CMakeDir%\bin\cmake.exe" ^
+    -G"Visual Studio 16 2019" ^
+    -A Win32 ^
+    -S "%ProjectDir%" ^
+    -DCMAKE_VS_INCLUDE_INSTALL_TO_DEFAULT_BUILD=1
```
xezon: I decided to enable `CMAKE_VS_INCLUDE_INSTALL_TO_DEFAULT_BUILD` by default here, because I see no reason it turned off. Maybe it could actually be in CMakeLists.txt itself. Let me know.

.gitignore
```diff
@@ -1,6 +1,7 @@
 *.user
 *.ncb
 /build*
+/.solution
```
xezon, line 4: `.solution/` rather?

OmniBlade: Could be, to be honest I'm not actually sure if that causes it only to be relative to the top level folder or ignore it whatever level it occurs on?

tools/build/win_x86/create_vs2019_solution.bat
OmniBlade: Its fine where it is, if someone was generating the solution other than by these bat files I'd rather it be opt in than opt out.

OmniBlade (approved):
Looks okay to me, though I'm far from an expert in windows batch files. If you want to write a short guide on how to use these I could add it to the wiki once its merged?

.gitignore
xezon: I checked. `/.solution` is good. Means in current directory.

tools/build/win_x86/create_vs2019_solution.bat
xezon: Ok.

PR #265, opened 2021-10-12
Creates a thyme_standalone_obj library.

OmniBlade:
Creates an object library to use for linking any standalone targets that
need to use code from the engine.

Intended use is to link standalone game, editor and misc modding tools.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -206,6 +206,7 @@ set(GAMEENGINE_SRC
     platform/w3dfunctionlexicon.cpp
     platform/win32bigfile.cpp
     platform/win32bigfilesystem.cpp
+    platform/win32gameengine.cpp
     platform/win32localfile.cpp
     platform/win32localfilesystem.cpp
     platform/w3dengine/client/w3ddebugdisplay.cpp
@@ -314,7 +315,6 @@ set(GAMEENGINE_SRC
 # Game binary only source
 set(GAMEENGINE_GAME_SRC
     game/main.cpp
-    platform/win32gameengine.cpp
 )
 
 set(GAMEENGINE_WB_SRC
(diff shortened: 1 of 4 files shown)


src/CMakeLists.txt
```diff
@@ -434,37 +430,44 @@ if(DINPUT8_FOUND)
     list(APPEND GAME_COMPILE_OPTIONS -DBUILD_WITH_DINPUT)
 endif()
 
+# Static library for standalone builds and linking tools.
+if(STANDALONE OR BUILD_TOOLS)
+    add_library(thyme_standalone_obj OBJECT ${GAMEENGINE_SRC} ${GAMEENGINE_HEADERS})
```
xezon: In Windows this would be a lib file right? Is `standalone object` a proper term for this? `thyme_library` or `thyme_lib`, at least in the realms of Windows, would be more fitting.

src/CMakeLists.txt
```diff
@@ -434,37 +430,44 @@ if(DINPUT8_FOUND)
...
+    target_compile_definitions(thyme_standalone_obj PUBLIC $<$<CONFIG:DEBUG>:GAME_DEBUG> $<$<CONFIG:DEBUG>:GAME_DEBUG_STRUCTS>)
+endif()
+
 if(STANDALONE)
-    add_executable(thyme ${GAMEENGINE_GAME_SRC} ${GAMEENGINE_SRC} ${GAMEENGINE_HEADERS})
-    target_include_directories(thyme PRIVATE ${GAMEENGINE_INCLUDES})
+    add_executable(thyme ${GAMEENGINE_GAME_SRC})
+    target_link_libraries(thyme PRIVATE thyme_standalone_obj)
 else()
     add_library(thyme SHARED ${HOOKER_SRC} ${HOOKER_GAME_SRC} ${GAMEENGINE_GAME_SRC} ${GAMEENGINE_SRC} ${GAMEENGINE_HEADERS})
```
xezon: We could also rename `thyme` to `thyme_shared` and `thyme_standalone_obj` to `thyme_static`. That would be quite explicit and easy to understand.

src/CMakeLists.txt
OmniBlade: This actually just builds a collection of objects to link into whatever, so while its a "library" in cmake terms, on disk it never is.

src/CMakeLists.txt
OmniBlade: Could do this and infact is probably worth putting a bit more work in so that on windows both standalone and dll can be built at the same time.

xezon: I would expect that tools just link against thyme.lib and have access to all thyme game headers, so they can just use the game code in whatever way they want.

PR #267, opened 2021-10-13
Replace padding comments and declarations with padding macro

xezon:
The purpose of this change is to
1. Make all pad bytes explicit
2. Document pad bytes consistently
3. Allow pads to be removed by preprocessor when needed

### DIFF
--- a/deps/baseconfig/src/macros.h
+++ b/deps/baseconfig/src/macros.h
@@ -287,4 +287,13 @@
 #define DEFINE_ENUMERATION_OPERATORS(ENUMTYPE) // NOP, C allows these operators.
 #endif // __cplusplus
 
+#ifndef BYTE_PAD
+#define BYTE_PAD_TOKEN(a, b, c) a##_##b##_##c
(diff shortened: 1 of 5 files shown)


deps/baseconfig/src/macros.h
```diff
@@ -287,4 +287,13 @@
...
 #endif // __cplusplus
 
+#ifndef BYTE_PAD
+#define BYTE_PAD_TOKEN(a, b, c) a##_##b##_##c
+#define BYTE_PAD_TOKEN2(a, b, c) BYTE_PAD_TOKEN(a, b, c)
+#define BYTE_PAD(size) char BYTE_PAD_TOKEN2(pad, size, __LINE__)[size]
+#ifndef TYPE_PAD
+#define TYPE_PAD(type, size) type BYTE_PAD_TOKEN2(pad, size, __LINE__)[size]
+#endif // TYPE_PAD
+#endif // BYTE_PAD
```
xezon, line 297: We could guard these by Win32 + Not Standalone.

PR #272, opened 2021-10-15
SortingRendererClass work

jonwil:


src/w3d/renderer/dx8wrapper.h
```diff
@@ -781,3 +786,60 @@ inline void DX8Wrapper::Set_Index_Buffer_Index_Offset(unsigned offset)
...
+    }
+
+    for (int i = 0; i < VERTEX_BUFFERS; i++) {
+        if (s_renderState.vertex_buffers[i]) {
+            s_renderState.vertex_buffers[i]->Release_Engine_Ref();
+        }
+    }
+
+    s_renderState = state;
+    s_renderStateChanged = 0xffffffff;
```
duncanspumpkin, line 808: L787 used this like a set of flags, one of them looks wrong unsure which.

tomsons26: Checked pseudo, it's correct, i guess it's a catch all for all states have changed

tomsons26:
Break up the lines so its not one monolith, new line after multiline, new line when context/operation changes( it sets polygon variables, then sets vertex variables - a context change), space out asserts

tomsons26:
Works, seems good otherwise at least as far as i can tell

OmniBlade (approved):

PR #274, opened 2021-10-16
Fix string literals for wchar_t character set

xezon:
utf16 was used for wide char strings. This does not look correct. This change fixes this in all of Thyme.

### DIFF
--- a/deps/baseconfig/src/unichar.h
+++ b/deps/baseconfig/src/unichar.h
@@ -16,9 +16,14 @@
 #define BASE_UNICHAR_H
 #include <wchar.h>
 
+#ifdef U_CHAR
+#error U_CHAR has been defined before
+#endif
+
 #ifdef BUILD_WITH_ICU
 #include <unicode/uchar.h>
 typedef UChar unichar_t;
+#define U_CHAR(str) (u##str)
 #elif defined _WIN32
 typedef wchar_t unichar_t;
 #define u_strlen wcslen
@@ -31,6 +36,7 @@ typedef wchar_t unichar_t;
 #define u_tolower towlower
 #define u_strchr(x, y) wcschr(x, y)
 #define U_COMPARE_CODE_POINT_ORDER 0x8000
+#define U_CHAR(str) (L##str)
 #else
 #error This platform requires ICU to support unichar.
 #endif
--- a/src/game/client/gametext.cpp
+++ b/src/game/client/gametext.cpp
@@ -203,22 +203,22 @@ void GameTextManager::Translate_Copy(unichar_t *out, char *in)
 
             switch (current) {
                 case '\\':
-                    *out++ = u'\\';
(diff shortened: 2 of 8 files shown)


deps/baseconfig/src/unichar.h
```diff
@@ -31,6 +36,7 @@ typedef wchar_t unichar_t;
 #define u_tolower towlower
 #define u_strchr(x, y) wcschr(x, y)
 #define U_COMPARE_CODE_POINT_ORDER 0x8000
+#define U_CHAR(str) (L##str)
```
xezon, line 39: Notice how `unichar_t` is `wchar_t` here, therefore we should not use `u` string literal but `L`.

See:
https://docs.microsoft.com/en-us/cpp/cpp/string-and-character-literals-cpp?view=msvc-160

Converting from utf16 to wide char would have to be performed via `codecvt_utf16`. See:
https://en.cppreference.com/w/cpp/locale/codecvt_utf16

However, conversion is not necessary here, because we just work with wide char and not utf16.

src/game/client/gametext.cpp
```diff
@@ -679,7 +679,7 @@ GameTextManager::GameTextManager() :
     m_noStringList(nullptr),
     m_useStringFile(true),
     m_language(LANGUAGE_ID_US),
-    m_failed((const unichar_t *)u"***FATAL*** String Manager failed to initialize properly"),
+    m_failed(U_CHAR("***FATAL*** String Manager failed to initialize properly")),
```
xezon, line 682: `(const unichar_t *)` here is a `reinterpret_cast`. utf16 string literal is cast to wide char string, which is undefined behavior. For English character set it probably is no problem, as they likely share the same characters, but if you would use more exotic symbols, then this will likely go wrong.

src/game/common/system/unicodestring.cpp
```diff
@@ -301,7 +301,7 @@ void Utf16String::Trim()
         return;
     }
 
-    for (int i = u_strlen(Peek()) - 1; i >= 0; --i) {
+    for (int i = static_cast<int>(u_strlen(Peek())) - 1; i >= 0; --i) {
```
xezon, line 304: I added a `static_cast` for good measure here to make it more obvious that this is not standard usage.

src/game/client/gametext.cpp
```diff
@@ -276,19 +276,19 @@ void GameTextManager::Strip_Spaces(unichar_t *buffer)
     unichar_t *putp = buffer;
 
     unichar_t current = *getp++;
-    unichar_t last = '\0';
+    unichar_t last = U_CHAR('\0');
```
xezon, line 279: This function was using wrong character set. No problem once again, because Ascii characters.

xezon:
You can squash this change on merge.

OmniBlade (approved):

PR #276, opened 2021-10-16
Add templates for strlcpy, strlcat, u_strlcpy to help avoid misuse

xezon:
These templates make passing size obsolete, because the template will deduce the size from the passed array type. This helps avoid making errors when passing array sizes.

### DIFF
--- a/src/game/network/lanapi.cpp
+++ b/src/game/network/lanapi.cpp
@@ -98,7 +98,7 @@ void LANAPI::Init()
 
     if (!GetUserNameA(user_name, &un_size)) {
         if (getenv("USERNAME") != nullptr) {
-            strlcpy_t(user_name, getenv("USERNAME"));
+            strlcpy_tpl(user_name, getenv("USERNAME"));
         } else {
             strcpy(user_name, "unknown");
         }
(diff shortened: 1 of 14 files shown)


src/game/network/lanapi.cpp
```diff
@@ -229,7 +229,7 @@ void LANAPI::Request_Game_Join_Direct(uint32_t addr)
         msg.message_type = LANMessage::MSG_REQUEST_GAME_INFO;
         Fill_In_Message(&msg);
         msg.direct_join.addr = Get_Local_IP();
-        u_strlcpy(msg.direct_join.name, (const unichar_t *)m_name.Str(), m_name.Get_Length() + 1);
```
xezon: This looks like it could be error. `m_name.Get_Length() + 1` could maybe be longer than target buffer and cause overflow.

xezon: I added a comment to this bugfix now so it is properly documented.

xezon:
I think I need to add a guard for `__cplusplus`
Edit: it's already there.

OmniBlade (approved):

PR #277, opened 2021-10-17
Add THYME_FEATURES and THYME_FIXES cmake options

xezon:
I need these very much for gating & documenting changed that deviate from the original.


xezon:
Just realizing that this change is not complete.

xezon:
As discussed internally, we will use `FEATURE` and `BUGFIX` comments for now to keep it simple.

PR #281, opened 2021-10-18
Animation work

jonwil:


src/w3d/renderer/w3d_file.h
```diff
@@ -338,3 +357,65 @@ enum
...
+};
+
+struct W3dAnimChannelStruct
+{
+    uint16_t FirstFrame;
+    uint16_t LastFrame;
+    uint16_t VectorLen;
+    uint16_t Flags;
+    uint16_t Pivot;
+    uint16_t pad;
```
xezon, line 375: Should this be a comment?

src/w3d/renderer/w3d_file.h
```diff
@@ -338,3 +357,65 @@ enum
...
+};
+
+struct W3dAdaptiveDeltaAnimChannelStruct
+{
+    uint32_t NumFrames;
+    uint16_t Pivot;
+    uint8_t VectorLen;
+    uint8_t Flags;
+    float Scale;
+    uint32_t Data[1];
```
xezon, line 420: They really seem to like this method...

src/w3d/renderer/motchan.cpp
```diff
@@ -0,0 +1,664 @@
...
+    int count2;
+
+    int count = 0;
+    int rightIdx = m_numTimeCodes - 2;
+
+    if (timecode >= (m_data[m_lastTimeCodeIdx] & 0x7FFFFFFF)) {
+        result = m_lastTimeCodeIdx;
+    } else {
+        for (;;) {
+            for (;;) {
```
xezon, line 329: I wonder if we can make a pass to change all `for(;;)` to `while(true)` and suppress the compile warning for it.

src/w3d/renderer/motchan.cpp
```diff
@@ -0,0 +1,664 @@
...
+
+        for (int i = 0; i < m_vectorLen; ++i) {
+            setvec[i] = data[i];
+        }
+    } else {
+        int index2 = m_packetSize + index;
+        unsigned int val = m_data[index2];
+
+        if (GameMath::Fast_Is_Float_Positive(val)) {
+            float f1 = (m_data[index] & 0x7FFFFFFF);
```
xezon, line 214: A lot of 7FFFFFFF are used in this file. I wonder if it is worthwhile to swap it for `INT_MAX` or `std::numeric_limits<int>::max()`

src/w3d/renderer/hrawanim.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+    cload.Close_Chunk();
+
+    if (header.Version < 0x30000) {
+        pre30 = true;
+    }
+
+    strcpy(m_name, header.HierarchyName);
+    strcat(m_name, ".");
+    strcat(m_name, header.Name);
```
xezon, line 66: I created a task report to look at usage of these functions later
* #282

src/w3d/renderer/w3d_file.h
tomsons26: In this case no, its explicitly there, thats how its stored in the file. Entire w3d_file info is how its stored in w3d file.

src/w3d/renderer/motchan.cpp
tomsons26: Yeah this is fishy

src/w3d/renderer/motchan.cpp
tomsons26: should be INT_MAX, unless its a bitmask, then we need to figure out what to do there

src/w3d/renderer/hanim.h
```diff
@@ -46,7 +47,7 @@ class HAnimClass : public RefCountClass, public HashableClass
     virtual bool Has_Z_Translation(int pividx) { return true; }
     virtual bool Has_Rotation(int pividx) { return true; }
     virtual bool Has_Visibility(int pividx) { return true; }
-    virtual int Class_ID() { return -1 /*CLASSID_UNKNOWN*/; }
```
tomsons26, line 49: should be CLASSID_UNKNOWN, was commented out while we didn't have the enum

src/w3d/renderer/hcanim.cpp
```diff
@@ -0,0 +1,374 @@
...
+
+bool HCompressedAnimClass::read_bit_channel(ChunkLoadClass &cload, TimeCodedBitChannelClass **newchan)
+{
+    *newchan = new TimeCodedBitChannelClass();
+    return (*newchan)->Load_W3D(cload);
+}
+
+void HCompressedAnimClass::add_bit_channel(TimeCodedBitChannelClass *newchan)
+{
+    if (!newchan->Get_Type()) {
```
tomsons26: check enum

src/w3d/renderer/motchan.cpp
xezon: Could also use a `constexpr int sign_bit_mask = 0x7FFFFFFF` in this file to help teach the intent.

tomsons26 (approved):

PR #284, opened 2021-10-18
Add natvis for Utf8String and Utf16String

xezon:
natvis


src/CMakeLists.txt
```diff
@@ -326,6 +326,15 @@ set(GAMEENGINE_WB_SRC
     editor/editmain.cpp
 )
 
+
+set(GAMEENGINE_NATVIS "")
+
+if (MSVC_IDE)
+    list(APPEND GAMEENGINE_NATVIS
+        natvis/thyme.natvis
```
xezon: Maybe put file in other folder? I would have put it in root/tools/natvis, but then I would have needed to use ../tools/natvis here.

src/CMakeLists.txt
```diff
@@ -418,11 +427,11 @@ endif()
 
 # Static library for standalone builds and linking tools.
 if(STANDALONE OR BUILD_TOOLS)
-    add_library(thyme_lib STATIC ${GAMEENGINE_SRC} ${GAMEENGINE_HEADERS} game/crashhandler.cpp)
+    add_library(thyme_lib STATIC ${GAMEENGINE_SRC} ${GAMEENGINE_HEADERS} ${GAMEENGINE_NATVIS} game/crashhandler.cpp)
```
xezon: Maybe this is wrong place. Natvis file must be added to one project only. If it is added to 2, Visual Studio will actually complain. But `thyme_lib` is conditional with `STANDALONE` and `BUILD_TOOLS`. Where to better put?

src/CMakeLists.txt
OmniBlade: I would suggest doing this at the top level CMake and have the xml in a new folder for none source resources, something like `resources/msvc/thyme.natvis`, then have it added to the targets after `add_subdirectory(src)` where you can check which targets exist and add it to the most appropriate, something liek the following:

```CMake
if(MSVC)
    if(TARGET thyme_lib)
        target_sources(thyme_lib PRIVATE resources/msvc/thyme.natvis)
    elseif(TARGET thyme_dll)
        target_sources(thyme_dll PRIVATE resources/msvc/thyme.natvis)
    endif()
endif()
```

And just add checks for as many targets as can potentially be built alone?

xezon: I opted for similar but different approach: I just created a new INTERFACE library for MSVC_IDE that contains all natvis files.

OmniBlade (approved):
Looks good to me.

PR #285, opened 2021-10-18
Improve GameTextManager

xezon:
With these changes, `GameTextManager` is no longer as confined and its functionality can be repurposed in new class `GameTextFile` to load individual files and add new functionality to write out loaded data to new files.

## More details

In new class `GameTextFile` we can add all new functionality by using the static functions, without touching anything further in `GameTextManager`. `GameTextManager` keeps exact same member functions and variables as before. The algorithms in the current static functions need no touching, so they remain in original state. What fundamentally changed is, that `GameTextManager` now calls the static functions of `GameTextFile`. So the code moved away, but it is the same logic. I already tested with both STR and CSF and it works as before.

I did a few more sanity fix ups, which are in all separate commits.


xezon:
I think I will remove all legacy functionality from `GameTextFile` and move it back into original class. And eventually move all new code into a new file, leaving legacy code mostly untouched.

xezon:
The new direction is moving all new functionality into new class `GameTextFile` and leave `GameTextManager` as is. It is then up to stakeholders to transition from legacy functionality to new functionality.

GameTextFile currently can
* Write Utf8 STR files
* Write CSF files

GameTextFile will be able to
* Read Utf8 STR files (as opposed to ASCII STR files from original)
* Read CSF files
* Merge files for compound Utf8 STR file write out
* Read compound Utf8 STR files

Original STR layout:
```
GUI:Label
"This is a label"
END
```
Compounded label:
```
GUI:Label
CH: "这是一个标签"
EN: "This is a label"
ES: "Esta es una etiqueta"
END

xezon:
Closing because this work is no longer reflective of progress.

PR #287, opened 2021-10-19
Weapon Store

duncanspumpkin:
Most of the WeaponStore has been implemented but there is a number of parsing functions that can not be implemented at this time.
WeaponTemplate has been implemented only as far as required to implement WeaponStore functions. Where possible both WorldBuilder and ZeroHour offsets of functions have been provided.

### DIFF
--- a/src/game/logic/object/weapon.cpp
+++ b/src/game/logic/object/weapon.cpp
@@ -16,6 +16,7 @@
 #include "audiomanager.h"
 #include "gamelogic.h"
 #include "ini.h"
+#include "particlesystemplate.h"
 #include "thingfactory.h"
 #include <algorithm>
 
@@ -585,21 +586,26 @@ static void Parse_Per_Vet_Level_FXList(INI *ini, void *formal, void *store, cons
 // zh: 0x004C2BB0 wb: 0x006D3859
 static void Parse_All_Vet_Levels_Particle_System(INI *ini, void *formal, void *store, const void *user_data)
 {
-    // TODO requires ParticleSystem::Parse
-#ifdef GAME_DLL
-    Call_Function<void, INI *, void *, void *, void const *>(
-        PICK_ADDRESS(0x004C2BB0, 0x006D3859), ini, formal, store, user_data);
-#endif
+    ParticleSystemTemplate *particle_template = nullptr;
+    ParticleSystemTemplate::Parse(ini, nullptr, &particle_template, nullptr);
+
+    auto *particle_syst_vet_array = static_cast<ParticleSystemTemplate **>(store);
+    for (auto i = 0; i < VETERANCY_COUNT; ++i) {
+        particle_syst_vet_array[i] = particle_template;
+    }
 }
 
 // zh: 0x004C2B60 wb: 0x006D3807
 static void Parse_Per_Vet_Level_Particle_System(INI *ini, void *formal, void *store, const void *user_data)
 {
-    // TODO requires ParticleSystem::Parse
-#ifdef GAME_DLL
-    Call_Function<void, INI *, void *, void *, void const *>(
-        PICK_ADDRESS(0x004C2B60, 0x006D3807), ini, formal, store, user_data);
-#endif
+    const auto vet_level = ini->Scan_IndexList(ini->Get_Next_Token(), g_veterancyNames);
+
+    auto *particle_syst_vet_array = static_cast<ParticleSystemTemplate **>(store);
+
+    ParticleSystemTemplate *particle_template = nullptr;
+    ParticleSystemTemplate::Parse(ini, nullptr, &particle_template, nullptr);
+
+    particle_syst_vet_array[vet_level] = particle_template;
 }
 
 // zh: 0x004C5C30 wb: 0x006D6C7E
--- a/src/game/logic/object/weapon.cpp
+++ b/src/game/logic/object/weapon.cpp
@@ -13,7 +13,22 @@
  *            LICENSE
  */
 #include "weapon.h"
+#include "audiomanager.h"
+#include "gamelogic.h"
 #include "ini.h"
+#include "thingfactory.h"
+#include <algorithm>
+
+#ifndef GAME_DLL
+WeaponStore *g_theWeaponStore = nullptr;
+#endif
+
+static void Parse_All_Vet_Levels_ASCII_String(INI *ini, void *formal, void *store, const void *user_data);
+static void Parse_Per_Vet_Level_ASCII_String(INI *ini, void *formal, void *store, const void *user_data);
+static void Parse_All_Vet_Levels_FXList(INI *ini, void *formal, void *store, const void *user_data);
+static void Parse_Per_Vet_Level_FXList(INI *ini, void *formal, void *store, const void *user_data);
+static void Parse_All_Vet_Levels_Particle_System(INI *ini, void *formal, void *store, const void *user_data);
+static void Parse_Per_Vet_Level_Particle_System(INI *ini, void *formal, void *store, const void *user_data);
 
 const char *TheWeaponBonusNames[] = {
     "GARRISONED",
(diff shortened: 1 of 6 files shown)


src/game/logic/object/weapon.cpp
```diff
@@ -107,19 +122,24 @@ void WeaponStore::Reset()
 // zh: 0x004C5AB0 wb: 0x006D69D3
 void WeaponStore::Update()
 {
-    for (auto DDI_iter = m_weaponDDI.begin(); DDI_iter != m_weaponDDI.end(); ++DDI_iter) {
-        auto &weapon_delayed_damage = *DDI_iter;
-        if (weapon_delayed_damage.m_delayDamageFrame > g_theGameLogic->Get_Frame()) {
+    // Find any delayed damage that should now be applied and apply it.
+    for (auto DDI_iter = m_weaponDDI.begin(); DDI_iter != m_weaponDDI.end();) {
+        // Have cur_iter point at the current item and DDI point at the next item
+        // two iters required so that items can be correctly errased.
```
xezon: > Two iterators are required so that items can be correctly erased.

src/game/logic/object/weapon.h
```diff
@@ -81,3 +88,221 @@ class WeaponBonusSet : public MemoryPoolObject
...
+enum DeathType : int32_t;
+
+class ThingTemplate;
+class FXList;
+
+class WeaponTemplate : public MemoryPoolObject
+{
+    IMPLEMENT_POOL(WeaponTemplate);
+
+    static FieldParse TheWeaponTemplateFieldParseTable[0];
```
xezon: What is this?

src/game/logic/object/weapon.cpp
```diff
@@ -55,3 +61,117 @@ void WeaponBonusSet::Parse_Weapon_Bonus_Set_Ptr(INI *ini, void *formal, void *st
...
+            weapon_template->Delete_Instance();
+        }
+    }
+}
+
+// zh: 0x004C5BF0 wb: 0x006D6BEE
+void WeaponStore::PostProcessLoad()
+{
+    auto *thing_factory = g_theThingFactory;
+    if (thing_factory == nullptr) {
```
xezon: `if (g_theThingFactory == nullptr)` should do.

src/game/logic/object/weapon.cpp
```diff
@@ -55,3 +61,117 @@ void WeaponBonusSet::Parse_Weapon_Bonus_Set_Ptr(INI *ini, void *formal, void *st
...
+// zh: 0x004C51B0 wb: 0x006D6703
+WeaponTemplate *WeaponStore::Find_Weapon_Template(Utf8String name)
+{
+    if (name.Is_None()) {
+        return nullptr;
+    }
+
+    const auto key = g_theNameKeyGenerator->Name_To_Key(name);
+    auto *weapon_template = Find_Weapon_Template_Private(key);
+    captainslog_dbgassert(weapon_template != nullptr, "Weapon %s not found!\n", name);
```
xezon: This likely would crash instead of asserting. `name.Str()`, otherwise printf will assume `Utf8String` to be the pointer, which it is not.

This is a good example for not using auto as excessively in c++. If `name` was `auto` instead of `Utf8String` in this review, then one could have assumed `name` may be a `const char*`. I would not have spotted the problem in that case.

src/game/logic/object/weapon.h
```diff
@@ -81,3 +88,221 @@ class WeaponBonusSet : public MemoryPoolObject
...
+    bool m_allowAttackGarrisonedBldgs;
+    bool m_playFXWhenStealthed;
+    uint32_t m_preAttackDelay;
+    float m_continueAttackRange;
+    float m_scatterRadiusVsInfantry;
+    DamageStatusType m_damageStatusType;
+    uint32_t m_suspendFXDelay;
+    bool m_missileCallsOnDie;
+    // TODO: Consider replacing the container type
+    std::list<HistoricWeaponDamageInfo> m_historicDamage; // 0x1D0
```
xezon: It's odd the members come first in this class. I assume it 1:1 with the disassembly, not by own choice?

duncanspumpkin: I tend to put members first followed by member functions. I see it as more natural as class members tend to be private

src/game/logic/object/weapon.cpp
duncanspumpkin: Yes its unfortunate we have to use such backwards code that uses the `printf` formatter instead of the type safe `std::format`.

src/game/logic/object/weapon.cpp
duncanspumpkin: I feel all globals should have accessor functions this was written assuming there was going to be an accessor function.

src/game/logic/object/weapon.h
duncanspumpkin: No idea this is the weaponstore pr I've not dont the WeaponTemplate apart from where necessary.

src/game/logic/object/weapon.cpp
xezon: That should not be necessary. Global variables are fine.

src/game/logic/object/weapon.h
xezon: Hmm. As far as I am aware most programmers put functions atop, because public functions are the most viable information to look at, as opposed to scrolling through a list of private member variables or private functions first.

Logical class member order:

1. public functions
2. protected functions
3. private functions
4. protected member variables
5. private member variables

duncanspumpkin: That seems illogical to me I would keep the member variables at the top as that's the important bit for abi compatibility. You want to know sizes, then virtuals then normal functions.

src/game/common/ini/ini.h
```diff
@@ -152,6 +152,7 @@ class INI
     static void Parse_Audio_Event_RTS(INI *ini, void *formal, void *store, const void *user_data);
     static void Parse_Science_Vector(INI *ini, void *formal, void *store, const void *user_data);
     static void Parse_Sounds_Vector(INI *ini, void *formal, void *store, const void *user_data);
+    static void Parse_Weapon_Template(INI *ini, void *formal, void *store, const void *user_data);
```
duncanspumpkin: Remove and put in the weapon template

src/game/common/bitflags.h
```diff
@@ -316,3 +317,9 @@ template<int bits> void BitFlags<bits>::Parse_INI(INI *ini, void *formal, void *
 {
     static_cast<BitFlags<bits> *>(store)->Parse(ini, nullptr);
 }
+
+template<int bits> void BitFlags<bits>::Parse_Single_Bit_From_INI(INI *ini, void *formal, void *store, const void *user_data)
```
xezon: What is the `void* formal` and `const void *user_data` for when not used in this function?

src/game/logic/object/weapon.cpp
```diff
@@ -42,9 +58,205 @@ const char *TheWeaponBonusNames[] = { "GARRISONED",
...
+    nullptr,
+};
+
+const char *TheWeaponBonusFieldNames[] = {
+    "DAMAGE",
+    "RADIUS",
+    "RANGE",
+    "RATE_OF_FIRE",
+    "PRE_ATTACK",
+    nullptr,
```
xezon, line 71: Maybe we can add a special alias for this null string to make it more descriptive that this is the array terminator.

src/game/logic/object/weapon.cpp
```diff
@@ -55,3 +267,363 @@ void WeaponBonusSet::Parse_Weapon_Bonus_Set_Ptr(INI *ini, void *formal, void *st
...
+            continue;
+        }
+
+        delayed_damage.m_delayedWeapon->Deal_Damage_Internal(delayed_damage.m_delaySourceID,
+            delayed_damage.m_delayIntendedVictimID,
+            &delayed_damage.m_delayDamagePos,
+            delayed_damage.m_bonus,
+            false);
+
+        m_weaponDDI.erase(cur_iter, DDI_iter);
```
xezon: `DDI_iter` can be omitted.

src/game/logic/object/weapon.h
```diff
@@ -81,3 +119,179 @@ class WeaponBonusSet : public MemoryPoolObject
...
+
+class WeaponTemplate : public MemoryPoolObject
+{
+    IMPLEMENT_POOL(WeaponTemplate);
+
+public:
+    WeaponTemplate &operator=(WeaponTemplate &rhs);
+
+    bool Has_Next_Template() const { return m_nextTemplate != nullptr; }
+    Utf8String Get_Name() const { return m_name; }
```
xezon: Can also return `const Utf8String&` here. Avoids incrementing ref count when not creating a copy of it.

src/game/logic/object/weapon.h
```diff
@@ -81,3 +119,179 @@ class WeaponBonusSet : public MemoryPoolObject
...
+    NameKeyType Get_Name_Key() const { return m_nameKey; }
+    Utf8String Get_Projectile_Name() const { return m_projectileObject; }
```
xezon: Can also return `const Utf8String&` here. Avoids incrementing ref count when not creating a copy of it.

src/game/logic/object/weapon.h
```diff
@@ -81,3 +119,179 @@ class WeaponBonusSet : public MemoryPoolObject
...
+    WeaponTemplate *New_Weapon_Template(Utf8String name);
+    WeaponTemplate *New_Override(WeaponTemplate *weapon_template);
+    // wb: 0x006D6AA9
+    void Delete_All_Delayed_Damage() { m_weaponDDI.clear(); }
+    void Reset_Weapon_Templates();
+    // wb: 0x006D6B8B
+    void Set_Delayed_Damage(
+        WeaponTemplate *weapon, Coord3D *pos, uint32_t whichFrame, uint32_t sourceID, uint32_t victimID, WeaponBonus &bonus)
+    {
+        m_weaponDDI.push_front(WeaponDelayedDamageInfo{ weapon, *pos, whichFrame, sourceID, victimID, bonus });
```
xezon, line 313: `WeaponTemplate *weapon` and `Coord3D *pos` pointers are unchecked here. Maybe we should do something about it.

src/game/common/bitflags.h
duncanspumpkin: That is the abi for the ini parse functions. Could remove the names to make it more clear perhaps.

src/game/logic/object/weapon.h
duncanspumpkin: Yeah it's an odd design but for some reason they like to pass these by pointer rather than by ref. In this particular case there are no other callers so i think the better way would be to pass the two by reference.

src/game/logic/object/weapon.h
```diff
@@ -112,3 +151,179 @@ class WeaponBonusSet : public MemoryPoolObject
...
+    DamageType m_damageType;
+    DeathType m_deathType;
+    float m_weaponSpeed;
+    float m_minWeaponSpeed;
+    bool m_scaleWeaponSpeed;
+    float m_weaponRecoil;
+    float m_minTargetPitch;
+    float m_maxTargetPitch;
+    float m_radiusDamageAngle;
+    Utf8String m_projectileObject;
```
tomsons26, line 227: this is m_projectileName

src/game/logic/object/weapon.h
```diff
@@ -112,3 +151,179 @@ class WeaponBonusSet : public MemoryPoolObject
...
+    uint32_t unk;
```
tomsons26, line 228: ThingTemplate *m_projectileTmpl

src/game/logic/object/weapon.h
```diff
@@ -112,3 +151,179 @@ class WeaponBonusSet : public MemoryPoolObject
...
+    Utf8String m_fireOCL[VETERANCY_COUNT];
+    Utf8String m_projectileDetonationOCL[VETERANCY_COUNT];
+    ParticleSystemTemplate *m_projectileExhaust[VETERANCY_COUNT];
+    Utf8String m_projectileName[VETERANCY_COUNT];
+    ThingTemplate *m_projectileTmpl[VETERANCY_COUNT];
+    FXList *m_fireFX[VETERANCY_COUNT];
```
tomsons26, line 234: m_fireFXs , later SAGE xsd's have comment naming it

src/game/logic/object/weapon.h
```diff
@@ -112,3 +151,179 @@ class WeaponBonusSet : public MemoryPoolObject
...
+    bool m_scaleWeaponSpeed;
+    float m_weaponRecoil;
+    float m_minTargetPitch;
+    float m_maxTargetPitch;
+    float m_radiusDamageAngle;
+    Utf8String m_projectileObject;
+    uint32_t unk;
+    Utf8String m_fireOCL[VETERANCY_COUNT];
+    Utf8String m_projectileDetonationOCL[VETERANCY_COUNT];
+    ParticleSystemTemplate *m_projectileExhaust[VETERANCY_COUNT];
```
tomsons26, line 231: m_projectileExhausts , later SAGE xsd's have comment naming it

src/game/logic/object/weapon.h
```diff
@@ -112,3 +151,179 @@ class WeaponBonusSet : public MemoryPoolObject
...
+    float m_weaponSpeed;
+    float m_minWeaponSpeed;
+    bool m_scaleWeaponSpeed;
+    float m_weaponRecoil;
+    float m_minTargetPitch;
+    float m_maxTargetPitch;
+    float m_radiusDamageAngle;
+    Utf8String m_projectileObject;
+    uint32_t unk;
+    Utf8String m_fireOCL[VETERANCY_COUNT];
```
tomsons26, line 229: ok these:
i need to check for real name
i need to check for real name
m_projectileExhausts , later SAGE xsd's have comment naming it
wrong
wrong
m_fireFXs , later SAGE xsd's have comment naming it
i need to check for real name

types seem to be
Utf8String[VETERANCY_COUNT]
Utf8String[VETERANCY_COUNT]
ParticleSystemTemplate *[VETERANCY_COUNT]
?[VETERANCY_COUNT] - not Utf8String
?[VETERANCY_COUNT] - not Utf8String
FXList *[VETERANCY_COUNT]
FXList *[VETERANCY_COUNT]

src/game/logic/object/weapon.h
```diff
@@ -112,3 +151,179 @@ class WeaponBonusSet : public MemoryPoolObject
...
+    int32_t m_maxDelayBetweenShots;
+    int32_t m_continuousFireOneShotsNeeded;
+    int32_t m_continuousFireTwoShotsNeeded;
+    uint32_t m_continuousFireCoastFrames;
+    uint32_t m_autoReloadWhenIdle;
+    int32_t m_shotsPerBarrel;
+    int32_t m_antiMask;
+    int32_t m_affectsMask;
+    int32_t m_collideMask;
+    int32_t m_damageDealtAtSelfPosition;
```
tomsons26: bool

src/game/logic/object/weapon.h
tomsons26: Two unknowns are ObjectCreationList *[VETERANCY_COUNT], WB handles them at 006D40BD WeaponStore::postProcessLoad

tomsons26: Seems its FireOCL and ProjectileDetonationOCL the actual OCLs while earlier ones are just names

jonwil:
PR obsoleted by my Weapon work. Closing.

PR #293, opened 2021-10-20
CompositeRenderObjClass

jonwil:


src/w3d/renderer/composite.cpp
```diff
@@ -0,0 +1,238 @@
...
+    Set_Name(that.Get_Name());
+    Set_Base_Model_Name(that.Get_Base_Model_Name());
+    return *this;
+}
+
+CompositeRenderObjClass::~CompositeRenderObjClass() {}
+
+void CompositeRenderObjClass::Restart()
+{
+    for (int i = 0; i < Get_Num_Sub_Objects(); i++) {
```
xezon, line 36: `Get_Num_Sub_Objects()` is a virtual function call. Not clear how expensive this call is or if this changes from loop body. Maybe
```c++
for (int i = 0, count = Get_Num_Sub_Objects(); i < count; ++i)
```
would be an alternative here, to avoid calling it on every iteration.

src/w3d/renderer/composite.cpp
```diff
@@ -0,0 +1,238 @@
...
+        robj->Notify_Removed(scene);
+        robj->Release_Ref();
+    }
+
+    RenderObjClass::Notify_Removed(scene);
+}
+
+bool CompositeRenderObjClass::Cast_Ray(RayCollisionTestClass &raytest)
+{
+    bool b = 0;
```
xezon: `bool collides = false;`

src/w3d/renderer/composite.cpp
```diff
@@ -0,0 +1,238 @@
...
+        b = (robj->Intersect_OBBox(boxtest) | b) != 0;
+        robj->Release_Ref();
+    }
+
+    return b;
+}
+
+void CompositeRenderObjClass::Create_Decal(DecalGeneratorClass *generator)
+{
+    captainslog_assert(0 && "decals not supported");
```
xezon: `captainslog_assert(0, "decals are not supported");`

src/w3d/renderer/composite.cpp
```diff
@@ -0,0 +1,238 @@
...
+void CompositeRenderObjClass::Update_Obj_Space_Bounding_Volumes()
+{
+    if (Get_Num_Sub_Objects() > 0) {
+        RenderObjClass *robj = Get_Sub_Object(0);
+        captainslog_assert(robj);
+        robj->Get_Obj_Space_Bounding_Sphere(m_objSphere);
+        AABoxClass obj_aabox;
+        robj->Get_Obj_Space_Bounding_Box(obj_aabox);
+        robj->Release_Ref();
+        int count = 1;
```
xezon: `int index = 1`

src/w3d/renderer/composite.h
```diff
@@ -0,0 +1,59 @@
...
+    bool Intersect_OBBox(OBBoxIntersectionTestClass &boxtest) override;
+
+    void Create_Decal(DecalGeneratorClass *generator) override;
+    void Delete_Decal(unsigned long decal_id) override;
+
+    void Get_Obj_Space_Bounding_Sphere(SphereClass &sphere) const override;
+    void Get_Obj_Space_Bounding_Box(AABoxClass &box) const override;
+    void Update_Obj_Space_Bounding_Volumes() override;
+
+    void Set_User_Data(void *value, bool recursive) override;
```
xezon: I think we can give these functions the `virtual` keyword as well. Makes it more obvious what these are.

src/w3d/renderer/composite.h
```diff
@@ -0,0 +1,59 @@
...
+    int Get_Num_Polys() const override;
+
+    void Notify_Added(SceneClass *scene) override;
+    void Notify_Removed(SceneClass *scene) override;
+
+    bool Cast_Ray(RayCollisionTestClass &raytest) override;
+    bool Cast_AABox(AABoxCollisionTestClass &boxtest) override;
+    bool Cast_OBBox(OBBoxCollisionTestClass &boxtest) override;
+    bool Intersect_AABox(AABoxIntersectionTestClass &boxtest) override;
+    bool Intersect_OBBox(OBBoxIntersectionTestClass &boxtest) override;
```
xezon: These functions look as if they should be `const`, no?

jonwil: All available information sources we have indicate that no they are not const.

xezon: Hmm ok odd. In that case maybe add a todo comment for investigation later. I would not expect intersection tests to be state changing.

src/w3d/renderer/composite.cpp
jonwil: Its theoretically possible that it could change in the loop body in some cases.

xezon: Ok in that case we need keep fresh function call indeed.

src/w3d/renderer/composite.cpp
```diff
@@ -0,0 +1,234 @@
...
+        collides = (robj->Cast_Ray(raytest) | collides) != 0;
+        robj->Release_Ref();
+    }
+
+    return collides;
+}
+
+bool CompositeRenderObjClass::Cast_AABox(AABoxCollisionTestClass &boxtest)
+{
+    bool b = 0;
```
xezon: There are several more of these to swap for `bool collides = false;`

src/w3d/renderer/composite.cpp
```diff
@@ -0,0 +1,234 @@
...
+}
+
+void CompositeRenderObjClass::Set_Name(const char *name)
+{
+    m_name = name;
+}
+
+const char *CompositeRenderObjClass::Get_Base_Model_Name() const
+{
+    return NULL;
```
xezon: `return nullptr;`

src/w3d/renderer/composite.cpp
```diff
@@ -0,0 +1,234 @@
...
+        collides = (robj->Cast_AABox(boxtest) | collides) != 0;
+        robj->Release_Ref();
+    }
+
+    return collides;
+}
+
+bool CompositeRenderObjClass::Cast_OBBox(OBBoxCollisionTestClass &boxtest)
+{
+    bool collides = 0;
```
xezon: `bool collides = false;`

src/w3d/renderer/composite.cpp
```diff
@@ -0,0 +1,234 @@
...
+        collides = (robj->Cast_OBBox(boxtest) | collides) != 0;
+        robj->Release_Ref();
+    }
+
+    return collides;
+}
+
+bool CompositeRenderObjClass::Intersect_AABox(AABoxIntersectionTestClass &boxtest)
+{
+    bool intersects = 0;
```
xezon: `bool intersects = false;`

src/w3d/renderer/composite.cpp
```diff
@@ -0,0 +1,234 @@
...
+        intersects = (robj->Intersect_AABox(boxtest) | intersects) != 0;
+        robj->Release_Ref();
+    }
+
+    return intersects;
+}
+
+bool CompositeRenderObjClass::Intersect_OBBox(OBBoxIntersectionTestClass &boxtest)
+{
+    bool intersects = 0;
```
xezon: `bool intersects = false;`

tomsons26 (approved):
looks good

PR #295, opened 2021-10-22
HLodClass

jonwil:


src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1532 @@
...
+        for (int i = 0; i < m_lod[m_curLod].Count(); i++) {
+            m_lod[m_curLod][i].m_model->Notify_Added(m_scene);
+        }
+    }
+}
+
+void HLodClass::Decrement_LOD()
+{
+    if (m_curLod > 0) {
+        if (Is_In_Scene()) {
```
tomsons26, line 1389: whole function is in if (m_curLod > 0) {

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1532 @@
...
+            m_lod[i].m_benefitFactor = 1.0f - 0.5f / (float)(polycount * polycount);
+        } else {
+            m_lod[i].m_benefitFactor = 0;
+        }
+    }
+}
+
+void HLodClass::Increment_LOD()
+{
+    if (m_curLod < m_lodCount - 1) {
```
tomsons26, line 1369: whole function is in if (m_curLod < m_lodCount - 1) {

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1532 @@
...
+    }
+
+    return false;
+}
+
+void HLodClass::Add_Lod_Model(int lod, RenderObjClass *robj, int bone_index)
+{
+    captainslog_assert(robj);
+
+    ModelNodeClass node;
```
tomsons26, line 1092: Add asserts from WB

src/w3d/renderer/animobj.h
```diff
@@ -24,30 +24,31 @@ class Animatable3DObjClass : public CompositeRenderObjClass
...
+    virtual void Set_Animation(
+        HAnimClass *motion0, float frame0, HAnimClass *motion1, float frame1, float percentage) override;
+    virtual void Set_Animation(HAnimComboClass *anim_combo) override;
+    virtual HAnimClass *Peek_Animation() override;
+
+    virtual int Get_Num_Bones() override;
+    virtual const char *Get_Bone_Name(int bone_index) override;
+    virtual int Get_Bone_Index(const char *bonename) override;
+    virtual const Matrix3D &Get_Bone_Transform(const char *bonename) override;
+    virtual const Matrix3D &Get_Bone_Transform(int boneindex) override;
```
xezon, line 45: There are a bunch of get functions here. Should they be `const`?

src/w3d/renderer/boxrobj.h
```diff
@@ -13,9 +13,128 @@
...
+    virtual int Get_Num_Polys() const override;
+    virtual void Update_Cached_Box() = 0;
+
+    void Set_Color(const Vector3 &color);
+    void Render_Box(RenderInfoClass &rinfo, const Vector3 &center, const Vector3 &extent);
+    void Vis_Render_Box(RenderInfoClass &rinfo, const Vector3 &center, const Vector3 &extent);
+
+    void Set_Opacity(float opacity) { m_opacity = opacity; }
+    const Vector3 &Get_Local_Center() { return m_objSpaceCenter; }
+    const Vector3 &Get_Local_Extent() { return m_objSpaceExtent; }
```
xezon: These 2 functions should be `const`.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1532 @@
...
+#include "rinfo.h"
+
+W3DErrorType SnapPointsClass::Load_W3D(ChunkLoadClass &cload)
+{
+    int count = cload.Cur_Chunk_Length() / sizeof(Vector3);
+    Resize(count);
+    for (int i = 0; i < count; i++) {
+        Vector3 v;
+
+        if (cload.Read(&v, sizeof(v))) {
```
xezon: Maybe we can rewrite function a bit to make it cleaner and easier to understand:
```c++
    Vector3 vector;
    constexpr int vector_size = sizeof(vector);
    int count = cload.Cur_Chunk_Length() / vector_size;
    Resize(count);
    for (int i = 0; i < count; i++) {
        if (cload.Read(&vector, vector_size)) {
        ...
```

Code looks suspicious. Has potential for fuck up.

xezon: Also, maybe compare the return of the read with the vector_size to make sure whole vector was read:

```c++
if (cload.Read(&vector, vector_size) == vector_size)
```

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1532 @@
...
+                    m_boneIndex = new int[m_modelCount];
+                    int count = 0;
+
+                    if (m_modelCount > 0) {
+                        W3dHLodSubObjectStruct subobjdef;
+
+                        while (cload.Open_Chunk() && cload.Cur_Chunk_ID() == W3D_CHUNK_HLOD_SUB_OBJECT
+                            && cload.Read(&subobjdef, sizeof(subobjdef)) == sizeof(subobjdef) && cload.Close_Chunk()) {
+                            m_modelName[count] = strdup(subobjdef.Name);
+                            m_boneIndex[count++] = subobjdef.BoneIndex;
```
xezon: Can put the `count++` in next line to make it easier on the eyes.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1532 @@
...
+{
+    if (cload.Open_Chunk()) {
+        if (cload.Cur_Chunk_ID() == W3D_CHUNK_HLOD_SUB_OBJECT_ARRAY_HEADER) {
+            W3dHLodArrayHeaderStruct header;
+            if (cload.Read(&header, sizeof(header)) == sizeof(header)) {
+                if (cload.Close_Chunk()) {
+                    m_modelCount = header.ModelCount;
+                    m_maxScreenSize = header.MaxScreenSize;
+                    m_modelName = new char *[m_modelCount];
+                    m_boneIndex = new int[m_modelCount];
```
xezon: Hmm. Here we allocate even if `m_modelCount` is 0. The check for 0 just comes later. It's valid C++, but looks dirty. What if the free function was just to look for `m_modelCount` and would not free the arrays because of that? Should we do something about it?

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1532 @@
...
+                    int count = 0;
```
xezon: This variable definition can be moved further down into the next scope.
`count` can be renamed to `model_index`.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1532 @@
...
+
+W3DErrorType HLodDefClass::Load_W3D(ChunkLoadClass &cload)
+{
+    Free();
+
+    if (!Read_Header(cload)) {
+        return W3D_ERROR_LOAD_FAILED;
+    }
+
+    int i = 0;
```
xezon: This variable definition can be moved further down into the next scope.
`i` can be renamed to `lod_index`.

src/w3d/renderer/hlod.h
```diff
@@ -0,0 +1,348 @@
...
+        bool Load_W3D(ChunkLoadClass &cload);
+        friend class HLodClass;
+    };
+
+    char *m_name;
+    char *m_hierarchyTreeName;
+    int m_lodCount;
+    SubObjectArrayClass *m_lod;
+    SubObjectArrayClass m_aggregates;
+    ProxyArrayClass *m_proxyArray;
```
xezon, line 127: Are all these members supposed to be public, or would it make sense to add some friend classes here instead to hide these in private?

src/w3d/renderer/hlod.h
```diff
@@ -0,0 +1,348 @@
...
+
+public:
+    HLodDefClass() : m_name(0), m_hierarchyTreeName(0), m_lodCount(0), m_lod(0), m_proxyArray(0) {}
```
xezon: Can use `nullpr` here.

src/w3d/renderer/hlod.h
```diff
@@ -0,0 +1,348 @@
...
+            free(m_hierarchyTreeName);
+            m_hierarchyTreeName = 0;
+        }
+        if (m_lod) {
+            delete[] m_lod;
+            m_lod = 0;
+        }
+        if (m_proxyArray) {
+            m_proxyArray->Release_Ref();
+            m_proxyArray = 0;
```
xezon: Can use `nullpr` in this function.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1532 @@
...
+        if (cload.Cur_Chunk_ID() == W3D_CHUNK_HLOD_SUB_OBJECT_ARRAY_HEADER) {
+            W3dHLodArrayHeaderStruct header;
+
+            if (cload.Read(&header, sizeof(header)) == sizeof(header)) {
+                if (cload.Close_Chunk()) {
+                    m_proxyArray = new ProxyArrayClass(header.ModelCount);
+                    int i = 0;
+
+                    if (m_proxyArray->Length() > 0) {
+                        int j = 0;
```
xezon: Better variable names for `i` and `j` would be nice.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1532 @@
...
+                        W3dHLodSubObjectStruct subobjdef;
+
+                        while (cload.Open_Chunk() && cload.Cur_Chunk_ID() == W3D_CHUNK_HLOD_SUB_OBJECT
+                            && cload.Read(&subobjdef, sizeof(subobjdef)) == sizeof(subobjdef) && cload.Close_Chunk()) {
+                            (*m_proxyArray)[j].Init(subobjdef);
+                            i++;
+                            j++;
+
+                            if (i > m_proxyArray->Length()) {
```
xezon: Hmm. I wonder if this should be `i >= m_proxyArray->Length()` like in the function a bit above.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1532 @@
...
+        if (def->Load_W3D(cload)) {
+            delete def;
+            return 0;
+        } else {
+            HLodPrototypeClass *proto = new HLodPrototypeClass(def);
+            return proto;
+        }
+    }
+
+    return 0;
```
xezon: `return nullptr;`

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1532 @@
...
+}
+
+PrototypeClass *HLodLoaderClass::Load_W3D(ChunkLoadClass &cload)
+{
+    HLodDefClass *def = new HLodDefClass();
+
+    if (def) {
+        if (def->Load_W3D(cload)) {
+            delete def;
+            return 0;
```
xezon: `return nullptr;`

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1532 @@
...
+HLodClass::HLodClass(const HLodClass &src) :
+    Animatable3DObjClass(src),
+    m_lodCount(0),
+    m_curLod(0),
+    m_lod(0),
+    m_boundingBoxIndex(-1),
+    m_cost(0),
+    m_value(0),
+    m_snapPoints(0),
+    m_proxyArray(0),
```
xezon: Looks like some of these should be `nullptr`

src/w3d/renderer/animobj.h
duncanspumpkin: That would not override unfortunately. const of a function is part of the virtual interface.

jonwil: Nope, they aren't meant to be const.

src/w3d/renderer/hlod.cpp
jonwil: Fixed that missing ==
No need to do anything else in the function.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: I only see one assert in this function and its there.

src/w3d/renderer/boxrobj.h
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: This is how the original code did it and the goal is to replicate the original code. We can't be sure just what code might rely on it being this way.

src/w3d/renderer/hlod.cpp
jonwil: Moved and renamed.

src/w3d/renderer/hlod.cpp
jonwil: Moved and renamed.

src/w3d/renderer/hlod.h
jonwil: Made them private.

src/w3d/renderer/hlod.h
jonwil: Fixed

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1524 @@
...
+void HLodClass::Update_Sub_Object_Transforms()
+{
+    Animatable3DObjClass::Update_Sub_Object_Transforms();
+
+    for (int i = 0; i < m_lodCount; i++) {
+        ModelArrayClass &lod = m_lod[i];
+        for (int m = 0; m < lod.Count(); ++m) {
+            RenderObjClass *model = lod[m].m_model;
+            int bone_index = lod[m].m_boneIndex;
+
```
tomsons26, line 806: add asserts from wb

jonwil: Already done, they are in HTreeClass::Get_Transform

tomsons26 (approved):
Mentioned minor changes in DMs to Jon. Otherwise looks good.

src/w3d/renderer/w3d_file.h
```diff
@@ -419,3 +419,33 @@ struct W3dAdaptiveDeltaAnimChannelStruct
     float Scale;
     uint32_t Data[1];
 };
+
+#define NO_MAX_SCREEN_SIZE GAMEMATH_FLOAT_MAX
+struct W3dHLodHeaderStruct
+{
+    uint32_t Version;
+    uint32_t LodCount;
+    char Name[16];
```
xezon, line 428: Maybe we can give this variable a different name, because there also is a second name in this struct. `LodName` ? On the other hand it is consistent with the other structs that also use `Name`. So it probably is fine.

src/w3d/renderer/rendobj.h
```diff
@@ -366,6 +378,7 @@ class RenderObjClass : public RefCountClass, public PersistClass, public MultiLi
         BOUNDING_VOLUMES_VALID = 0x00002000,
         IS_TRANSLUCENT = 0x00004000,
         IS_UNK_15 = 0x00008000,
+        SUBOBJS_MATCH_LOD = 0x10000,
```
xezon: Minor: Add the leading zeros for style consistency. Or remove from the others. This would actually be easier to read if the numbers were in one column, but with clang formatting required... :/

src/w3d/renderer/rendobj.h
```diff
@@ -356,6 +357,17 @@ class RenderObjClass : public RefCountClass, public PersistClass, public MultiLi
     void Invalidate_Cached_Bounding_Volumes() const { m_bits &= ~BOUNDING_VOLUMES_VALID; }
     void Validate_Cached_Bounding_Volumes() const { m_bits |= BOUNDING_VOLUMES_VALID; }
 
+    void Set_Sub_Objects_Match_LOD(int onoff)
```
xezon, line 360: `int on` or `int enable` would be better. And maybe passing a `bool` type would be sufficient?

src/w3d/renderer/hlod.h
```diff
@@ -0,0 +1,330 @@
...
+    virtual int Get_Additional_Model_Count();
+    virtual RenderObjClass *Peek_Additional_Model(int model_index);
+    virtual RenderObjClass *Get_Additional_Model(int model_index);
+    virtual int Get_Additional_Model_Bone(int model_index);
+
+    virtual bool Is_NULL_Lod_Included();
+    virtual void Include_NULL_Lod(bool include);
+
+    virtual int Get_Proxy_Count();
+    virtual bool Get_Proxy(int index, ProxyClass &proxy);
```
xezon, line 326: A lot of these virtual functions above are getters. Can any of them be `const`?

src/w3d/renderer/hlod.h
```diff
@@ -0,0 +1,330 @@
...
+    float *m_cost;
+    float *m_value;
+    ModelArrayClass m_additionalModels;
+    SnapPointsClass *m_snapPoints;
+    ProxyArrayClass *m_proxyArray;
+    float m_lodBias;
+
+public:
+    HLodClass(const HLodClass &src);
+    HLodClass(const HLodDefClass &src);
```
xezon, line 236: Does this class no have standard constructor? I can't find it. I would expect it one to have, because it has various variable members that look like they would want initializing. If it does not, you can put a deleted one to make it explicit.
```c++
HLodClass() = delete;
```
Or Add a constructor and initialize variables?

src/w3d/renderer/hlod.h
```diff
@@ -0,0 +1,330 @@
...
+
+class HLodPrototypeClass : public W3DMPO, public PrototypeClass
+{
+    IMPLEMENT_W3D_POOL(HLodPrototypeClass);
+
+private:
+    HLodDefClass *m_definition;
+
+public:
+    HLodPrototypeClass(HLodDefClass *def) : m_definition(def) {}
```
xezon, line 186: Can add
```c++
HLodPrototypeClass() = delete;
```
to make no default constructor explicit and obvious.

src/w3d/renderer/hlod.h
```diff
@@ -0,0 +1,330 @@
...
+};
+
+class ProxyClass
+{
+private:
+    StringClass m_name;
+    Matrix3D m_transform;
+
+public:
+    ProxyClass(char *name, class Matrix3D &tm) : m_name(name), m_transform(tm) {}
```
xezon: `const char* name` I think. `Set_Name` function below also takes const.

src/w3d/renderer/hlod.h
```diff
@@ -0,0 +1,330 @@
...
+    ProxyRecordClass() : m_boneIndex(0) { memset(m_name, 0, sizeof(m_name)); }
+
+    bool operator==(const ProxyRecordClass &src) { return false; }
+    bool operator!=(const ProxyRecordClass &src) { return true; }
+
+    void Init(W3dHLodSubObjectStruct &str)
+    {
+        m_boneIndex = str.BoneIndex;
+        strcpy(m_name, str.Name);
+        m_name[31] = 0;
```
xezon, line 48: Uh. The `strcpy` works here because `str.Name` is 31 bytes and `m_name` is 32 bytes. But this does not look cool. The least we should add is a static assert such that

```c++
static_assert(sizeof(m_name) >= sizeof(str.Name));
```
Alternative would be to use
```c++
strlcpy(m_name, str.Name, sizeof(m_name));
// remove: m_name[31] = 0;
```

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+            robj->Set_Container(nullptr);
+            robj->Release_Ref();
+        }
+        m_lod[i].Delete_All();
+    }
+
+    delete[] m_lod;
+    m_lodCount = 0;
+    delete[] m_cost;
+    delete[] m_value;
```
xezon, line 1532: This looks suspicious. Is `Free()` called more than once? Can one of the pointers be 0? If one of these 2 is true, this can break. I think we would all be more comfortable if all pointers are guarded by condition and then set to `nullptr` after delete.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+    return nullptr;
+}
+
+bool HLodClass::Cast_Ray(RayCollisionTestClass &raytest)
+{
+    if (Are_Sub_Object_Transforms_Dirty()) {
+        Update_Sub_Object_Transforms();
+    }
+
+    ModelArrayClass *lod = &m_lod[m_lodCount - 1];
```
xezon, line 1500: I assume `m_lodCount` cannot be 0 at this point by design? If so we can add a sanity assert here for this to make it clear.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+
+bool HLodClass::Is_NULL_Lod_Included()
+{
+    if (m_lodCount > 0) {
+        RenderObjClass *obj = (*m_lod)[0].m_model;
+        if (obj) {
+            return (obj->Class_ID() == RenderObjClass::CLASSID_NULL);
+        }
+    }
+    return 0;
```
xezon: `return false;`

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+
+    for (int i = 0; i < m_additionalModels.Count(); i++) {
+        m_additionalModels[i].m_model->Set_LOD_Bias(newbias);
+    }
+}
+
+int HLodClass::Get_Lod_Model_Count(int lod_index)
+{
+    captainslog_assert(lod_index >= 0);
+    captainslog_assert(lod_index < m_lodCount);
```
xezon, line 1211: You added these asserts to various functions, but the functions also have condition checks. Is this intended?

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+void HLodClass::Set_Max_Screen_Size(int lod_index, float size)
+{
+    captainslog_assert(lod_index >= 0);
+    captainslog_assert(lod_index < m_lodCount);
+
+    if (lod_index >= 0) {
+        if (lod_index < m_lodCount) {
+            m_lod[lod_index].m_maxScreenSize = size;
+            Recalculate_Static_LOD_Factors();
+            int lod = Calculate_Cost_Value_Arrays(1.0, m_value, m_cost);
```
xezon: `1.0f`

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+}
+
+void HLodClass::Special_Render(SpecialRenderInfoClass &rinfo)
+{
+    if (Is_Not_Hidden_At_All()) {
+        Animatable3DObjClass::Special_Render(rinfo);
+        int lod = m_curLod;
+
+        if (rinfo.m_renderType == SpecialRenderInfoClass::RENDER_SHADOW) {
+            lod = m_lodCount - 1;
```
xezon, line 1113: I assume `m_lodCount` cannot be 0 at this point by design? If so we can add a sanity assert here for this to make it clear.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+            m_cost = costs;
+            m_value = values;
+            m_lodCount++;
+
+            Add_Lod_Model(0, null_object, 0);
+            null_object->Release_Ref();
+        }
+    }
+
+    int lod = Calculate_Cost_Value_Arrays(1.0, m_value, m_cost);
```
xezon: `1.0f`

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+            }
+
+            float *costs = new float[m_lodCount];
+            float *values = new float[m_lodCount + 1];
+            memcpy(costs, &m_cost + 1, m_lodCount * 4);
+            memcpy(values, &m_value + 1, (m_lodCount * 4) + 4);
+
+            delete[] m_lod;
+            delete[] m_cost;
+            delete[] m_value;
```
xezon, line 1019: Unchecked deletes and access to class pointers here. If they are guaranteed to be allocated, we should at least add asserts further up.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+            ModelArrayClass *lods = new ModelArrayClass[m_lodCount];
+
+            for (int j = 0; j < m_lodCount; j++) {
+                lods[j] = m_lod[j + 1];
+            }
+
+            float *costs = new float[m_lodCount];
+            float *values = new float[m_lodCount + 1];
+            memcpy(costs, &m_cost + 1, m_lodCount * 4);
+            memcpy(values, &m_value + 1, (m_lodCount * 4) + 4);
```
xezon, line 1015: What it does is erase the first element from `m_cost` and `m_value` arrays. Do we not have an array type that has methods for that? This looks to me it should not be done manually here in this convoluted way, because this is not maintainable.

If we have no such simple Array wrapper, I can create one.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+        }
+    } else {
+        m_objSphere.Init(Vector3(0, 0, 0), 0);
+        m_objBox.Init(Vector3(0, 0, 0), Vector3(0, 0, 0));
+    }
+}
+
+void HLodClass::Scale(float scale)
+{
+    if (scale != 1.0) {
```
xezon: `1.0f`

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+        if (robj && robj->Class_ID() == CLASSID_OBBOX) {
+            Matrix3D tm1;
+            Simple_Evaluate_Bone(m_lod[m_lodCount - 1][m_boundingBoxIndex].m_boneIndex, &tm1);
+            Matrix3D tm2;
+            Get_Transform().Get_Orthogonal_Inverse(tm2);
+
+            Matrix3D tm3;
+            Matrix3D::Multiply(tm2, tm1, &tm3);
+            tm3.Transform_Center_Extent_AABox(((OBBoxRenderObjClass *)robj)->Get_Local_Center(),
+                ((OBBoxRenderObjClass *)robj)->Get_Local_Extent(),
```
xezon:
```c++
static_cast<OBBoxRenderObjClass *>(robj)->Get_Local_Extent()
```

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+
+        if (robj && robj->Class_ID() == CLASSID_OBBOX) {
+            Matrix3D tm1;
+            Simple_Evaluate_Bone(m_lod[m_lodCount - 1][m_boundingBoxIndex].m_boneIndex, &tm1);
+            Matrix3D tm2;
+            Get_Transform().Get_Orthogonal_Inverse(tm2);
+
+            Matrix3D tm3;
+            Matrix3D::Multiply(tm2, tm1, &tm3);
+            tm3.Transform_Center_Extent_AABox(((OBBoxRenderObjClass *)robj)->Get_Local_Center(),
```
xezon:
```c++
static_cast<OBBoxRenderObjClass *>(robj)->Get_Local_Center()
```

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+{
+    AABoxClass box;
+    Get_Obj_Space_Bounding_Box(box);
+    sphere.Center = box.m_center;
+    sphere.Radius = box.m_extent.Length();
+}
+
+void HLodClass::Get_Obj_Space_Bounding_Box(AABoxClass &box) const
+{
+    if (m_boundingBoxIndex < 0 || m_boundingBoxIndex >= m_lod[m_lodCount - 1].Count()) {
```
xezon, line 859: I assume `m_lodCount` cannot be 0 at this point by design? If so we can add a sanity assert here for this to make it clear.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+            return m_additionalModels[k].m_boneIndex;
+        }
+    }
+
+    return 0;
+}
+
+int HLodClass::Get_Sub_Object_Bone_Index(int lodindex, int modelindex) const
+{
+    return m_lod[lodindex][modelindex].m_boneIndex;
```
xezon, line 764: Do the indexes need checking? If not, at least can add asserts here if possible? (Larger equal 0, smaller X)
Also can rename to `lod_index` and `model_index` for readability.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+    for (int k = 0; k < m_additionalModels.Count(); k++) {
+        if (m_additionalModels[k].m_boneIndex == bone_index) {
+            count++;
+        }
+    }
+
+    return count;
+}
+
+RenderObjClass *HLodClass::Get_Sub_Object_On_Bone(int index, int bone_index) const
```
xezon, line 713: Does `index` refer to `lod_index`? Better name would be good.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+}
+
+void HLodClass::Scale(float scalex, float scaley, float scalez) {}
+
+void HLodClass::Set_HTree(HTreeClass *htree)
+{
+    Animatable3DObjClass::Set_HTree(htree);
+}
+
+RenderObjClass *HLodClass::Get_Sub_Object(int index) const
```
xezon, line 621: Can we have a better name for the `index`?

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+{
+    captainslog_assert(index >= 0);
```
xezon, line 623: Can we also add assert for upper limit? (smaller X)

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+    if (m_snapPoints) {
+        return m_snapPoints->Count();
+    } else {
+        return 0;
+    }
+}
+
+void HLodClass::Get_Snap_Point(int point, Vector3 *set)
+{
+    captainslog_assert(set != nullptr);
```
xezon, line 605: Can `int point` have an assert? (larger equal 0, smaller X)

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+    return result;
+}
+
+bool HLodClass::Intersect_OBBox(OBBoxIntersectionTestClass &boxtest)
+{
+    if (Are_Sub_Object_Transforms_Dirty()) {
+        Update_Sub_Object_Transforms();
+    }
+
+    ModelArrayClass *lod = &m_lod[m_lodCount - 1];
```
xezon, line 570: I assume `m_lodCount` cannot be 0 at this point by design? If so we can add a sanity assert here for this to make it clear.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+    return result;
+}
+
+bool HLodClass::Intersect_AABox(AABoxIntersectionTestClass &boxtest)
+{
+    if (Are_Sub_Object_Transforms_Dirty()) {
+        Update_Sub_Object_Transforms();
+    }
+
+    ModelArrayClass *lod = &m_lod[m_lodCount - 1];
```
xezon, line 550: I assume `m_lodCount` cannot be 0 at this point by design? If so we can add a sanity assert here for this to make it clear.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+    return result;
+}
+
+bool HLodClass::Cast_OBBox(OBBoxCollisionTestClass &boxtest)
+{
+    if (Are_Sub_Object_Transforms_Dirty()) {
+        Update_Sub_Object_Transforms();
+    }
+
+    ModelArrayClass *lod = &m_lod[m_lodCount - 1];
```
xezon, line 530: I assume `m_lodCount` cannot be 0 at this point by design? If so we can add a sanity assert here for this to make it clear.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+    Set_Sub_Object_Transforms_Dirty(true);
+}
+
+bool HLodClass::Cast_AABox(AABoxCollisionTestClass &boxtest)
+{
+    if (Are_Sub_Object_Transforms_Dirty()) {
+        Update_Sub_Object_Transforms();
+    }
+
+    ModelArrayClass *lod = &m_lod[m_lodCount - 1];
```
xezon, line 510: I assume `m_lodCount` cannot be 0 at this point by design? If so we can add a sanity assert here for this to make it clear.

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+            }
+
+            m_additionalModels.Add(node);
+        }
+
+        m_lodBias = that.m_lodBias;
+    }
+
+    Recalculate_Static_LOD_Factors();
+    int cost = Calculate_Cost_Value_Arrays(1.0, m_value, m_cost);
```
xezon: `1.0f`

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+        if (robj) {
+            Add_Sub_Object_To_Bone(robj, src.m_aggregates.m_boneIndex[i]);
+            robj->Release_Ref();
+        }
+    }
+
+    ProxyArrayClass *array = (ProxyArrayClass *)src.m_proxyArray;
+    Ref_Ptr_Set(m_proxyArray, array);
+
+    int cost = Calculate_Cost_Value_Arrays(1.0, m_value, m_cost);
```
xezon: `1.0f`

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+    for (int i = 0; i < src.m_aggregates.m_modelCount; i++) {
+        RenderObjClass *robj = W3DAssetManager::Get_Instance()->Create_Render_Obj(src.m_aggregates.m_modelName[i]);
+
+        if (robj) {
+            Add_Sub_Object_To_Bone(robj, src.m_aggregates.m_boneIndex[i]);
+            robj->Release_Ref();
+        }
+    }
+
+    ProxyArrayClass *array = (ProxyArrayClass *)src.m_proxyArray;
```
xezon: Can this be a `static_cast`?

src/w3d/renderer/hlod.cpp
```diff
@@ -0,0 +1,1565 @@
...
+            return proto;
+        }
+    }
+
+    return nullptr;
+}
+
+RenderObjClass *HLodPrototypeClass::Create()
+{
+    return new HLodClass(*m_definition);
```
xezon, line 235: This looks suspicious. In destructor below `m_definition` is tested, but here in this function it is not. Bug?

src/w3d/renderer/rendobj.h
jonwil: Fixed.

src/w3d/renderer/w3d_file.h
jonwil: Not changing the name.

src/w3d/renderer/rendobj.h
jonwil: This matches the stock code (which is the goal here) so this isn't changing.

src/w3d/renderer/hlod.h
jonwil: All the functions in HLodClass have const or not const matching with the stock code. Nothing will be changed here.

src/w3d/renderer/hlod.h
jonwil: No there is no standard constructor for this class (at least not in the bits of code that actually exist in the game exe) and no adding an explicit dummy constructor doesn't make sense.

src/w3d/renderer/hlod.h
jonwil: Again adding a dummy constructor doesn't make sense.

src/w3d/renderer/hlod.h
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Yes, this is how the original code looked.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Fixed.

src/w3d/renderer/hlod.cpp
jonwil: Made it a const_cast (which makes sense here)

PR #299, opened 2021-10-22
Fix format buffer overflow, utf conversion, character access functions, const correctness in Utf8String and Utf16String

xezon:
I organized all changes in atomic commits, so its all clean. When merging, ideally don't squash it to preserve clean history. I was not able to test ICU, but I am confident the code is correct anyway.

### DIFF
--- a/src/game/common/system/asciistring.cpp
+++ b/src/game/common/system/asciistring.cpp
@@ -54,7 +54,17 @@ Utf8String::Utf8String(const char *s) : m_data(nullptr)
 /**
  * A utility method to test nullptr on string content and log if that happens
  */
-char *Utf8String::Peek() const
+const char *Utf8String::Peek() const
+{
+    captainslog_dbgassert(m_data != nullptr, "null string ptr");
+
+    return m_data->Peek();
+}
+
+/**
+ * A utility method to test nullptr on string content and log if that happens
+ */
+char *Utf8String::Peek()
 {
     captainslog_dbgassert(m_data != nullptr, "null string ptr");
 
@@ -152,8 +162,16 @@ int Utf8String::Get_Length() const
 
 char Utf8String::Get_Char(int index) const
 {
-    captainslog_dbgassert(index >= 0, "bad index in getCharAt.");
-    captainslog_dbgassert(Get_Length() > 0, "strlen returned less than or equal to 0 in getCharAt.");
+    captainslog_dbgassert(index >= 0, "Index must be equal or larger than 0.");
+    captainslog_dbgassert(index < Get_Length(), "Index must be smaller than length.");
+
+    return Peek()[index];
+}
+
+char &Utf8String::Get_Char(int index)
+{
+    captainslog_dbgassert(index >= 0, "Index must be equal or larger than 0.");
+    captainslog_dbgassert(index < Get_Length(), "Index must be smaller than length.");
 
     return Peek()[index];
 }
--- a/src/game/common/system/asciistring.cpp
+++ b/src/game/common/system/asciistring.cpp
@@ -233,69 +233,67 @@ void Utf8String::Set(Utf8String const &string)
 /**
  * Converts a Utf16 string to Utf8
  */
-void Utf8String::Translate(Utf16String const &string)
+void Utf8String::Translate_Internal(const unichar_t *utf16_string, const int utf16_len)
 {
     Release_Buffer();
 
-#if defined BUILD_WITH_ICU // Use ICU converters.
-    int32_t length;
-    UErrorCode error = U_ZERO_ERROR;
-    u_strToUTF8(nullptr, 0, &length, string, -1, &error);
+#if defined BUILD_WITH_ICU
+    // Use ICU converters.
+    if (utf16_len > 0) {
+        int32_t utf8_len;
+        UErrorCode error = U_ZERO_ERROR;
+        // Get utf8 string length.
+        u_strToUTF8(nullptr, 0, &utf8_len, utf16_string, utf16_len, &error);
 
-    if (U_SUCCESS(error) && length > 0) {
-        u_strToUTF8(Get_Buffer_For_Read(length), length, nullptr, string, -1, &error);
+        if (U_SUCCESS(error) && utf8_len > 0) {
+            // Allocate and fill new utf8 string.
+            char *utf8_buffer = Get_Buffer_For_Read(utf8_len);
+            u_strToUTF8(utf8_buffer, utf8_len, nullptr, utf16_string, utf16_len, &error);
 
-        if (U_FAILURE(error)) {
-            Clear();
-        }
-    }
-#elif defined PLATFORM_WINDOWS // Use WIN32 API converters.
-    int length = WideCharToMultiByte(CP_UTF8, 0, string, -1, nullptr, 0, nullptr, nullptr);
-
-    if (length > 0) {
-        WideCharToMultiByte(CP_UTF8, 0, string, -1, Get_Buffer_For_Read(length), length, nullptr, nullptr);
-    }
-#else // Naive copy, this is what the original does.
-    int str_len = string.Get_Length();
-
-    for (int i = 0; i < str_len; ++i) {
-        // This is a debug assert from the look of it.
-        /*if ( v4 < 8 || (!stringSrc.m_data ? (v6 = 0) : (v5 = stringSrc.Peek(), v6 = wcslen(v5)), v3 >= v6) )
-        {
-            if ( `Utf16String::Get_Char'::`14'::allowCrash )
-            {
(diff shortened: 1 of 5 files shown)


src/game/common/system/unicodestring.h
```diff
@@ -110,26 +108,28 @@ class Utf16String
...
 
-    // TODO
-    // unichar_t *operator[](int index) const { return m_data->Peek()[index]; }
+    unichar_t operator[](int index) const { return Get_Char(index); }
+    unichar_t &operator[](int index) { return Get_Char(index); }
 
     void Validate();
-    unichar_t *Peek() const;
+    const unichar_t *Peek() const;
+    unichar_t *Peek();
```
xezon, line 116: Original `Peek()` was `const`, but returned writable pointer to data. Not good!

src/game/common/system/asciistring.h
```diff
@@ -97,20 +95,26 @@ class Utf8String
 
     operator const char *() const { return Str(); }
 
+    char operator[](int index) const { return Get_Char(index); }
+    char &operator[](int index) { return Get_Char(index); }
```
xezon, line 99: Every c++ programmer expects [] operator for char access. So I added it here for convenience.

src/game/common/system/asciistring.cpp
```diff
@@ -471,7 +487,7 @@ void Utf8String::Format_VA(const char *format, va_list args)
 void Utf8String::Format_VA(Utf8String &format, va_list args)
 {
     char buf[MAX_FORMAT_BUF_LEN];
-    int res = vsnprintf(buf, sizeof(buf), format.Str(), args);
+    int res = vsnprintf(buf, ARRAY_SIZE(buf), format.Str(), args);
```
xezon, line 490: `sizeof(buf)` is twice as large as `ARRAY_SIZE(buf)` but should not be.

src/game/common/system/asciistring.cpp
```diff
@@ -215,71 +233,69 @@ void Utf8String::Set(Utf8String const &string)
...
-        Concat(reinterpret_cast<char *>(&c));
+    }
+#else
+    // Naive copy, this is what the original does.
+    for (int i = 0; i < utf16_len; ++i) {
+        char c = static_cast<char>(utf16_string[i]);
+        // FEATURE: Append ? character for non ASCII characters
+        if (c < 0)
+            c = '?';
+        Concat(c);
```
xezon: The original naive copy was a complete mess. I simplified it.

src/game/common/system/asciistring.cpp
```diff
@@ -215,71 +233,69 @@ void Utf8String::Set(Utf8String const &string)
...
+            if (U_FAILURE(error)) {
+                Clear();
+            } else {
+                // Add null terminator manually.
+                utf8_buffer[utf8_len] = '\0';
+            }
         }
     }
-#elif defined PLATFORM_WINDOWS // Use WIN32 API convertors.
-    int length = WideCharToMultiByte(CP_UTF8, 0, string, -1, nullptr, 0, nullptr, nullptr);
```
xezon, line 235: This did not work correctly. Returned `length` was 1 on empty string, because `WideCharToMultiByte` includes the null terminator as a string character. And empty string will then cause an allocation here, which will trigger asserts further down the line. New code does not treat null terminator as source character.

src/game/common/system/asciistring.cpp
```diff
@@ -152,8 +162,16 @@ int Utf8String::Get_Length() const
 
 char Utf8String::Get_Char(int index) const
 {
-    captainslog_dbgassert(index >= 0, "bad index in getCharAt.");
-    captainslog_dbgassert(Get_Length() > 0, "strlen returned less than or equal to 0 in getCharAt.");
```
xezon, line 156: Condition does not look right. Fixed.

src/game/common/system/asciistring.h
```diff
@@ -97,20 +95,26 @@ class Utf8String
...
+    char &Get_Char(int index);
     const char *Str() const;
     char *Get_Buffer_For_Read(int len);
     // These two should probably be private with the = operator being the preferred interface?
     void Set(const char *s);
     void Set(Utf8String const &string);
 
-    void Translate(Utf16String const &stringSrc);
+    void Translate(Utf16String const &utf16_string);
+    void Translate(const unichar_t *utf16_string);
```
xezon, line 117: The `Utf16String` class has this `Translate` overload which takes c string. So I added it here too for consistency sake.

src/game/common/system/asciistring.h
duncanspumpkin: If this is a true utf8 string this is very dangerous to ever use.

src/game/common/system/asciistring.cpp
duncanspumpkin: This comment was on the wrong impl. Utf8String was always fine but yes Utf16String equivalent function to this has double the size.

src/game/common/system/asciistring.cpp
```diff
@@ -215,71 +233,69 @@ void Utf8String::Set(Utf8String const &string)
...
         }
-#endif
-
-        Concat(reinterpret_cast<char *>(&c));
+    }
+#else
+    // Naive copy, this is what the original does.
+    for (int i = 0; i < utf16_len; ++i) {
+        char c = static_cast<char>(utf16_string[i]);
+        // FEATURE: Append ? character for non ASCII characters
```
duncanspumpkin, line 280: This would only append them for half of the non ASCII characters. It could also be inserting nulls half way through the string. This is really not safe. I think you should probably just do:

```cpp
const auto c = utf16_string[i];
// FEATURE: Append ? character for non ASCII characters
Concat((c < 128) ? c : '?');
```
It's still not great and will output emojibake on multi code point characters but at least it doesn't have the null issues or ignoring half of the non ascii.

src/game/common/system/asciistring.h
xezon: Well. The class is called `Utf8String`. The file is called `asciistring`. The vast majority of use is expected to be for ASCII. It is a 1 byte character sequence container first. I don't think this is dangerous per se. It is up to the user of this class how to interact with it. If it is meant to store utf8 character sequence, user of class is responsible to interact in a meaningful way.

duncanspumpkin: It doesn't return codepoints. So every consumer has to understand that you can only compare the result with ASCII characters and it's encouraging users to access byte by byte instead of by codepoint or grapheme cluster.

src/game/common/system/asciistring.cpp
xezon: Thanks. I will fix.

src/game/common/system/asciistring.h
xezon: Ok but there has already been a `char Get_Char()` function to begin with. What you suggest sounds like a feature beyond the scope of this change.

src/game/common/system/asciistring.cpp
xezon: Ok fix is done. I tested with following. Result looked proper.
```c++
    Utf16String str16;
    str16.Concat(U_CHAR('\x0010'));
    str16.Concat(U_CHAR('\x0020'));
    str16.Concat(U_CHAR('\x0030'));
    str16.Concat(U_CHAR('\x0040'));
    str16.Concat(U_CHAR('\x0050'));
    str16.Concat(U_CHAR('\x0060'));
    str16.Concat(U_CHAR('\x0070'));
    str16.Concat(U_CHAR('\x0080'));
    str16.Concat(U_CHAR('\x0090'));
    str16.Concat(U_CHAR('\x0100'));
    str16.Concat(U_CHAR('\x0110'));
    str16.Concat(U_CHAR('\x0120'));
    str16.Concat(U_CHAR('\x0130'));
    str16.Concat(U_CHAR('\x0140'));
    str16.Concat(U_CHAR('\x0150'));
    str16.Concat(U_CHAR('\x0160'));
    str16.Concat(U_CHAR('\x0170'));
    str16.Concat(U_CHAR('\x0180'));
    str16.Concat(U_CHAR('\x0190'));
    str16.Concat(U_CHAR('\x0200'));
    str16.Concat(U_CHAR('\x0210'));
    str16.Concat(U_CHAR('\x0220'));
    str16.Concat(U_CHAR('\x0230'));
    str16.Concat(U_CHAR('\x0240'));
    str16.Concat(U_CHAR('\x0250'));
    str16.Concat(U_CHAR('\x0260'));
    str16.Concat(U_CHAR('\x0270'));

    Utf8String str8;
    str8.Translate(str16);

    str8 = "This is a beautiful sentence!";
    str16.Translate(str8);
```

xezon:
The first fixup commit is added to the wrong commit.

xezon:
Fixup commits corrected.

PR #300, opened 2021-10-22
Particle System Parse

duncanspumpkin:
Implements the main ParticleSystem::Parse_Particle_System_Definition and also ParticleSystemTemplate::Parse (this second one is not hooked so it is not actually in use atm)

### DIFF
--- a/src/game/client/system/particlesystem/particlesystemplate.cpp
+++ b/src/game/client/system/particlesystem/particlesystemplate.cpp
@@ -79,7 +79,7 @@ FieldParse ParticleSystemTemplate::s_fieldParseTable[] = {
      { "SlavePosOffset", &INI::Parse_Coord3D,  nullptr,  offsetof(ParticleSystemTemplate, m_slavePosOffset) },
      { "PerParticleAttachedSystem", &INI::Parse_AsciiString,  nullptr,  offsetof(ParticleSystemTemplate, m_attachedSystemName) },
      { "Lifetime", &GameClientRandomVariable::Parse,  nullptr,  offsetof(ParticleSystemTemplate, m_lifetime) },
-     { "SystemLifetime", &INI::Parse_Unsigned,  nullptr,  offsetof(ParticleSystemTemplate, m_systemLifetime) },
+     { "SystemLifetime", &INI::Parse_Unsigned_Int,  nullptr,  offsetof(ParticleSystemTemplate, m_systemLifetime) },
      { "Size", &GameClientRandomVariable::Parse,  nullptr,  offsetof(ParticleSystemTemplate, m_startSize) },
      { "StartSizeRate", &GameClientRandomVariable::Parse,  nullptr,  offsetof(ParticleSystemTemplate, m_startSizeRate) },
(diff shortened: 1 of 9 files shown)


src/game/client/system/particlesystem/particlesystemplate.cpp
```diff
@@ -16,6 +16,127 @@
 #include "ini.h"
 #include "particlesysmanager.h"
 
+constexpr const char *_particleShaderTypeNames[] = {
```
xezon: Why do these variable names have a leading underscore? It looks odd.

src/game/client/system/particlesystem/particlesystemplate.h
```diff
@@ -35,13 +36,18 @@ class ParticleSystemTemplate : public MemoryPoolObject, public ParticleSystemInf
...
+    static FieldParse *Get_Particle_System_Template_Field_Parse_Table() { return s_fieldParseTable; }
+
 private:
     ParticleSystem *Create_Slave_System(bool create_slaves) const;
 
 private:
     Utf8String m_name;
     mutable ParticleSystemTemplate *m_slaveTemplate;
+
+    static FieldParse s_fieldParseTable[];
```
xezon, line 52: Is this array meant to be writable on runtime? If not I recommend to make it `const`.

src/game/client/system/particlesystem/particlesystemplate.cpp
duncanspumpkin: I was going by the style guide but on second inspection i didn't follow the correct line and just combined two!
```
Static variables within a function use snake case prefixed by "_".
```
There isn't anything in the style guide for non global static variables. I guess you could say they are still global even though they are local to just this cpp file. Hmm weapon.cpp goes for PascalCase. I'll switch to that.

src/game/client/system/particlesystem/particlesystemplate.h
tomsons26: no, should be const

src/game/client/system/particlesystem/particlesystemplate.cpp
```diff
@@ -45,6 +166,18 @@ void ParticleSystemTemplate::Parse_RGB_Color_Keyframe(INI *ini, void *formal, vo
...
 }
 
+// zh: 0x0041CD00 wb: 0x007A3AE0
+void ParticleSystemTemplate::Parse(INI *ini, void *, void *store, const void *)
+{
+    Utf8String template_name = ini->Get_Next_Token();
+
+    auto *particle_template = g_theParticleSystemManager->Find_Template(template_name);
+
+    captainslog_dbgassert(particle_template != nullptr || template_name.Is_None(), "ParticleSystem %s not found!\n");
```
tomsons26: missing argument for the name

src/game/client/system/particlesystem/particlesystemplate.h
duncanspumpkin: Can't be const as the INI functions require mutable. That i think could be changed but its not an easy fix.

tomsons26 (approved):

PR #301, opened 2021-10-23
Add begin, end iterators to the dynamic vector classes

duncanspumpkin:
This will allow use of ranged based for loops and std algorithims

cbegin and cend versions could also be added.

### DIFF
--- a/src/w3d/lib/simplevec.h
+++ b/src/w3d/lib/simplevec.h
@@ -190,7 +190,11 @@ public:
     }
 
     T *begin() { return &(*this)[0]; }
+    const T *begin() const { return cbegin(); }
+    const T *cbegin() const { return &(*this)[0]; }
     T *end() { return &(*this)[Count()]; }
+    const T *end() const { return cend(); }
+    const T *cend() const { return &(*this)[Count()]; }
 
 protected:
     bool Grow(int new_size_hint)
--- a/src/w3d/lib/simplevec.h
+++ b/src/w3d/lib/simplevec.h
@@ -189,6 +189,9 @@ public:
         }
     }
 
+    T *begin() { return &(*this)[0]; }
+    T *end() { return &(*this)[Count()]; }
+
 protected:
     bool Grow(int new_size_hint)
     {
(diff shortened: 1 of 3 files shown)


src/w3d/lib/simplevec.h
```diff
@@ -189,6 +189,9 @@ template<typename T> class SimpleDynVecClass : public SimpleVecClass<T>
         }
     }
 
+    T *begin() { return &(*this)[0]; }
+    T *end() { return &(*this)[Count()]; }
```
xezon, line 195: Shouldn't there also be a `const` version of these functions returning a `const T` so this can be used with const objects?

src/w3d/lib/vector.h
```diff
@@ -240,6 +240,9 @@ template<typename T> class DynamicVectorClass : public VectorClass<T>
     int Growth_Step() { return m_growthStep; }
     T *Uninitialized_Add();
 
+    T *begin() { return &(*this)[0]; }
+    T *end() { return &(*this)[Count()]; }
```
xezon, line 246: Shouldn't this return reference aka `T&` ? Reason being, if T is not a primitive, we would iterate over copies, which is a) costly and b) erroneous if T class has no Default Constructor.

src/w3d/lib/simplevec.h
duncanspumpkin: Yeah could add cbegin/end as well

src/w3d/lib/vector.h
duncanspumpkin: No it shouldn't a T& isn't an iterator so how could you perform ++, or dereference.

xezon: I failed to see the `*`.

src/w3d/lib/simplevec.h
```diff
@@ -189,6 +189,13 @@ template<typename T> class SimpleDynVecClass : public SimpleVecClass<T>
         }
     }
 
+    T *begin() { return &(*this)[0]; }
```
xezon, line 192: `return &m_vector[0]` would make this a easier on the eyes and debugging.

xezon (approved):

src/w3d/lib/simplevec.h
duncanspumpkin: Looking at it again you can't use the [] operator anyway due to the asserts so will have to make that change for that reason as well!

PR #302, opened 2021-10-23
Add trailing commas (as per style guide) to improve readability

duncanspumpkin:
When implementing a couple of the parse functions I kept not seeing the first item on a list as it would be inline with the name confusing me as to if a list was identical. This makes it much clearer and follows the style guide

```
Leave a trailing ',' in an enum to make it easier for anyone wanting to expand it.
```
Okay these aren't enums but the same reasoning should be applied to the enum's name arrays.

### DIFF
--- a/src/game/common/bitflags.cpp
+++ b/src/game/common/bitflags.cpp
@@ -16,7 +16,8 @@
 
 // TODO move this somewhere more appropriate?
 template<>
-const char *BitFlags<OBJECT_STATUS_COUNT>::s_bitNamesList[] = { "NONE",
+const char *BitFlags<OBJECT_STATUS_COUNT>::s_bitNamesList[] = {
+    "NONE",
     "DESTROYED",
     "CAN_ATTACK",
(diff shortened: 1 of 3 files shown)


src/game/common/bitflags.cpp
```diff
@@ -61,10 +62,12 @@ const char *BitFlags<OBJECT_STATUS_COUNT>::s_bitNamesList[] = { "NONE",
     "IMMOBILE",
     "DISGUISED",
     "DEPLOYED",
-    nullptr };
+    nullptr,
```
xezon, line 65: Isn't `nullptr` supposed to be the array terminator? We would not add an element behind that, no?

duncanspumpkin: Well yes but that's not the point. It just looks better if everything follows the same style.

OmniBlade: Clang format also likes the terminating comma better as well.

OmniBlade (approved):

PR #303, opened 2021-10-23
Implements BWRenderClass

tomsons26:


src/w3d/renderer/bwrender.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+    if (y >= m_minv && y < m_maxv && end_x >= m_minv && start_x < m_maxv) {
+        if (start_x < m_minv) {
+            start_x = m_minv;
+        }
+        if (end_x >= m_maxv) {
+            end_x = m_maxv - 1;
+        }
+        if (end_x != start_x) {
+            memset(&m_buffer[(m_scale * y) + start_x], 0, end_x - start_x);
```
duncanspumpkin: 0 here represents colour. I wonder if perhaps you should add a constant for it.

src/w3d/renderer/bwrender.cpp
```diff
@@ -0,0 +1,216 @@
...
+        }
+    }
+}
+void BWRenderClass::Buffer::Fill(unsigned char c)
```
duncanspumpkin: I think naming this variable `bkgndColor` would be more appropriate

src/w3d/renderer/bwrender.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+    memset(m_buffer, c, m_scale * 2);
+}
+
+BWRenderClass::BWRenderClass(unsigned char *buffer, int scale) : m_pixelBuffer(buffer, scale), m_vertices(nullptr) {}
+
+BWRenderClass::~BWRenderClass() {}
+
+void BWRenderClass::Fill(unsigned char c)
```
duncanspumpkin: same here

src/w3d/renderer/bwrender.cpp
```diff
@@ -0,0 +1,216 @@
...
+                GameMath::Float_To_Int_Floor(corner_3[1]));
+
+            Vector3 xcf(corner_1[0], corner_2[0], corner_3[0]);
+
+            Render_Preprocessed_Triangle(xcf, yci);
+        }
+    }
+}
+
+void BWRenderClass::Render_Preprocessed_Triangle(Vector3 &xcf, Vector3i &yci)
```
duncanspumpkin: these parameter names are very cryptic. Not sure why they stuffed things that arent vectors in vectors. Perhaps `xCorners` and `yCorners` would be easier to understand. The float or int aspect can be told by the type no need to post fix.

src/w3d/renderer/bwrender.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+void BWRenderClass::Set_Vertex_Locations(Vector2 *vertices, int count)
+{
+    m_vertices = vertices;
+    float half_scale = m_pixelBuffer.Scale() / 2.0f;
+    VectorProcessorClass::MulAdd(&vertices[0][0], half_scale, half_scale, 2 * count);
+}
+
+// needs a name
+bool Vector2_4C6400(const Vector2 &v1, const Vector2 &v2, const Vector2 &v3)
```
duncanspumpkin: name the variables `corner1` -> `corner3`

src/w3d/renderer/bwrender.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+    float d0_0 = v2[0] - v1[0];
```
duncanspumpkin: use .X and .Y and it's much easier to see or even just do `Vector2 d0 = v2 - v1;`

src/w3d/renderer/bwrender.cpp
```diff
@@ -0,0 +1,216 @@
...
+            Render_Preprocessed_Triangle(xcf, yci);
+        }
+    }
+}
+
+void BWRenderClass::Render_Triangle_Strip(unsigned int *indices, int index_count)
+{
+    int count = index_count - 2;
+
+    bool bool1 = false;
```
duncanspumpkin, line 126: remove this is constant false.

src/w3d/renderer/bwrender.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+        bool1 = bool1 == 0;
+
+        unsigned int idx_0 = indices[0];
+        unsigned int idx_1 = indices[1];
+        unsigned int idx_2 = indices[2];
+
+        ++indices;
+
+        if (Vector2_4C6400(m_vertices[idx_0], m_vertices[idx_1], m_vertices[idx_2]) != bool1) {
```
duncanspumpkin: This is identical to the above function it would be safe at this time to make a common sub function.

src/w3d/renderer/bwrender.cpp
```diff
@@ -0,0 +1,216 @@
...
+                GameMath::Float_To_Int_Floor(corner_3[1]));
+
+            Vector3 xcf(corner_1[0], corner_2[0], corner_3[0]);
+
+            Render_Preprocessed_Triangle(xcf, yci);
+        }
+    }
+}
+
+void BWRenderClass::Render_Triangle_Strip(unsigned int *indices, int index_count)
```
duncanspumpkin, line 122: Render triangles as a running moving frame through all indices. i.e. corners 0, 1, 2 -> 1, 2, 3 -> 2, 3, 4 ->...

src/w3d/renderer/bwrender.cpp
duncanspumpkin: http://www.sunshine2k.de/coding/java/TriangleRasterization/TriangleRasterization.html has a good documentation of how this algorithm works.

src/w3d/renderer/bwrender.cpp
duncanspumpkin: https://www.scratchapixel.com/lessons/3d-basic-rendering/rasterization-practical-implementation/rasterization-stage
```
bool edgeFunction(const Vec2f &a, const Vec3f &b, const Vec2f &c) 
{ 
    return ((c.x - a.x) * (b.y - a.y) - (c.y - a.y) * (b.x - a.x) >= 0); 
} 
```
This doesnt quite match but its the same idea. The function is trying to check if corner 3 is on the left or right of the line from corner 1 to 2.

src/w3d/renderer/bwrender.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+bool Edge_Function(const Vector2 &corner_1, const Vector2 &corner_2, const Vector2 &corner_3)
+{
+    Vector2 a = corner_2 - corner_1;
+    Vector2 b = corner_3 - corner_1;
+
+    // 2D cross product
+    float cross = a.X * b.Y - a.Y * b.X;
+
+    return cross >= 0.0;
```
xezon, line 71: 0.0f

src/w3d/renderer/bwrender.cpp
```diff
@@ -0,0 +1,211 @@
...
+            left = x_corners[1];
+        }
+    }
+
+    // This represents the height of the flat bottom triangle
+    int height = y_corners[2] - y_corners[1];
+    y = y_corners[1];
+
+    // Flat bottom Triangle Draw
+    if (height) {
```
xezon, line 197: This can be written more explicit aka `if (height != 0)`

src/w3d/renderer/bwrender.cpp
duncanspumpkin: So as discussed this is constantly flipping each iteration, it's not constant. I feel using ^ would make this more obvious it's flipping. The reason it does this we are unsure at this time. I don't understand what situation this code is called.

xezon: Maybe originally there was a `i % 1` and the compiler optimized it this way getting rid of the modulo.

PR #305, opened 2021-10-24
Finish MeshMatDescClass

jonwil:


src/w3d/renderer/meshmatdesc.cpp
```diff
@@ -37,6 +45,25 @@ MatBufferClass::~MatBufferClass()
...
 }
 
+void MatBufferClass::Set_Element(int index, VertexMaterialClass *mat)
+{
+    Ref_Ptr_Set(m_array[index], mat);
+}
+
+VertexMaterialClass *MatBufferClass::Get_Element(int index)
+{
+    if (m_array[index]) {
```
duncanspumpkin: Use explicit nullptr comparison

src/w3d/renderer/meshmatdesc.cpp
```diff
@@ -71,108 +99,299 @@ TextureClass *TexBufferClass::Peek_Element(int index)
...
 }
 
-VertexMaterialClass *MatBufferClass::Get_Element(int index)
+bool UVBufferClass::Is_Equal_To(const UVBufferClass &that)
 {
-    if (m_array[index]) {
-        m_array[index]->Add_Ref();
-    }
-    return m_array[index];
+    return m_CRC == that.m_CRC;
```
duncanspumpkin, line 114: Rather than write this twice it's best that one calls the other. Suggest replacing with *this == that

src/w3d/renderer/meshmatdesc.cpp
```diff
@@ -71,108 +99,299 @@ TextureClass *TexBufferClass::Peek_Element(int index)
...
-    return m_texture[pass][stage];
 }
 
-VertexMaterialClass *MeshMatDescClass::Get_Material(int vidx, int pass) const
+MeshMatDescClass::MeshMatDescClass(const MeshMatDescClass &that) : m_passCount(1), m_vertexCount(0), m_polyCount(0)
 {
-    if (m_materialArray[pass]) {
+    int pass;
+    int stage;
+    int array;
```
duncanspumpkin, line 300: No need to forward declare. Scope to the for loops only

duncanspumpkin (commented):
There seems to be a good bit of duplicated code in this. I wonder if some of it could be deduplicated especially in these copy functions.

src/w3d/renderer/meshmatdesc.cpp
```diff
@@ -71,108 +99,299 @@ TextureClass *TexBufferClass::Peek_Element(int index)
...
-    return m_texture[pass][stage];
+    for (int i = 0; i < alternate_materials.Get_UV_Array_Count(); i++) {
+        Ref_Ptr_Set(m_UV[i], alternate_materials.m_UV[i]);
+    }
+
+    for (int pass = 0; pass < MAX_PASSES; pass++) {
+        for (int stage = 0; stage < MAX_TEX_STAGES; stage++) {
+
+            if (alternate_materials.m_UVSource[pass][stage] == -1) {
+                if (default_materials.m_UVSource[pass][stage] != -1) {
```
xezon, line 441: Can we make the `-1` a named label for readability?

tomsons26 (approved):
Looks good

PR #306, opened 2021-10-24
Add support for sized enum operators

xezon:
I noticed the enum operators do not support custom sizes. For robustness sake I added support same way as seen in winnt.h. I had to move the macros into a new file, otherwise there would have been a cyclic include dependency with bittype.h. I disabled clang-format on the enum macros, because otherwise they just become a huge mess.

In always.h there are some unrelated lines touched because of the clang format. It seems when clang format was introduced in compile check, the files were not updated with this format.

### DIFF
--- a/deps/baseconfig/src/typesize.h
+++ b/deps/baseconfig/src/typesize.h
@@ -1,0 +1,53 @@
+/**
+ * @file
+ *
+ * @author xezon
+ *
+ * @brief Basic header files and defines that are always needed.
+ *
+ * @copyright BaseConfig is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#ifndef BASE_TYPESIZE_H
+#define BASE_TYPESIZE_H
+
+#include "bittype.h"
+
+#ifdef __cplusplus
+extern "C++" {
+
+template<size_t S> struct IntegerForSize;
+
+template<> struct IntegerForSize<1>
+{
+    typedef int8_t type;
+};
+
+template<> struct IntegerForSize<2>
+{
+    typedef int16_t type;
+};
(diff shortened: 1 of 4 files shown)


OmniBlade (approved):

deps/baseconfig/src/typesize.h
```diff
@@ -0,0 +1,53 @@
...
+#define BASE_TYPESIZE_H
+
+#include "bittype.h"
+
+#ifdef __cplusplus
+extern "C++" {
+
+template<size_t S> struct IntegerForSize;
+
+template<> struct IntegerForSize<1>
```
duncanspumpkin, line 25: All of these should be unsigned. There will be some really odd interactions with enums that are at the byte boundaries.
Also it would be neater to use a using statement `using type = int8_t`. http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rt-using

This doesn't follow the coding style either. It should be `type_t`

xezon: signed integer is the same way Windows does it in winnt.h. See DEFINE_ENUM_FLAG_OPERATORS

duncanspumpkin: Which is fine for bitwise operations used in `DEFINE_ENUM_FLAG_OPERATORS`. But its not fine for arithmetic operations which we are doing or the shift operations. 
Consider the following:
```cpp
enum Test : uint8_t
{
    flag0 = (1 << 0),
    flag1 = (1 << 1),
    flag2 = (1 << 2),
    flag3 = (1 << 3),
    flag4 = (1 << 4),
    flag5 = (1 << 5),
    flag6 = (1 << 6),
    flag7 = (1 << 7),
};
DEFINE_ENUMERATION_BITWISE_OPERATORS(Test);


Test a = Test::flag7;
a >>= 1;
// At this point a is Test::flag7 | Test::flag6 which is not what you would expect
```

xezon: Apart from the shifts, which one exactly need unsigned and why? Note: they have all been signed int originally.

PR #307, opened 2021-10-25
FXListStore

duncanspumpkin:
Implemented as far as possible.

### DIFF
--- a/src/game/client/fxlist.cpp
+++ b/src/game/client/fxlist.cpp
@@ -1,0 +1,72 @@
+/**
+ * @file
+ *
+ * @author Duncans_Pumpkin
+ *
+ * @brief FXList objects.
+ *
+ * @copyright Thyme is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#include "fxlist.h"
+
+#ifndef GAME_DLL
+FXListStore *g_theFXListStore = nullptr;
+#endif
+
+// zh: 0x004CC200 wb: 0x0075F326
+FXList *FXListStore::Find_FXList(const char *name)
+{
+    Utf8String str(name);
+
+    if (str.Is_None()) {
+        return nullptr;
+    }
+    const auto key = g_theNameKeyGenerator->Name_To_Key(name);
+    auto res = m_fxmap.find(key);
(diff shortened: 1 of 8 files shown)


duncanspumpkin:
Copy https://github.com/TheAssemblyArmada/Thyme/blob/2f8051f277d4e890b77aa9a647ec00495a6fdea3/src/game/common/audio/audiomanager.h#L64 for hashmap

src/game/rtsutils.h
```diff
@@ -62,6 +62,10 @@ template<typename T> struct equal_to
 {
     bool operator()(const T &left, const T &right) const { return (left.Compare(right) == 0); }
```
xezon: To me it looks like this should be the specialized one, and the one below for `int32_t` should be the default. Then you can also remove the specialization from the other file and just have it here in rtsutils.h

duncanspumpkin: Yeah it's hard to tell without every consumer implemented. I think this is just a utf8 specialisation rather than a T generic version.

xezon: Yep. In std its also a `==` by default. See:
https://en.cppreference.com/w/cpp/utility/functional/equal_to

src/game/client/fxlist.h
```diff
@@ -0,0 +1,88 @@
...
+using fxlistmap_t = std::unordered_map<NameKeyType, FXList, rts::hash<NameKeyType>, std::equal_to<NameKeyType>>;
+#endif
+
+class FXListStore : public SubsystemInterface
+{
+public:
+    virtual ~FXListStore() override {}
+
+    virtual void Init() override {}
+    virtual void PostProcessLoad() override {}
```
jonwil: This class doesn't override PostProcessLoad.

src/game/client/fxlist.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+    const auto key = g_theNameKeyGenerator->Name_To_Key(ini->Get_Next_Token());
+    auto &fx_list = g_theFXListStore->m_fxmap[key];
+
+    fx_list.Clear();
+    ini->Init_From_INI(&fx_list, s_theFXListFieldParse);
+}
+
+// zh: 0x0041CCB0 wb: 0x007A3A51
+void FXList::Parse(INI *ini, void *, void *store, const void *)
```
jonwil, line 62: This already exists as INI::Parse_FX_List. Remove the INI function and replace its use with this one.

jonwil (changes requested):

jonwil (approved):
Looks good now, approved.

PR #309, opened 2021-10-25
Fix various major code issues

xezon:
I ran cppcheck and fixed the biggest offenders. I will have 2 more follow up changes after this, but this one is the toughest to review.

There are 2 changes I am uncertain of:
* Fix uninitialized variable in ScreenMotionBlurFilter::Post_Render
* Fix issues where container iteration would continue with potentially...

And in the first commit I added a few comments to code portions that need someone to fix. I am not sure how to tackle them right now.
* Add comments to code that needs attention

### DIFF
--- a/deps/baseconfig/src/classes.h
+++ b/deps/baseconfig/src/classes.h
@@ -1,0 +1,32 @@
+/**
+ * @file
+ *
+ * @author xezon
+ * 
+ * @brief various Useful Standard Template Library functions
+ *
+ * @copyright Baseconfig is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#ifndef BASE_CLASSES_H
+#define BASE_CLASSES_H
+
+#ifdef __cplusplus
+
+class NotCopyable
+{
+public:
+    NotCopyable() = default;
+
+private:
+    NotCopyable(const NotCopyable &) = delete;
+    NotCopyable &operator=(const NotCopyable &) = delete;
+};
+
+#endif // __cplusplus
+
+#endif // BASE_CLASSES_H
--- a/deps/baseconfig/src/templates.h
+++ b/deps/baseconfig/src/templates.h
@@ -1,0 +1,63 @@
+/**
+ * @file
+ *
+ * @author xezon
+ *
+ * @brief various useful Standard Template Library functions
+ *
+ * @copyright Baseconfig is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#ifndef BASE_TEMPLATES_H
+#define BASE_TEMPLATES_H
+
+#ifdef __cplusplus
+
+#include <algorithm>
+
+namespace stl
+{
+// Unable to put remove_if in std namespace. STL Port will report std::remove_if missing, but at the same time complain if
+// added in std namespace here.
+
+// Sorts elements and returns iterator from where to erase from.
+template<class ForwardIterator, class UnaryPredicate>
+ForwardIterator remove_if(ForwardIterator first, ForwardIterator last, UnaryPredicate pred)
+{
+    ForwardIterator result = first;
+    while (first != last) {
+        if (!pred(*first)) {
+            if (result != first)
+                *result = *first;
+            ++result;
+        }
+        ++first;
+    }
+    return result;
+}
+
+// Erases all elements where predicate is true.
+template<class Container, class UnaryPredicate>
+typename Container::size_type erase_if(Container &container, UnaryPredicate pred)
+{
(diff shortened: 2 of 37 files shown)


deps/baseconfig/src/classes.h
```diff
@@ -0,0 +1,32 @@
...
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#ifndef BASE_CLASSES_H
+#define BASE_CLASSES_H
+
+#ifdef __cplusplus
+
+class NotCopyable
```
OmniBlade, line 20: I would suggest that this needs to live in game/common rather than be a part of baseconfig always.h

deps/baseconfig/src/templates.h
```diff
@@ -0,0 +1,63 @@
...
+#ifndef BASE_TEMPLATES_H
+#define BASE_TEMPLATES_H
+
+#ifdef __cplusplus
+
+#include <algorithm>
+
+namespace stl
+{
+// Unable to put remove_if in std namespace. STL Port will report std::remove_if missing, but at the same time complain if
```
OmniBlade, line 24: We should add std stuff from later C++ iterations to deps/stlport such that it all behaves exactly the same for when we don't use stlport as well. We have it in tree so there is no issue with modifying it as far as I'm concerned.

src/game/client/system/particlesystem/particlesysmanager.cpp
```diff
@@ -353,6 +353,7 @@ void ParticleSystemManager::Remove_Particle_System(ParticleSystem *system)
         if (*it == system) {
             m_allParticleSystemList.erase(it);
             --m_particleSystemCount;
+            break;
```
OmniBlade, line 356: Are we sure this list can't contain duplicates?

src/game/client/videoplayer/videoplayer.cpp
```diff
@@ -110,6 +110,7 @@ void VideoPlayer::Remove_Video(Video *video)
     for (auto it = m_videosAvailableToPlay.begin(); it != m_videosAvailableToPlay.end(); ++it) {
         if (it->internal_name == video->internal_name) {
             m_videosAvailableToPlay.erase(it);
+            break;
```
OmniBlade, line 113: Are we sure this list can't contain duplicates?

src/game/common/system/gamememory.cpp
```diff
@@ -28,7 +28,9 @@ bool g_theMainInitFlag = false;
 #include "hooker.h"
 #endif
 
+#ifndef NDEBUG
```
OmniBlade, line 31: use `#ifdef GAME_DEBUG` which we set for debug builds in the build system.

OmniBlade (changes requested):
In addition to the other suggestions, it might be worth splitting the NotCopyable changes from the rest in a separate PR as there doesn't seem to be dependencies between the two.

xezon:
How do I take out one change? I think it is not possible. I would need to open new Pull Requests for everything.

duncanspumpkin:
You can cherry pick commits into different branches and pr each branch.

xezon:
Yeah that is probably more sane for review sake. It's too much different things here.

deps/baseconfig/src/templates.h
xezon: I don't agree with modifying STL Port. It's more sane to leave it outside. It makes it easier to reason about things we added as opposed to being buried in original third party library. Using regular Standard Library will still be compatible here, because these function will keep on living.

Besides `std::erase_if` is c++20 only.

And `std::fill` has no such version as I added here.

It's ok to have this separate. There will likely be more custom template functions catered to std added here, so it can naturally grow.

src/game/client/system/particlesystem/particlesysmanager.cpp
xezon: I can change to `stl::erase_if`, then it will be safe.

src/game/common/system/gamememory.cpp
xezon: Ok. I wasn't sure. GAME_DEBUG could also refer to native game debug features, so I wanted to not confuse such distinction.

deps/baseconfig/src/templates.h
OmniBlade: I disagree that we should avoid modifying stlport all together as we have the git history to consult and can also label commits as being a modification to stlport in the commit message for oversight.

However in the case that we are adding stuff additional to the nominal C++ version we are targetting, I think I'd prefer such extensions to live under the rts namespace like the hashing methods and other odd bits and bats that exist for interacting with the stl containers already.

src/game/common/system/gamememory.cpp
OmniBlade: GAME_DEBUG is intended to mean the project was built in a debug configuration similar to how NDEBUG works, we also have GAME_DEBUG_STRUCTS which is for enabling the structures and code from the true debug builds such as world builder. I should probably document that somewhere on the wiki.

deps/baseconfig/src/templates.h
xezon: Ok. Yes we can move there. What does "rts" stand for?

src/game/common/system/gamememory.cpp
xezon: A simple `DEBUG` and `RELEASE` would be nice. `NDEBUG` is poor naming, because it is inverted truth, where `#ifndef NDEBUG` becomes a double negative.

PR #310, opened 2021-10-26
Implements LineGroupClass

tomsons26:


src/w3d/renderer/linegroup.cpp
```diff
@@ -0,0 +1,353 @@
...
+            && m_shader.Get_Alpha_Test() == ShaderClass::ALPHATEST_DISABLE && W3D::Is_Sorting_Enabled();
+
+        static Vector3 _offset_a(GameMath::Cos(DEG_TO_RAD(90)), GameMath::Sin(DEG_TO_RAD(90)), 0);
+        static Vector3 _offset_b(GameMath::Cos(DEG_TO_RAD(210)), GameMath::Sin(DEG_TO_RAD(210)), 0);
+        static Vector3 _offset_c(GameMath::Cos(DEG_TO_RAD(330)), GameMath::Sin(DEG_TO_RAD(330)), 0);
+        static Vector3 _offset[3];
+
+        _offset[0].Set(_offset_a);
+        _offset[1].Set(_offset_b);
+        _offset[2].Set(_offset_c);
```
xezon: Could we simplify this to:
```
static Vector3 _offset[3] = {
Vector3(GameMath::Cos(DEG_TO_RAD(90)), GameMath::Sin(DEG_TO_RAD(90)), 0),
Vector3(GameMath::Cos(DEG_TO_RAD(210)), GameMath::Sin(DEG_TO_RAD(210)), 0),
Vector3(GameMath::Cos(DEG_TO_RAD(330)), GameMath::Sin(DEG_TO_RAD(330)), 0)
}
```

src/w3d/renderer/linegroup.cpp
```diff
@@ -0,0 +1,353 @@
...
+            start_vtx.Set(m_startLoc->Get_Element(idx));
+
+            if (m_mode == TETRAHEDRON) {
+
+                vertex->x = end_vtx.X;
+                vertex->y = end_vtx.Y;
+                vertex->z = end_vtx.Z;
+                vertex->diffuse = DX8Wrapper::Convert_Color(tail_diffuse);
+                vertex->u1 = u_coord;
+                vertex->v1 = 1.0;
```
xezon: `1.0f`, also some more below.

src/w3d/renderer/linegroup.h
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+    bool Get_Flag(FlagsType flag) const { return (m_flags >> flag) & 1; }
+
+    void Set_Texture(TextureClass *texture);
+    TextureClass *Get_Texture() { return m_texture; }
+    TextureClass *Peek_Texture() { return m_texture; }
+
+    void Set_Shader(const ShaderClass &shader){ m_shader = shader;}
+    ShaderClass Get_Shader() const {return  m_shader;}
```
xezon: Spacing.
Also, return copy of `m_shader`? If Shader class is complex, better return `const ShaderClass&`.

src/w3d/renderer/linegroup.cpp
```diff
@@ -0,0 +1,353 @@
...
+        } else {
+            captainslog_assert(0);
+        }
+
+        DynamicVBAccessClass vertex_buffer(
+            sort ? VertexBufferClass::BUFFER_TYPE_DYNAMIC_SORTING : VertexBufferClass::BUFFER_TYPE_DYNAMIC_DX8,
+            D3DFVF_XYZ | D3DFVF_NORMAL | D3DFVF_DIFFUSE | D3DFVF_TEX2,
+            vertex_count);
+
+        DynamicVBAccessClass::WriteLockClass vertex_lock(&vertex_buffer);
```
xezon, line 258: Does the lock need to be lifted before passing it into `DX8Wrapper::Set_Vertex_Buffer(vertex_buffer)` and calling `DX8Wrapper::Draw_Triangles` further down below? Right now the lock is held beyond that point if I interpret this correctly (`WriteLockClass` being RAII unlock on destruction).

I am asking because:
https://docs.microsoft.com/en-us/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3dvertexbuffer9-lock

> As a general rule, do not hold a lock across more than one frame. When working with vertex buffers, you are allowed to make multiple lock calls; however, you must ensure that the number of lock calls match the number of unlock calls. DrawPrimitive calls will not succeed with any outstanding lock count on any currently set vertex buffer.

tomsons26: Good catch, it would had been in a branch as ctor gets called at the end of the code block, the compiler optimized it out

src/w3d/renderer/linegroup.cpp
tomsons26: _offset gets modified, the static init is so the init happens only once, and from there on it can restore the offset

src/w3d/renderer/linegroup.h
tomsons26: original returns instance, not sure if this won't affect anything so not changing it

xezon: In that case can just add a `#TODO check if return copy is ok` comment, so we can look at it later again when we know more.

src/w3d/renderer/linegroup.cpp
```diff
@@ -0,0 +1,358 @@
...
+        DynamicIBAccessClass index_buffer(
+            sort ? IndexBufferClass::BUFFER_TYPE_DYNAMIC_SORTING : IndexBufferClass::BUFFER_TYPE_DYNAMIC_DX8, index_count);
+        { // added to control the lifetime of the WriteLockClass object
+            DynamicIBAccessClass::WriteLockClass index_lock(&index_buffer);
+
+            unsigned short *index = index_lock.Get_Index_Array();
+
+            if (m_mode == TETRAHEDRON) {
+                for (int i = 0; i < m_lineCount; ++i) {
+                    *index++ = short(4 * i + 0);
```
duncanspumpkin: i think it would help to split this up into the vertex's. So just put an additional return after each 3 so that you end up with 4 vertex's

src/w3d/renderer/linegroup.cpp
```diff
@@ -0,0 +1,358 @@
...
+        } else {
+            DX8Wrapper::Set_Transform(D3DTS_VIEW, identity);
+        }
+
+        int polygon_count = 0;
+        int index_count = 0;
+        int vertex_count = 0;
+
+        if (m_mode == TETRAHEDRON) {
+            polygon_count = 4 * m_lineCount;
```
duncanspumpkin: It would be useful to extract out the constants to somewhere to ease understanding:

```cpp
constexpr int32_t TETRAHEADRON_NUM_POLYGONS = 4;
constexpr int32_t TETRAHEADRON_NUM_VERTEXS = 4;

constexpr int32_t PRISM_NUM_POLYGONS = 8;
constexpr int32_t PRISM_NUM_VERTEXS = 6;
```
Then the other constants in this function would make more sense such as at L194

src/w3d/renderer/linegroup.cpp
```diff
@@ -0,0 +1,358 @@
...
+void LineGroupClass::Set_Arrays(ShareBufferClass<Vector3> *start_locs,
+    ShareBufferClass<Vector3> *end_locs,
+    ShareBufferClass<Vector4> *diffuse,
+    ShareBufferClass<Vector4> *tail_diffuse,
+    ShareBufferClass<unsigned int> *alt,
+    ShareBufferClass<float> *sizes,
+    ShareBufferClass<float> *u_coords,
+    int count)
+{
+    captainslog_assert(start_locs);
```
duncanspumpkin: These should all be explicit `nullptr` checks.

src/w3d/renderer/linegroup.h
```diff
@@ -0,0 +1,114 @@
...
+#include "always.h"
+#include "shader.h"
+#include "sharebuf.h"
+#include "vector3.h"
+#include "vector4.h"
+
+class RenderInfoClass;
+class TextureClass;
+
+class LineGroupClass
```
duncanspumpkin, line 26: Are there no hooks required for this?

src/w3d/renderer/linegroup.cpp
```diff
@@ -0,0 +1,358 @@
...
+    if (texture != nullptr) {
+        texture->Add_Ref();
+    }
+    m_texture = texture;
+}
+
+void LineGroupClass::Render(RenderInfoClass &rinfo)
+{
+#ifdef BUILD_WITH_D3D8
+    if (m_lineCount > 0) {
```
duncanspumpkin: This whole function is inside of this if. Suggest we return early.

src/w3d/renderer/linegroup.cpp
```diff
@@ -0,0 +1,358 @@
...
+        m_shader.Set_Cull_Mode(ShaderClass::CULL_MODE_DISABLE);
```
duncanspumpkin: Think you've made a mistake here `CULL_MODE_ENABLE`

src/w3d/renderer/linegroup.cpp
```diff
@@ -0,0 +1,358 @@
...
+        } else {
+            captainslog_assert(0);
+        }
+
+        DynamicIBAccessClass index_buffer(
+            sort ? IndexBufferClass::BUFFER_TYPE_DYNAMIC_SORTING : IndexBufferClass::BUFFER_TYPE_DYNAMIC_DX8, index_count);
+        { // added to control the lifetime of the WriteLockClass object
+            DynamicIBAccessClass::WriteLockClass index_lock(&index_buffer);
+
+            unsigned short *index = index_lock.Get_Index_Array();
```
duncanspumpkin: you define this as uint16_t but then use it as an `int16_t`. This will needlessly cause conversions stick to `uint16_t`

src/w3d/renderer/linegroup.cpp
tomsons26: indeed..

src/w3d/renderer/linegroup.h
tomsons26: uh so this is only called by w3d particle classes, meaning we can only test it when w3d particles are in and @jonwil can do a battery of tests as he's the most familiar with this system.
So im not sure should i bother at this time as none of this would be called anyway..

duncanspumpkin (approved):
All looks good.

PR #311, opened 2021-10-26
MeshGeometryClass

jonwil:


src/w3d/renderer/w3d_file.h
```diff
@@ -464,3 +464,95 @@ struct W3dMeshAABTreeNode
     uint32_t FrontOrPoly0;
     uint32_t BackOrPolyCount;
 };
+
+#define SORT_LEVEL_NONE 0
+#define MAX_SORT_LEVEL 32
+#define SORT_LEVEL_BIN1 20
+#define SORT_LEVEL_BIN2 15
+#define SORT_LEVEL_BIN3 10
```
xezon, line 473: Could we use another way to set these? constexpr, enum?

src/w3d/renderer/w3d_file.h
```diff
@@ -464,3 +464,95 @@ struct W3dMeshAABTreeNode
...
+    char MeshName[16];
+    char ContainerName[16];
+
+    uint32_t NumTris;
+    uint32_t NumVertices;
+    uint32_t NumMaterials;
+    uint32_t NumDamageStages;
+    int32_t SortLevel;
+    uint32_t PrelitVersion;
+    uint32_t FutureCounts[1];
```
xezon, line 489: Array of 1. Is this intentional?

src/w3d/renderer/meshgeometry.cpp
```diff
@@ -158,3 +167,1012 @@ MeshGeometryClass &MeshGeometryClass::operator=(const MeshGeometryClass &that)
...
+    m_vertex(nullptr),
+    m_vertexNorm(nullptr),
+    m_planeEq(nullptr),
+    m_vertexShadeIdx(nullptr),
+    m_vertexBoneLink(nullptr),
+    m_boundBoxMin(0, 0, 0),
+    m_boundBoxMax(1, 1, 1),
+    m_boundSphereCenter(0, 0, 0),
+    m_boundSphereRadius(1),
+    m_cullTree(nullptr)
```
xezon, line 213: It looks like `operator=` function already sets most if not all of these. So it would not be strictly necessary to have this initializer list.

src/w3d/renderer/meshgeometry.cpp
```diff
@@ -158,3 +167,1012 @@ MeshGeometryClass &MeshGeometryClass::operator=(const MeshGeometryClass &that)
...
+
+        m_poly->Clear();
+        m_polySurfaceType->Clear();
+        m_vertex->Clear();
+
+        m_vertexNorm = new ShareBufferClass<Vector3>(m_vertexCount);
+        m_vertexNorm->Clear();
+    }
+
+    return;
```
xezon: Obsolete in C++

src/w3d/renderer/meshgeometry.cpp
```diff
@@ -158,3 +167,1012 @@ MeshGeometryClass &MeshGeometryClass::operator=(const MeshGeometryClass &that)
...
+    const Vector3 *loc = Get_Vertex_Array();
+    const Vector4 *norms = Get_Plane_Array();
+    const TriIndex *polys = Get_Polygon_Array();
+    TriClass tri;
+
+    for (int poly_counter = 0; poly_counter < m_polyCount; poly_counter++) {
+        tri.V[0] = &(loc[polys[poly_counter][0]]);
+        tri.V[1] = &(loc[polys[poly_counter][1]]);
+        tri.V[2] = &(loc[polys[poly_counter][2]]);
+        tri.N = (Vector3 *)&(norms[poly_counter]);
```
xezon, line 316: Is this a `static_cast` or `reinterpret_cast` ? If `reinterpret_cast`, then I would prefer to see a `static_assert` nearby guaranteeing this crap. But Hopyfully Vector3 is a parent of Vector4. Then it's all good and we can write `static_cast` here.

The same thing repears a few times below.

src/w3d/renderer/meshgeometry.cpp
```diff
@@ -158,3 +167,1012 @@ MeshGeometryClass &MeshGeometryClass::operator=(const MeshGeometryClass &that)
...
+            if (boxtest.m_result->compute_contact_point) {
+                Matrix3D::Transform_Vector(transform, obbox.m_result->contact_point, &(obbox.m_result->contact_point));
+            }
+        }
+    }
+
+    return hit;
+}
+
+int MeshGeometryClass::Cast_Semi_Infinite_Axis_Aligned_Ray(const Vector3 &start_point, int axis_dir, unsigned char &flags)
```
xezon, line 506: `unsigned char &flags` is meant to be reference?

src/w3d/renderer/meshgeometry.cpp
```diff
@@ -158,3 +167,1012 @@ MeshGeometryClass &MeshGeometryClass::operator=(const MeshGeometryClass &that)
...
+        flags = TRI_RAYCAST_FLAG_NONE;
+        int poly_count = Get_Polygon_Count();
+
+        for (int poly_counter = 0; poly_counter < poly_count; poly_counter++) {
+            const Vector3 &v0 = loc[polyverts[poly_counter][0]];
+            const Vector3 &v1 = loc[polyverts[poly_counter][1]];
+            const Vector3 &v2 = loc[polyverts[poly_counter][2]];
+            const Vector4 &tri_plane = plane[poly_counter];
+
+            count += (unsigned int)Cast_Semi_Infinite_Axis_Aligned_Ray_To_Triangle(v0,
```
xezon, line 533: `static_cast<unsigned int>`

src/w3d/renderer/meshgeometry.cpp
```diff
@@ -158,3 +167,1012 @@ MeshGeometryClass &MeshGeometryClass::operator=(const MeshGeometryClass &that)
...
+
+    if (verts == nullptr) {
+        verts = m_vertex->Get_Array();
+    }
+
+    VectorProcessorClass::MinMax(verts, m_boundBoxMin, m_boundBoxMax, m_vertexCount);
+
+    m_boundSphereCenter = (m_boundBoxMin + m_boundBoxMax) / 2.0f;
+    m_boundSphereRadius = (float)(m_boundBoxMax - m_boundSphereCenter).Length2();
+    m_boundSphereRadius = ((float)GameMath::Sqrt(m_boundSphereRadius)) * 1.00001f;
```
xezon, line 831: Why is this
a) `Length2` + `Sqrt` and not `Length`?
b) Why * `1.00001f`?

If reasons are known, maybe add a comment. Otherwise it just looks odd.

src/w3d/renderer/meshgeometry.cpp
```diff
@@ -158,3 +167,1012 @@ MeshGeometryClass &MeshGeometryClass::operator=(const MeshGeometryClass &that)
...
+
+    W3dMeshHeader3Struct header;
+
+    if (cload.Read(&header, sizeof(W3dMeshHeader3Struct)) != sizeof(W3dMeshHeader3Struct)) {
+        goto Error;
+    }
+
+    cload.Close_Chunk();
+    char *tmpname;
+    int namelen;
```
xezon, line 868: These variable declarations can be merged with the lines below.

src/w3d/renderer/meshgeometry.cpp
```diff
@@ -158,3 +167,1012 @@ MeshGeometryClass &MeshGeometryClass::operator=(const MeshGeometryClass &that)
...
+
+    Reset_Geometry(header.NumTris, header.NumVertices);
+
+    namelen = strlen(header.ContainerName);
+    namelen += strlen(header.MeshName);
+    namelen += 2;
```
xezon, line 874: Oh boi. I wonder if the original was really written like this. Maybe they had some inline helper function for such string builds.

src/w3d/renderer/meshgeometry.cpp
```diff
@@ -158,3 +167,1012 @@ MeshGeometryClass &MeshGeometryClass::operator=(const MeshGeometryClass &that)
...
+            return W3D_ERROR_LOAD_FAILED;
+        }
+        shade_index[i] = si;
+    }
+    return W3D_ERROR_OK;
+}
+
+void MeshGeometryClass::Get_Deformed_Vertices(Vector3 *dst_vert, Vector3 *dst_norm, const HTreeClass *htree)
+{
+    int vi;
```
xezon: `vi` can be moved into the for loop like the other functions below.

src/w3d/renderer/meshgeometry.cpp
```diff
@@ -158,3 +167,1012 @@ MeshGeometryClass &MeshGeometryClass::operator=(const MeshGeometryClass &that)
...
+W3DErrorType MeshGeometryClass::Read_Chunks(ChunkLoadClass &cload)
+{
+    while (cload.Open_Chunk()) {
+        W3DErrorType error = W3D_ERROR_OK;
+
+        switch (cload.Cur_Chunk_ID()) {
+            case W3D_CHUNK_VERTICES:
+                error = Read_Vertices(cload);
+                break;
+            case W3D_CHUNK_VERTEX_NORMALS:
```
tomsons26, line 954: also W3D_CHUNK_SURRENDER_NORMALS

src/w3d/renderer/meshgeometry.cpp
```diff
@@ -158,3 +167,1012 @@ MeshGeometryClass &MeshGeometryClass::operator=(const MeshGeometryClass &that)
...
+}
+
+bool MeshGeometryClass::Intersect_OBBox_Brute_Force(OBBoxIntersectionTestClass &localtest)
+{
+    TriClass tri;
+    const Vector3 *loc = Get_Vertex_Array();
+    const TriIndex *polyverts = Get_Polygon_Array();
+    const Vector4 *norms = Get_Plane_Array();
+
+    for (int srtri = 0; srtri < Get_Polygon_Count(); srtri++) {
```
tomsons26: all srtri , just use i, no need for a cryptic name

src/w3d/renderer/meshgeometry.cpp
tomsons26: Its nasty but this is what they did unfortunately, multiple binaries confirm it

src/w3d/renderer/w3d_file.h
tomsons26: Yes

src/w3d/renderer/meshgeometry.cpp
xezon: Ok. In that case we need a `static_assert` here that makes sure the memory layout of Vector4 and Vector3 is identical for x y z. This way its properly protected and documented. Plus the C cast should become a `reinterpret_cast`.

src/w3d/renderer/meshgeometry.cpp
tomsons26: yes

tomsons26 (approved):

PR #313, opened 2021-10-26
Fix various problems in cpudetect

xezon:


src/w3d/lib/cpudetect.cpp
```diff
@@ -1124,7 +1124,10 @@ void CPUDetectClass::Init_Processor_String()
         strtrim(ProcessorString);
     } else {
         char str[2048];
-        strcat(str, Get_Processor_Manufacturer_Name());
+        str[0] = '\0';
```
duncanspumpkin: This wont be a hot function just zero initialise the buffer.

src/w3d/lib/cpudetect.cpp
```diff
@@ -1124,7 +1124,10 @@ void CPUDetectClass::Init_Processor_String()
...
+        if (const char *name = Get_Processor_Manufacturer_Name()) {
```
duncanspumpkin: this is horrible. Don't do assignments and comparisions in an if. If you are going to do it use the syntax provided by the language:
```cpp
if (const auto* name = Get_Processor_Manufacturer_Name(); name != nullptr) {
```

xezon: It's perfectly valid syntax. No need to over-engineer.

src/w3d/lib/cpudetect.cpp
```diff
@@ -1374,7 +1384,7 @@ bool CPUDetectClass::CPUID(uint32_t &u_eax_, uint32_t &u_ebx_, uint32_t &u_ecx_,
         return false;
     }
 
-    int32_t regs[4];
+    int32_t regs[4] = { 0 };
```
duncanspumpkin, line 1388: This isn't C99 code zero intialise in c++ with `{}`

src/w3d/lib/cpudetect.cpp
```diff
@@ -1395,7 +1405,7 @@ bool CPUDetectClass::CPUID_Count(
...
```
duncanspumpkin, line 1409: as above

src/w3d/lib/cpudetect.cpp
xezon: Ok.

src/w3d/lib/cpudetect.cpp
xezon: This is a matter of taste I guess. `{ 0 }` is perfectly valid c++, has always been, and is more expressive than `{}`. Can stay as is.

src/w3d/lib/cpudetect.cpp
duncanspumpkin: Perfectly valid syntax that is so often incorrect that there are warnings (-Wparentheses) designed to prevent people  using it.

xezon: Could you please give an example where a pointer check would fail with this legacy approach? I am not aware of such case.

src/w3d/lib/cpudetect.cpp
duncanspumpkin: It just looks old fashioned to use the C version instead of the C++ version.

src/w3d/lib/cpudetect.cpp
duncanspumpkin: This instance is fine its when people expect to do a comparison and accidentally do an assignment that the -Wparentheses is designed to catch. There is also now no explicit comparison to nullptr.

xezon: Ah. Yeah I am also not a fan of fancy expressions in the `if ()` or `while ()`. But this is perfectly safe and solid. There is no need to worry.

OmniBlade: I agree that this is perfectly safe, but it does fail our style guide of being explicit on nullptr checks and not treating them like a bool.

I believe `(const char *name = Get_Processor_Manufacturer_Name()) != nullptr` would also work if you didn't want to be quite as verbose in the if.

xezon: Changed. Does not look more readable, but adheres to style guide as requested.

OmniBlade (approved):

PR #314, opened 2021-10-26
Avoid reading beyond buffer size in User_Memory_Init_Pools function

xezon:


src/game/common/system/gamememoryinit.cpp
```diff
@@ -680,6 +680,9 @@ void User_Memory_Init_Pools()
     int initial_alloc;
     int overflow_alloc;
 
+    // #FIX Zero terminate buffer.
+    pool_name[0] = (char)0;
```
duncanspumpkin: I can't see this being a hot function may as well zero intialise both path and pool_name.

xezon: Oke no problem. I initialized all of them.

OmniBlade (approved):

PR #315, opened 2021-10-26
Fix issues in ScienceInfo and RankInfoStore where container iteration…

xezon:
… would continue with potentially invalidated iterator. I hope the rewrite makes sense. But looks most logical to me this way. I tried debugging, but did not find a case where it would do an erase.

### DIFF
--- a/src/game/common/rts/science.cpp
+++ b/src/game/common/rts/science.cpp
@@ -15,6 +15,7 @@
 #include "science.h"
 #include "ini.h"
 #include "namekeygenerator.h"
+#include "rtsutils.h"
 #include <algorithm>
 #include <captainslog.h>
 #include <cstddef>
@@ -61,15 +62,13 @@ void ScienceInfo::Add_Root_Sciences(std::vector<ScienceType> &rootScience) const
 
 void ScienceStore::Reset()
 {
-    for (auto it = m_infoVec.begin(); it != m_infoVec.end(); ++it) {
-        if (*it != nullptr) {
-            *it = reinterpret_cast<ScienceInfo *>((*it)->Delete_Overrides());
+    // #BUGFIX Does no longer iterate with invalidated iterator.
+    rts::erase_if(m_infoVec, [](ScienceInfo *science) {
+        if (science != nullptr) {
+            science = static_cast<ScienceInfo *>(science->Delete_Overrides());
         }
-
-        if (*it == nullptr) {
-            m_infoVec.erase(it);
-        }
-    }
+        return (science == nullptr);
+    });
 }
 
 ScienceType ScienceStore::Lookup_Science(const char *name)
--- a/src/game/logic/system/rankinfo.cpp
+++ b/src/game/logic/system/rankinfo.cpp
@@ -14,6 +14,7 @@
  */
 #include "rankinfo.h"
 #include "ini.h"
+#include "rtsutils.h"
 #include <captainslog.h>
 #include <cstddef>
 
@@ -38,16 +39,13 @@ void RankInfoStore::Init()
  */
 void RankInfoStore::Reset()
 {
-    auto it = m_infoStore.begin();
-    while (it != m_infoStore.end()) {
-        if (*it != nullptr) { // Original code has this check but result will be infinte loop if it is null?
-            if ((*it)->Delete_Overrides() == nullptr) {
-                m_infoStore.erase(it);
-            } else {
-                ++it;
-            }
+    // #BUGFIX Does no longer iterate with invalidated iterator.
+    rts::erase_if(m_infoStore, [](RankInfo *rank) {
+        if (rank != nullptr) {
+            rank = static_cast<RankInfo *>(rank->Delete_Overrides());
         }
-    }
+        return (rank == nullptr);
+    });
 }
 
 /**
--- a/src/game/rtsutils.h
+++ b/src/game/rtsutils.h
@@ -17,6 +17,7 @@
 #include "always.h"
 #include "asciistring.h"
 #include "endiantype.h"
+#include <algorithm>
 #include <ctime>
 
 #ifdef PLATFORM_WINDOWS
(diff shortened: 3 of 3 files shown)


src/game/common/rts/science.cpp
```diff
@@ -42,15 +43,13 @@ ScienceInfo::ScienceInfo() :
...
 {
-    for (auto it = m_infoVec.begin(); it != m_infoVec.end(); ++it) {
-        if (*it != nullptr) {
-            *it = reinterpret_cast<ScienceInfo *>((*it)->Delete_Overrides());
-        }
-
-        if (*it == nullptr) {
-            m_infoVec.erase(it);
+    // #BUGFIX Does no longer iterate with invalidated iterator.
+    rts::erase_if(m_infoVec, [](ScienceInfo *science) {
```
tomsons26, line 66: and how does this call delete overrides on all of them...?

src/game/logic/system/rankinfo.cpp
```diff
@@ -38,16 +39,13 @@ void RankInfoStore::Init()
...
-    while (it != m_infoStore.end()) {
-        if (*it != nullptr) { // Original code has this check but result will be infinte loop if it is null?
-            if ((*it)->Delete_Overrides() == nullptr) {
-                m_infoStore.erase(it);
-            } else {
-                ++it;
-            }
+    // #BUGFIX Does no longer iterate with invalidated iterator.
+    rts::erase_if(m_infoStore, [](RankInfo *rank) {
+        if (rank != nullptr) {
```
tomsons26, line 44: and how does this call delete overrides on all of them...?

src/game/common/rts/science.cpp
xezon: I don't understand question. 2 lines below it calls `Delete_Overrides()`. It's a lambda passed to `erase_if` function that is called on each container element.

duncanspumpkin:
There is one of two issues I have with this PR:

1. I get that these are meant to reflect the c++17 library feature but they should be renamed. `remove_if` and `erase_if` the predicate is not allowed to modify state its meant to only query state.
> unary predicate which returns ​true if the element should be removed.
The expression p(v) must be convertible to bool for every argument v of type (possibly const) VT, where VT is the value type of ForwardIt, regardless of [value category](https://en.cppreference.com/w/cpp/language/value_category), and must not modify v. Thus, a parameter type of VT&is not allowed, nor is VT unless for VT a move is equivalent to a copy (since C++11).​

2.  Alternatively the callers should be modified so that the delete overrides do not happen in the predicate. This would probably take the form of two loops one for deleting overrides and one for erasing from the vector.

xezon:
I don't understand what you say. I based off of here:
https://en.cppreference.com/w/cpp/algorithm/remove

duncanspumpkin:
> I don't understand what you say. I based off of here: https://en.cppreference.com/w/cpp/algorithm/remove

Yup and although it works fine for this application if in the future this is switched to std then it could potentially perform unexpectedly as the predicate is modifying state.

xezon:
Ok. I need to check if this is problem.

duncanspumpkin:
https://timsong-cpp.github.io/cppwp/n4861/algorithms.requirements#7 this is the clause that it violates. As said its not an issue if we are defining our own implementation but if we ever switch to std::remove_if or std::erase_if then this will need to be modified as it is not conforming.

xezon:
I look at clause 7 and my English is not good enough to comprehend the way it is worded.

I tested your suggestion with godbolt, and mutable lambda argument compiles on both latest clang and MSVC. So it looks like you ask for something here that even the clang and MSVC don't do. I also wouldn't know how to word that syntactically in C++. I guess you would need to pass const iterators to the erase_if? But even if so, why would that make any sense? Shouldn't it be up to the user do to whatever he needs in his algorithm? I think this clause, if it says what you said it says, makes no sense.

https://godbolt.org/z/q9GWzbzcj

```c++
#include <iostream>
#include <vector>
#include <algorithm>

struct A
{
    int val = 0;

    void Set() {
        std::cout << "Change val" << std::endl;
        val = 1;
    }
};

int main() {
    
    std::vector<A> vec = { A(), A(), A() };
    std::cout << "vec size: " << vec.size() << std::endl;
    std::erase_if(vec, [](A& a){ a.Set(); return true; });
    std::cout << "vec size: " << vec.size() << std::endl;
}
```

duncanspumpkin:
> I look at clause 7 and my English is not good enough to comprehend the way it is worded.
> 
> I tested your suggestion with godbolt, and mutable lambda argument compiles on both latest clang and MSVC. So it looks like you ask for something here that even the clang and MSVC don't do. I also wouldn't know how to word that syntactically in C++. I guess you would need to pass const iterators to the erase_if? But even if so, why would that make any sense? Shouldn't it be up to the user do to whatever he needs in his algorithm? I think this clause, if it says what you said it says, makes no sense.
> 
> https://godbolt.org/z/q9GWzbzcj
> 
> ```c++
> #include <iostream>
> #include <vector>
> #include <algorithm>
> 
> struct A
> {
>     int val = 0;
> 
>     void Set() {
>         std::cout << "Change val" << std::endl;
>         val = 1;
>     }
> };
> 
> int main() {
>     
>     std::vector<A> vec = { A(), A(), A() };
>     std::cout << "vec size: " << vec.size() << std::endl;
>     std::erase_if(vec, [](A& a){ a.Set(); return true; });
>     std::cout << "vec size: " << vec.size() << std::endl;
> }
> ```

The reason this comes up is that the standard allows for the lambda to be called more than once per element. I don't know of any implementation that does that for `erase_if` but its still possible. 

So yes MSVC and gcc wont warn on it because there is nothing wrong with it as it stands. Its just we are violating the contract of the standard so in the future it might cause a warning or it might cause the wrong result as it will call the predicate twice.

xezon:
Ok. I think we do not need to worry then. In rts namespace it will be separate from std:: and it does the exact same as current implementation from major compilers. I will make one fix though and add the rts::move as well.

xezon:
Tested and worked.

duncanspumpkin:
You could at least add a note next to the user of it saying that it does not conform to std::erase_if's contract then

xezon:
Ok. I added comment, not that anyone really cares, because this is how everyone uses remove_if anyway :P

PR #316, opened 2021-10-26
Avoid cppcheck warning about non-evaluated expression in comma separa…

xezon:
…ted condition

### DIFF
--- a/src/platform/audio/milesaudiomanager.cpp
+++ b/src/platform/audio/milesaudiomanager.cpp
@@ -974,8 +974,12 @@ void MilesAudioManager::Release_Bink_Handle()
  */
 void MilesAudioManager::friend_Force_Play_Audio_Event(AudioEventRTS *event)
 {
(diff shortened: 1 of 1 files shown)


src/platform/audio/milesaudiomanager.cpp
```diff
@@ -974,8 +974,11 @@ void MilesAudioManager::Release_Bink_Handle()
  */
 void MilesAudioManager::friend_Force_Play_Audio_Event(AudioEventRTS *event)
 {
-    if (event->Get_Event_Info() == nullptr && (Get_Info_For_Audio_Event(event), event->Get_Event_Info() == nullptr)) {
-        return;
+    if (event->Get_Event_Info() == nullptr) {
+        Get_Info_For_Audio_Event(event);
+        if (event->Get_Event_Info() == nullptr) {
+            return;
```
tomsons26, line 981: can you add a debug warning here before the return that WB has?
`"No info for forced audio event '%s'", event->Get_Event_Name();`

tomsons26 (approved):

PR #317, opened 2021-10-26
Fix uninitialized variable in ScreenMotionBlurFilter::Post_Render

xezon:
Jonathan wrote `true` is good. And it also makes sense, because it is set to false in all branches below exclusively.


tomsons26 (commented):
yea seems we missed that
probably check the others too if you can

xezon:
Cppcheck would have warned if there was another.

tomsons26 (approved):

PR #319, opened 2021-10-26
Fix issue in TriClass::Point_In_Triangle_2D where wrong length is cal…

xezon:
…culated and thus certain code branch is never reached.

Fix proposal. It is not clear what the consequence of this is in the game, but it makes no sense that there is a branch with code that can never be reached. Logically this change makes sense. We need to verify this from math perspective and then it should be good to go.

### DIFF
--- a/src/w3d/math/tri.h
+++ b/src/w3d/math/tri.h
@@ -95,7 +95,7 @@ inline bool Point_In_Triangle_2D(const Vector3 &tri_point0,
     } else {
         float p0p1dist2 = p0p1.Length2();
         float p1p2dist2 = p1p2.Length2();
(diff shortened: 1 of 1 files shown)


tomsons26:
This is in fact how it is in the game, and wasn't even fixed in BFME2 so this needs careful checking that it doesn't break any game collision checks, bridge code, and terrain code

xezon:
Don't trust the game code to be correct! The code is broken one way or another, because one branch can never be reached right now (why is it there then?). I do know there is issue where units can fall through bridges. Maybe this is related.

xezon:
I asked community to provide video or replay of bridge bugs.

PR #320, opened 2021-10-26
Fix potential null dereference in TextureLoadTaskListClass::Remove

xezon:


src/w3d/renderer/textureloadtasklist.cpp
```diff
@@ -107,7 +107,9 @@ TextureLoadTaskClass *TextureLoadTaskListClass::Pop_Back()
  */
 void TextureLoadTaskListClass::Remove(TextureLoadTaskClass *task)
 {
-    if (task->m_parent == this) {
+    // #BUGFIX Avoid task null dereference.
+
+    if (task && task->m_parent == this) {
```
duncanspumpkin: `!= nullptr`

xezon: Ok.

tomsons26 (approved):

PR #321, opened 2021-10-26
Provide MotionChannelClass::Get_Vector_Safe function to help avoid mi…

xezon:
…suse and detect potential problems.

This is a proposal to discourage the use of the original unchecked function `Get_Vector` which allows for non-obvious memory corruptions. The new functions make corruption less likely, because it is checked in Debug.

Likely there is an even better solution, but it did not occur to me yet.

### DIFF
--- a/src/w3d/renderer/hrawanim.cpp
+++ b/src/w3d/renderer/hrawanim.cpp
@@ -192,15 +192,15 @@ void HRawAnimClass::Get_Translation(Vector3 &trans, int pividx, float frame) con
         Vector3 t0;
 
         if (motion->X) {
-            motion->X->Get_Vector(frame0, &t0.X);
+            motion->X->Get_Vector_Safe(frame0, &t0.X);
         }
 
         if (motion->Y) {
-            motion->Y->Get_Vector(frame0, &t0.Y);
+            motion->Y->Get_Vector_Safe(frame0, &t0.Y);
         }
 
         if (motion->Z) {
-            motion->Z->Get_Vector(frame0, &t0.Z);
+            motion->Z->Get_Vector_Safe(frame0, &t0.Z);
         }
 
         if (delta == 0) {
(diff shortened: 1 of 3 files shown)


src/w3d/renderer/hrawanim.cpp
```diff
@@ -192,15 +192,15 @@ void HRawAnimClass::Get_Translation(Vector3 &trans, int pividx, float frame) con
         Vector3 t0;
 
         if (motion->X) {
-            motion->X->Get_Vector(frame0, &t0.X);
+            motion->X->Get_Vector_Safe(frame0, &t0.X);
```
xezon, line 195: I need to check if any of these actually is above 1 float...

tomsons26:
this doesn't make sense to me
what original does is process frames starting from a frame offset, not a single frame like what you made it do

xezon:
Yeah it's possible there is error. I need to look at it again.

xezon:
I looked into it again. The design is just terrible. I do not yet understand what the motivation was for this design.

Look at:
```c++
motion->Z->Get_Vector(frame1, &t1.Z);
```
If `motion->Z` had a vector len > 1, then we would fill `t1` past `Z`, essentially corrupting memory Z + 1. This is so dirty, you need the whole city cleaning staff to make it tidy.

The goal with changing this would be to make the vector read explicit, so that writes past the output vector cannot happen.

I put a breakpoint, and those few calls that hit only ever picked one float from the motion vector, because the vector len of it was indeed just 1.

src/w3d/renderer/motchan.h
```diff
@@ -79,6 +81,65 @@ class MotionChannelClass : public W3DMPO
...
+            *y = 0.0f;
+            *z = 0.0f;
+        } else if (m_data) {
+            frame -= m_firstFrame;
+            *x = m_data[0 + m_vectorLen * frame];
+            *y = m_data[1 + m_vectorLen * frame];
+            *z = m_data[2 + m_vectorLen * frame];
+        }
+    }
+    void Get_Vector_Safe(int frame, float *x, float *y, float *z, float *w)
```
duncanspumpkin, line 126: These functions aren't modifying state so can be const

tomsons26:
The WW design may or may not be horrible but we can't regression test this against every single model it would take way too long

PR #323, opened 2021-10-26
Fix issues in VideoPlayer and ParticleSystemManager where loop could …

xezon:
…continue with invalidated iterator


tomsons26:
Remove comment from videoplayer, this is actually my mistake that return is missing.

tomsons26 (approved):

PR #324, opened 2021-10-26
Fix compiler signed/unsigned warnings

xezon:


src/w3d/renderer/motchan.cpp
```diff
@@ -410,11 +410,11 @@ int TimeCodedBitChannelClass::Get_Bit(int frame)
 
     unsigned int count = 0;
 
-    if (frame >= (m_bits[m_cachedIdx] & 0x7FFFFFFF)) {
+    if (frame >= int(m_bits[m_cachedIdx] & 0x7FFFFFFF)) {
         count = m_cachedIdx + 1;
     }
 
-    while (count < m_numTimeCodes && frame >= (m_bits[count] & 0x7FFFFFFF)) {
+    while (count < m_numTimeCodes && frame >= int(m_bits[count] & 0x7FFFFFFF)) {
```
duncanspumpkin: Why don't we just create an accessor function that returns `m_bits[count] & 0x7FFFFFFF`

duncanspumpkin: Likewise for the flag checked at L427

OmniBlade: I agree, a small inline accessor could tidy up that cast nicely.

OmniBlade (dismissed):
I agree with @duncanspumpkin, creating an accessor would tidy that up nicely and will keep the mask constant defined in a single place as well.

xezon:
It is not clear to me what you mean with accessor and how that would help. I now added template function to hide the 0x7FFFFFFF mask.

duncanspumpkin:
This still doesn't make much sense. You have m_data which contains a frame reference and also contains a flag to indicate something. Just make a `uint32_t Get_Frame_From_Data(uint32_t)` and `bool Get_Flag_From_Data(uint32_t)` function. Get data frame should return a uint32_t. The float casts are just due to the maths and shouldn't be part of the function call.

xezon:
I am afraid I don't understand the point of this exercise. Someone else please feel free to take over this change to get rid of the compiler warnings.

duncanspumpkin:
I have pushed my suggested functions for this.

src/w3d/renderer/motchan.cpp
```diff
@@ -210,15 +228,15 @@ void TimeCodedMotionChannelClass::Get_Vector(float frame, float *setvec)
         int index2 = m_packetSize + index;
         unsigned int val = m_data[index2];
 
-        if (GameMath::Fast_Is_Float_Positive(val)) {
-            float f1 = (m_data[index] & 0x7FFFFFFF);
+        if (Get_Flag_From_Data(val)) {
```
xezon, line 231: Hmm. Is this different from what it did earlier? Before it was converting integer to float to check the bit, now its checking the bit straight from unsigned integer.

src/w3d/renderer/motchan.cpp
```diff
@@ -410,11 +428,11 @@ int TimeCodedBitChannelClass::Get_Bit(int frame)
 
     unsigned int count = 0;
 
-    if (frame >= (m_bits[m_cachedIdx] & 0x7FFFFFFF)) {
+    if (frame >= Get_Frame_From_Data(m_bits[m_cachedIdx])) {
```
xezon: Doesn't this still throw compiler warning because we still compare `int` with `unsigned int` ?

src/w3d/renderer/motchan.cpp
duncanspumpkin: https://github.com/TheAssemblyArmada/Thyme/blob/56218d995d253f3f6b426ff7c0b83c1d481f0390/src/w3d/math/gamemath.h#L346
It was incorrectly using this function. As it's all inlined it's easy to get it mixed up.

src/w3d/renderer/motchan.cpp
duncanspumpkin: Yes that's true. Just need to add another cast here.

duncanspumpkin (approved):

PR #325, opened 2021-10-26
Avoid using invalid file handle in Win32LocalFileSystem::Get_File_Lis…

xezon:
…t_From_Dir

### DIFF
--- a/src/platform/win32localfilesystem.cpp
+++ b/src/platform/win32localfilesystem.cpp
@@ -98,10 +98,11 @@ void Win32LocalFileSystem::Get_File_List_From_Dir(Utf8String const &subdir,
                     filelist.insert(filepath);
                 }
             }
-        } while (FindNextFileW(hndl, &data));
-    }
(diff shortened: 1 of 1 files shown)


tomsons26:
same issue a few lines lower, mark these as bugfixes as this is how the game had it

src/platform/win32localfilesystem.cpp
```diff
@@ -126,10 +127,11 @@ void Win32LocalFileSystem::Get_File_List_From_Dir(Utf8String const &subdir,
...
                     }
                 }
-            } while (FindNextFileW(hndl, &data) != 0);
-        }
+            } while (FindNextFileW(hndl, &data) != FALSE);
 
-        FindClose(hndl);
+            // #BUGFIX Moved into scope to avoid close call on invalid handle.
+            FindClose(hndl);
+        }
```
duncanspumpkin, line 134: We should create an issue for the other side of the ifdef. It can all be done use std::filesystem cross platform and utf8 safe.

OmniBlade (approved):

PR #326, opened 2021-10-26
Fix issue where element past array end would be accessed

xezon:


tomsons26:
remove bugfix comment, it should in fact be this, i just missed it in my review :(

tomsons26 (approved):

PR #327, opened 2021-10-26
Fix issue which allows to index out of bounds in Xfer::xferKindOf

xezon:


tomsons26:
Remove bugfix comment, this is our mistake
This is actually a function that can be found in ZH WB at 006FC5E0
```
char *__cdecl KindOfTypeAsString(int a1)
{
    char *v2; 

    if ( a1 >= 0 && a1 < 116 )
    {
        v2 = BitFlags<116ul>::s_bitNameList[a1];
    }
    else
    {
        v2 = 0;
    }
    return v2;
}
```
, no idea what the real name is at least currently, i suspect its part of BitFlags class

tomsons26 (approved):

PR #328, opened 2021-10-26
Delete copy constructor from classes with memory allocations to avoid…

xezon:
… misuse

### DIFF
--- a/deps/baseconfig/src/macros.h
+++ b/deps/baseconfig/src/macros.h
@@ -176,4 +176,13 @@
 #define SIZE_OF(typ, id) sizeof(((typ *)0)->id)
 #endif // !SIZE_OF
 
+/**
+ * Deletes copy constructor and assignment operator in given class body.
+ */
+#ifndef NOT_COPYABLE
+#define NOT_COPYABLE(TypeName) \
+    TypeName(const TypeName &) = delete; \
+    TypeName &operator=(const TypeName &) = delete;
+#endif // !NOT_COPYABLE
+
 #endif // BASE_MACROS_H
--- a/src/game/common/utility/notcopyable.h
+++ b/src/game/common/utility/notcopyable.h
@@ -1,9 +1,9 @@
 /**
  * @file
  *
- * @author OmniBlade
+ * @author xezon
  *
- * @brief Wrapper to mask STLPort ABI differences for config loader and crashpad init code.
+ * @brief Utility class that can be privately inherited from to avoid copies.
  *
  * @copyright Thyme is free software: you can redistribute it and/or
  *            modify it under the terms of the GNU General Public License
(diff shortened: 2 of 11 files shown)


tomsons26 (changes requested):
Add the functions to said classes, introducing the class via inheritance has a potential of offsetting the layout because a no member class still occupies 4 bytes
alternatively make a macro that adds the two functions

duncanspumpkin:
I don't think that is true http://eel.is/c++draft/intro.object#8 as per the standard you can have 0 size classes. You just need to not have virtual methods.

xezon:
Hmm yeah it is 0 size. We can make it a macro if you have very strong opinion about this. Writing it by hand each time is kinda cumbersome :-)

xezon:
We can add a `static_assert(sizeof(NotCopyable) == 0)`. That should do.

duncanspumpkin:
I've been thinking about this for a bit and I think going for the macro that implements the functions would be best. Although I'm sure we would never encounter any issues with introducing it via inheritance we are entering some definite UB territory as some of these classes will get created by the original executable and would never call the constructor of the `NotCopyable`. Again I really doubt it would cause issues but since we have an alternative that we know would be fine we may as well use that for now.

```cpp
#define DISALLOW_COPY_AND_ASSIGN(TypeName) \
  TypeName(const TypeName&) = delete;   \
  void operator=(const TypeName&) = delete;
```

xezon:
Could you show an example where this would be a problem? If it is broken, it would be good to know for sure.

I punched in a sample code in godbolt and was not able to spot a difference in op codes, other than the vtable label, which will have the same address though.

https://godbolt.org/z/zP3r4KxYr

duncanspumpkin:
The problem is only hypothetical. But imagine this situation: non copyable class gets constructed in vanilla code. Reference to that class is passed to thyme code. Technically thyme now has a class that hasn't been constructed fully. It's the same reason reinterpret cast is pretty much UB. I can't see any reason why it would cause issues just that i think it's UB. I've no idea if the above is correct just my understanding of the language.

duncanspumpkin:
https://quuxplusone.github.io/blog/2021/05/07/std-iterator-as-a-base-class/ this is sort of what i mean. We need to minimise abi differences for the hooks to work. But again i really doubt it would cause issues.
Looks like there would be issues if a class has two empty base classes

xezon:
> In C++, base-class subobjects are distinct objects in their own right, and two objects of the same type cannot occupy the same address; so the EBO applies to only one of them. (Jonathan Wakely has called this the “empty-base exclusion principle.”) Our reversed reverse-iterator object ends up occupying 16 bytes, not just 8!

If I understood correctly he has an iterator which has a size of 8 alone. And his reverse reverse iterator has 16, because it contains the forward and reverse iterator - 2 distinct objects. So it would make sense it is 16 bytes now?

I don't understand why he says empty base classes. Iterator is state full, so it should contain something. 8 bytes sounds reasonable for x64.

xezon:
Macro approach has been added.

PR #331, opened 2021-10-27
Use std::sort

duncanspumpkin:
I don't know enough about sorting algos to know exactly what vanilla was but std::sort is I think fine for this.

### DIFF
--- a/src/w3d/renderer/sortingrenderer.cpp
+++ b/src/w3d/renderer/sortingrenderer.cpp
@@ -19,6 +19,7 @@
 #include "dx8wrapper.h"
 #include "sphere.h"
 #include "w3d.h"
+#include <algorithm>
 #ifdef BUILD_WITH_D3D8
 #include <d3dx8.h>
 #endif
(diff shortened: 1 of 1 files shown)


src/w3d/renderer/sortingrenderer.cpp
```diff
@@ -127,17 +128,55 @@ void Release_Refs(SortingNodeStruct *state)
...
-void Sort(TempIndexStruct *array1, TempIndexStruct *array2)
-{
-#ifdef GAME_DLL
-    Call_Function<void, TempIndexStruct *, TempIndexStruct *>(PICK_ADDRESS(0x0080D3C0, 0x0056A9A0), array1, array2);
-#endif
-    // todo
+    std::sort(begin, end);
+    // Vanilla used some sort of specialised sort see below
+
+    //auto numElements = end - begin;
```
xezon: Is not possible to make the code below work? The -1 indexing looks odd. Maybe remove the commented code if not used and just add a comment that original sort algorithm was replaced.

xezon: And I recommend to add the FEATURE or #FEATURE tag here to document that this deviates from the original, in case we need to look at it later again. (Keywords can be searched in whole Project)

duncanspumpkin: I never bothered finishing it. I would have labelled my variables better if I was continuing it. I would keep this code as one commit and remove it in the next commit then there is history of it if anyone cares.

duncanspumpkin:
Okay i went through and implemented it exactly as per vanilla as well now. Then commented it out.

OmniBlade (approved):

PR #332, opened 2021-10-27
Implements SegmentedLineClass

tomsons26:


src/w3d/renderer/segline.cpp
```diff
@@ -0,0 +1,461 @@
...
+{
+    m_lineRenderer.Set_Merge_Abort_Factor(factor);
+}
+
+void SegmentedLineClass::Set_Subdivision_Levels(unsigned int levels)
+{
+    if (levels >= 7) {
+        levels = 7;
+    }
+    m_maxSubdivisionLevels = levels;
```
duncanspumpkin: Could do with making a constant for max level. Also i like the style of `std::min(levels, 7);` i feel it makes bounding calcs easier to read.

src/w3d/renderer/segline.cpp
```diff
@@ -0,0 +1,461 @@
...
+    } else {
+        PredictiveLODOptimizerClass::Add_Cost(Get_Cost());
+    }
+}
+
+void SegmentedLineClass::Increment_LOD()
+{
+    unsigned int level = m_lineRenderer.Get_Current_Subdivision_Level();
+
+    if (level + 1 < m_maxSubdivisionLevels) {
```
duncanspumpkin: Again a `level = std::min(level + 1, m_maxSubdivisionLevels);`

PR #336, opened 2021-10-28
Move Parse_Science_Vector from INI to ScienceStore

duncanspumpkin:
This has been reworked and is now a move from INI to ScienceStore.

### DIFF
--- a/src/game/common/rts/science.cpp
+++ b/src/game/common/rts/science.cpp
@@ -21,7 +21,7 @@ ScienceStore *g_theScienceStore = nullptr;
 #endif
 
 FieldParse ScienceInfo::s_scienceFieldParseTable[] = {
-    { "PrerequisiteSciences", &INI::Parse_Science_Vector, nullptr, offsetof(ScienceInfo, m_prerequisites) },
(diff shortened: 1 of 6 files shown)


duncanspumpkin:
Suggest removing INI::Scan_Science as well since I suspect almost all users will be from the science store.

jonwil:
This already exists as INI::Parse_Science_Vector so this PR is un-necessary.

duncanspumpkin:
Suggest we remove the INI::Parse_Science_Vector and replace it with one in Science.cpp to make this not happen again.

src/game/common/rts/science.cpp
```diff
@@ -68,3 +68,19 @@ ScienceType ScienceStore::Lookup_Science(const char *name)
...
+    auto &science_vec = *static_cast<std::vector<ScienceType> *>(store);
+    science_vec.clear();
+
+    for (Utf8String name = ini->Get_Next_Token_Or_Null(); name.Is_Not_Empty(); name = ini->Get_Next_Token_Or_Null()) {
+        if (name.Is_None()) {
+            science_vec.clear();
+            break;
+        }
+
+        science_vec.push_back(g_theScienceStore->Lookup_Science(name.Str()));
```
tomsons26: hm this is insert in PPC...

tomsons26 (approved):

PR #337, opened 2021-10-28
Finish WW3DAssetManager

jonwil:


src/w3d/renderer/htreemgr.cpp
```diff
@@ -94,3 +95,33 @@ int HTreeManagerClass::Get_Tree_ID(const char *name)
...
+    int count = 0;
+
+    for (auto i = 0; i < m_numTrees; ++i) {
+        auto &tree = m_treePtr[i];
+
+        if (list.Is_Excluded(tree)) {
+            count++;
+        } else {
+            if (m_treePtr[i]) {
+                delete m_treePtr[i];
```
duncanspumpkin: delete already checks for nullptr afaik

src/w3d/renderer/assetmgr.h
```diff
@@ -102,6 +107,22 @@ class W3DAssetManager
...
+        PROTOTYPES_VECTOR_SIZE = 256,
+        PROTOTYPES_GROWTH_RATE = 32,
+    };
+
+    enum
+    {
+        PROTOTYPE_HASH_TABLE_SIZE = 4096,
+        PROTOTYPE_HASH_BITS = 12,
+        PROTOTYPE_HASH_MASK = 0x00000FFF
+    };
```
duncanspumpkin, line 124: None of these are enums suggest using constexprs. Enum implies that they enumerate which these don't they are all constants which is what constexpr implies.

duncanspumpkin: Also only the hash table size is used the others have no use.

src/w3d/renderer/htreemgr.cpp
xezon: By default it does not. And it should not.

duncanspumpkin: https://isocpp.org/wiki/faq/freestore-mgmt i went and looked it up. Yeah checking for nullptr is not required and some see it as a code smell.

xezon: Interesting. Never knew this.

src/w3d/renderer/assetmgr.h
jonwil: The code is full of places that use enums to hold things like the size of arrays and etc so I see no reason why this one file should be any different when we know the original used enums.

And we don't know for sure that nothing uses the other values (we may have not found it yet but it could be there somewhere)

tomsons26 (approved):

PR #338, opened 2021-10-29
ObjectCreationListStore

duncanspumpkin:
Store has been implemented.

### DIFF
--- a/src/game/logic/object/objectcreationlist.cpp
+++ b/src/game/logic/object/objectcreationlist.cpp
@@ -1,0 +1,99 @@
+/**
+ * @file ObjectCreationList
+ *
+ * @author Duncans_pumpkin
+ *
+ * @brief ObjectCreationNugget, ObjectCreationList, ObjectCreationListStore
+ *
+ * @copyright Thyme is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+
+#include "objectcreationlist.h"
+#include "object.h"
+
+#ifndef GAME_DLL
+ObjectCreationListStore *g_theObjectCreationListStore = nullptr;
+#endif
+
+// zh: 0x004C1310 wb: 0x0075BB30
+ObjectCreationListStore::~ObjectCreationListStore()
+{
+    for (auto *nugget : m_nuggets) {
+        nugget->Delete_Instance();
+    }
+    Clear();
+}
+
+// zh: 0x004C1500 wb: 0x00756CFB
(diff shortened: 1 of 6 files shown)


src/game/logic/object/objectcreationlist.h
```diff
@@ -0,0 +1,73 @@
...
+class Object;
+class Coord3D;
+
+class ObjectCreationNugget : public MemoryPoolObject
+{
+    IMPLEMENT_ABSTRACT_POOL(ObjectCreationNugget);
+
+public:
+    virtual ~ObjectCreationNugget() {}
+    virtual void Create(Object *primaryObj, Coord3D *primary, Coord3D *secondary, float unk1, uint32_t unk2) = 0;
```
tomsons26: so this is where the fun starts with MSVC, this needs to be verified that these 3 functions match in order to the compiled binary. This is cause MSVC resorts same name virtuals
IIRC, GCC does leave the original order so ZH mac should show it.

duncanspumpkin: I dont have the ZH mac binary to be able to check this against.

duncanspumpkin: Yeah in ZH mac its reordered the opposite way around so my layout is correct.

src/game/logic/object/objectcreationlist.cpp
```diff
@@ -0,0 +1,61 @@
...
+ObjectCreationListStore *g_theObjectCreationListStore = nullptr;
+#endif
+
+// zh: 0x004C1310 wb: 0x0075BB30
+ObjectCreationListStore::~ObjectCreationListStore()
+{
+    for (auto *nugget : m_nuggets) {
+        nugget->Delete_Instance();
+    }
+    m_nuggets.clear();
```
tomsons26: Clear call?

xezon: Makes no difference whether ot not is typed out. It will clear and destruct anyway.

tomsons26 (changes requested):
creates need to be implemented, they are simple functions that don't need anything that isn't in already

src/game/logic/object/objectcreationlist.cpp
duncanspumpkin: I think @tomsons26 is saying this should call `Clear();` but the oclstore doesn't have a Clear function only the ocl has a Clear function. But yes it is pretty pointless as the vector is destroyed anyway after this so it ends up being duplicated anyway. So not sure what you want: 

1. add a Clear function and call it instead of clear?
2.  Remove the clear as it's pointless?

tomsons26: I prefer the stl accesses to be buried behind a function if possible, and judging from the functions we know of they did too

tomsons26 (approved):
If its been tested and no further changes are to be made this can be merged

PR #349, opened 2021-10-31
Fix wrong assert in ScriptList::Parse_Scripts_Chunk and improve readability

### LINKED ISSUE #339: ScriptList::Parse_Scripts_Chunk asserts/crashes when loading test map
Call stack:
```
>	thyme.dll!ScriptList::Parse_Scripts_Chunk(DataChunkInput & input, DataChunkInfo * info, void * data) Line 282	C++
 	thyme.dll!DataChunkInput::Parse(void * user_data) Line 119	C++
 	thyme.dll!SidesList::Parse_Sides_Chunk(DataChunkInput & input, DataChunkInfo * info, void * data) Line 374	C++
 	thyme.dll!DataChunkInput::Parse(void * user_data) Line 119	C++
 	[External Code]	
 	game.dat![Frames below may be incorrect and/or missing, no symbols loaded for game.dat]	Unknown
 	thyme.dll!Game_Main(int argc, char * * argv) Line 36	C++
 	thyme.dll!main(int argc, char * * argv) Line 382	C++
 	thyme.dll!Main_Func(HINSTANCE__ * hInstance, HINSTANCE__ * hPrevInstance, char * lpCmdLine, int nCmdShow) Line 406	C++
 	[External Code]	
```

Assert hit:
```c++
captainslog_dbgassert(read_info.num_lists < 16, "Read too many, overrun buffer.");
```

Copy map file into user data maps folder. Then load map from Network room.

[!TestAllFactionsBuildingsUnitsPromo.zip](https://github.com/TheAssemblyArmada/Thyme/files/7444897/TestAllFactionsBuildingsUnitsPromo.zip)

xezon:
Fixes #339

### DIFF
--- a/src/game/logic/scriptengine/scriptlist.cpp
+++ b/src/game/logic/scriptengine/scriptlist.cpp
@@ -18,7 +18,7 @@
 #include "xfer.h"
 #include <captainslog.h>
 
-ScriptList *ScriptList::s_readLists[16];
+ScriptList *ScriptList::s_readLists[MAX_LIST_COUNT];
 int ScriptList::s_numInReadList = 0;
 ScriptGroup *ScriptList::s_emptyGroup = nullptr;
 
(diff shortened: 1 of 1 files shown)


src/game/logic/scriptengine/scriptlist.cpp
```diff
@@ -243,19 +243,24 @@ int ScriptList::Get_Read_Scripts(ScriptList **scripts)
  */
 bool ScriptList::Parse_Script_List_Chunk(DataChunkInput &input, DataChunkInfo *info, void *data)
 {
-    captainslog_dbgassert(
-        static_cast<ScriptListReadInfo *>(data)->num_lists < MAX_LIST_COUNT, "Attempting to parse too many script lists.");
+    // The function body has been marginally rearranged to improve readability in regards to the list index and list count.
```
duncanspumpkin: I just really think these comments are useless. Adding temporary variables doesn't affect anything. And now this code is very confusing. The list index isn't an index until you go to allocate. L257 should be where the index temp variable is created prior to the increment. `const auto list_index = ...->num_lists++;`

xezon: What is confusing?

duncanspumpkin: The confusing aspect is comparing an index to the max count. That is confusing. You should only compare totals with the max count. What happens in the future if we use a different indexing strategy like some sort of map. Then there is the whole aspect that you shouldn't have index's of invalid entries. It's just asking for trouble to make a potentially invalid index when you could just move the index creation down past the if and we now know that it is valid.

xezon: Hmm. It's odd that this would be confusing for you. It's like comparing an index to the size of a vector, array. It's most plausible to check if our desired insertion index reached the max or not.
```c++
for (size_t index = 0; index < MyVector.size(); ++index)
   ...
```

That is why you see in some functions also check like this:
```c++
if (index >= MyVector.size())
    return;
```

xezon: I can improve the comments to make the function purpose easier to understand.

xezon: New comments added.

tomsons26 (approved):
Checks out

PR #350, opened 2021-11-01
Dazzle

jonwil:


src/w3d/renderer/dazzle.h
```diff
@@ -14,21 +14,286 @@
...
+private:
+    StringClass m_name;
+    int m_dazzleType;
 };
 
 class DazzleLoaderClass : public PrototypeLoaderClass
 {
 public:
+    DazzleLoaderClass() {}
+    ~DazzleLoaderClass() {}
```
duncanspumpkin, line 296: Odd that prototype loader doesn't have a virtual destructor. This destructor will therefore likely never be called.

xezon: Virtualness of functions from parent class should propagate to derived class, even when not specifying virtual keyword here.

duncanspumpkin: Yes and since the parent is not virtual this is not virtual. Just pointing out that it was odd. I've had a quick look and it's not UB though as loaders are static instances which are never deleted from their base pointer.

xezon: Not calling a destructor is not Undefined Behavior, because it is very well defined what happens if not called. Whether or not that is bad depends on the context. If the Destructor has nothing to do, then there is logically no difference between being called or not called. The Compiler likely even makes it a no-op with Release.

duncanspumpkin: It is UB to delete a polymorphic type from its base pointer without a virtual destructor. Luckily unlike most items in Thyme we don't delete from a base pointer in this instance as all the classes are static instances and destructed from their derived type. Just because you think there is nothing to do doesn't change something being UB. 

```
std::vector<PrototypeLoader*> loaders;
for (auto* loader : loaders) delete loader;
```
This would be UB but we don't do that so it's fine.

xezon: I guess that depends on our definition of Undefined Behavior. C++ is accurate about events in case of non virtual destructors, meaning the results are reproducible. Undefined Behavior for me means something is not defined and can have different outcomes, so for example an uninitialized integer could be 0 or 1354623. You would not know until you run the program.

PR #359, opened 2021-11-04
ScienceStore

duncanspumpkin:
Implemented all possible ScienceStore functions. Was not able to implement the Player functions at this time.

### DIFF
--- a/src/game/common/rts/science.cpp
+++ b/src/game/common/rts/science.cpp
@@ -99,8 +99,8 @@ Utf8String ScienceStore::Get_Internal_Name_From_Science(ScienceType science) con
     return g_theNameKeyGenerator->Key_To_Name(static_cast<NameKeyType>(science));
 }
 
-// wb 0x0072731A
-std::vector<Utf8String> ScienceStore::Get_All_Science() const
+// wb: 0x0072731A
+std::vector<Utf8String> ScienceStore::Get_All_Sciences() const
 {
     std::vector<Utf8String> all_science;
     for (const auto *science : m_infoVec) {
@@ -125,14 +125,15 @@ bool ScienceStore::Is_Science_Grantable(ScienceType science) const
 }
 
 // zh: 0x00488CE0 wb: 0x007277AF
-void ScienceStore::Get_Name_And_Description(ScienceType science, Utf16String &name, Utf16String &description)
+bool ScienceStore::Get_Name_And_Description(ScienceType science, Utf16String &name, Utf16String &description)
 {
     const auto *info = Get_Science_Info(science);
     if (info == nullptr) {
-        return;
+        return false;
     }
     name = info->Get_Name();
     description = info->Get_Description();
+    return true;
 }
 
 // zh: 0x00488D50 wb: 0x007277F7
@@ -178,20 +179,20 @@ void ScienceStore::Parse_Science_Definition(INI *ini)
     }
 
     if (ini->Get_Load_Type() == INILoadType::INI_LOAD_CREATE_OVERRIDES) {
-        auto *override_info = NEW_POOL_OBJ(ScienceInfo);
+        auto *new_info = new ScienceInfo;
         if (found_info == nullptr) {
-            override_info->Set_Is_Allocated();
+            new_info->Set_Is_Allocated();
             g_theScienceStore->m_infoVec.push_back(found_info);
         } else {
             found_info = static_cast<ScienceInfo *>(found_info->Friend_Get_Final_Override());
-            *override_info = *found_info;
-            found_info->Set_Next(override_info);
-            override_info->Set_Is_Allocated();
+            *new_info = *found_info;
+            found_info->Set_Next(new_info);
+            new_info->Set_Is_Allocated();
         }
-        found_info = override_info;
+        found_info = new_info;
     } else {
         captainslog_dbgassert(found_info == nullptr, "Duplicate Science %s!", name);
-        found_info = NEW_POOL_OBJ(ScienceInfo);
+        found_info = new ScienceInfo;
         g_theScienceStore->m_infoVec.push_back(found_info);
     }
 
--- a/src/game/common/rts/science.cpp
+++ b/src/game/common/rts/science.cpp
@@ -33,10 +33,10 @@ const FieldParse ScienceInfo::s_scienceFieldParseTable[] = {
 };
 
 ScienceInfo::ScienceInfo() :
-    m_nameKey(ScienceType::SCIENCE_INVALID),
+    m_scienceType(ScienceType::SCIENCE_INVALID),
     m_displayName(),
     m_description(),
-    m_rootScience(),
+    m_rootSciences(),
     m_prerequisites(),
     m_purchaseCost(0),
     m_isGrantable(true)
@@ -47,9 +47,9 @@ ScienceInfo::ScienceInfo() :
 void ScienceInfo::Add_Root_Sciences(std::vector<ScienceType> &rootScience) const
 {
     if (m_prerequisites.empty()) {
-        auto res = std::find(rootScience.begin(), rootScience.end(), m_nameKey);
+        auto res = std::find(rootScience.begin(), rootScience.end(), m_scienceType);
         if (res == rootScience.end()) {
-            rootScience.push_back(m_nameKey);
+            rootScience.push_back(m_scienceType);
         }
     } else {
         for (auto science : m_prerequisites) {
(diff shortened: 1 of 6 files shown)


src/game/common/rts/science.cpp
```diff
@@ -68,3 +68,42 @@ ScienceType ScienceStore::Lookup_Science(const char *name)
     captainslog_relassert(false, 0xDEAD0006, "Failed to find matching ScienceType in Science Store.");
     return SCIENCE_INVALID;
 }
+
+void ScienceStore::Parse_Science_Definition(INI *ini)
+{
+    const auto *name = ini->Get_Next_Token();
+    const auto key = g_theNameKeyGenerator->Name_To_Key(name);
```
xezon: `const auto &key` ?

duncanspumpkin: It's a value type. Faster to pass by value.

xezon: Ok. It's not obvious in Review what this is due `auto`.

duncanspumpkin: And it would be just as non obvious from `const NameKey key`

duncanspumpkin: Also to top it off NameKey would have been the wrong type. So just goes to show typing out the type when not required can and will cause mistakes. The correct type is ScienceType which is a sort of alias of NameKey.

src/game/common/rts/science.cpp
```diff
@@ -55,16 +74,139 @@ void ScienceStore::Reset()
...
+        const auto key = static_cast<const ScienceInfo *>(science->Get_Final_Override())->Get_Name_Key();
+        all_science.push_back(g_theNameKeyGenerator->Key_To_Name(static_cast<NameKeyType>(key)));
+    }
+    return all_science;
+}
+
+// zh: 0x00488C40 wb: 0x00727753
+int32_t ScienceStore::Get_Science_Purchase_Cost(ScienceType science) const
+{
+    auto *info = Get_Science_Info(science);
```
xezon: const auto?

src/game/common/rts/science.cpp
```diff
@@ -55,16 +74,139 @@ void ScienceStore::Reset()
...
+        PICK_ADDRESS(0x00488DC0, 0x00727867), this, player, science);
+#else
+    return false;
+#endif
+}
+
+// zh: 0x00488950 wb: 0x0072752B
+void ScienceStore::Parse_Science_Definition(INI *ini)
+{
+    const auto *name = ini->Get_Next_Token();
```
xezon, line 166: What is `Get_Next_Token`? Does it return a `const char*` or `Utf8String`? Looks suspicious with the auto.

duncanspumpkin: Doesn't really matter what it returns and now it's all ready for if the type is modified.

xezon: It looks like bug if `ini->Get_Next_Token()` were to return Utf8String by value. From review I can't see what it returns, so I am raising this here.

duncanspumpkin: The beauty of auto* is that it would be a compile error if it returned it by value and this would have been subtly broken if i had used char* and it turned out the function returned a utf8string by value. See https://godbolt.org/z/v88Tdc33f where I have a mock class "test" that is similar to utf8string with the operator that converts to char*.

xezon: Interesting. I woud have expected the implicit cast to take effect there as well. On the other hand, it probably makes most sense the way it is, because there could be multiple cast operator overloads, and there would be no way auto could decide on that either.

src/game/common/rts/science.cpp
```diff
@@ -55,16 +74,139 @@ void ScienceStore::Reset()
...
+
+// zh: 0x00488C90 wb: 0x00727781
+bool ScienceStore::Is_Science_Grantable(ScienceType science) const
+{
+    const auto *info = Get_Science_Info(science);
+    return info != nullptr ? info->Is_Grantable() : false;
+}
+
+// zh: 0x00488CE0 wb: 0x007277AF
+void ScienceStore::Get_Name_And_Description(ScienceType science, Utf16String &name, Utf16String &description)
```
xezon: Minor: Maybe call this `ScienceType type` in all these functions instead? Just `science` reads ambiguous here. 2 lines below the `ScienceInfo` is also just called `info`. So `type` seems appropriate. Or `science_type`.

src/game/common/rts/science.cpp
```diff
@@ -55,16 +74,139 @@ void ScienceStore::Reset()
...
+            g_theScienceStore->m_infoVec.push_back(found_info);
+        } else {
+            found_info = static_cast<ScienceInfo *>(found_info->Friend_Get_Final_Override());
+            *override_info = *found_info;
+            found_info->Set_Next(override_info);
+            override_info->Set_Is_Allocated();
         }
+        found_info = override_info;
+    } else {
+        captainslog_dbgassert(found_info == nullptr, "Duplicate Science %s!\n", name);
```
xezon: `\n` can be omitted.

src/game/common/rts/science.h
```diff
@@ -56,14 +71,23 @@ class ScienceInfo : public Overridable
...
+
+    static FieldParse *Get_Field_Parse() { return s_scienceFieldParseTable; }
 
 private:
-    NameKeyType m_nameKey;
+    ScienceType m_nameKey;
     Utf16String m_displayName;
     Utf16String m_description;
-    std::vector<ScienceType> m_unkVec1;
+    std::vector<ScienceType> m_rootScience;
```
xezon: Minor: `m_rootSciences` plural because there can be multiple ones?

src/game/common/rts/science.h
```diff
@@ -56,14 +71,23 @@ class ScienceInfo : public Overridable
...
+    const Utf16String &Get_Name() const { return m_displayName; }
+    const Utf16String &Get_Description() const { return m_description; }
+    std::vector<ScienceType> &Get_Root_Science() { return m_rootScience; }
+    void Add_Root_Sciences(std::vector<ScienceType> &rootScience) const;
+
+    static FieldParse *Get_Field_Parse() { return s_scienceFieldParseTable; }
 
 private:
-    NameKeyType m_nameKey;
+    ScienceType m_nameKey;
```
xezon: Minor: Maybe `m_type` ?

src/game/common/rts/science.cpp
duncanspumpkin: It looks like `science` was used in vanilla as the name for a lot of the instances of `ScienceType`. I'm not overly attached to the name though. I don't think `type` would be great there are already lots of `type`s.

duncanspumpkin: I've adjusted to use just science throughout. There were a few instances that still used name_key but its now at least consistent to one thing.

src/game/common/rts/science.cpp
```diff
@@ -55,16 +74,139 @@ void ScienceStore::Reset()
...
+Utf8String ScienceStore::Get_Internal_Name_From_Science(ScienceType science) const
+{
+    if (science == ScienceType::SCIENCE_INVALID) {
+        return Utf8String::s_emptyString;
+    }
+    return g_theNameKeyGenerator->Key_To_Name(static_cast<NameKeyType>(science));
+}
+
+// wb 0x0072731A
+std::vector<Utf8String> ScienceStore::Get_All_Science() const
```
xezon: Is this a function you gave name yourself? Maybe `Get_All_Science_Names` is better.

src/game/common/rts/science.cpp
```diff
@@ -55,16 +74,139 @@ void ScienceStore::Reset()
...
+
+// zh: 0x00488C90 wb: 0x00727781
+bool ScienceStore::Is_Science_Grantable(ScienceType science) const
+{
+    const auto *info = Get_Science_Info(science);
+    return info != nullptr ? info->Is_Grantable() : false;
+}
+
+// zh: 0x00488CE0 wb: 0x007277AF
+void ScienceStore::Get_Name_And_Description(ScienceType science, Utf16String &name, Utf16String &description)
```
xezon: Is this function not returning `bool` ?

src/game/common/rts/science.cpp
```diff
@@ -55,16 +74,139 @@ void ScienceStore::Reset()
...
+    ScienceInfo *found_info = nullptr;
+    for (auto *info : g_theScienceStore->m_infoVec) {
+        if (info->Check_Science_Type(key)) {
+            found_info = info;
+            break;
+        }
+    }
+
+    if (ini->Get_Load_Type() == INILoadType::INI_LOAD_CREATE_OVERRIDES) {
+        auto *override_info = NEW_POOL_OBJ(ScienceInfo);
```
xezon: `auto *new_info` for better understanding.

xezon (approved):
Looks ok logically. Check with @tomsons26 & Co if something needs further attention.

PR #360, opened 2021-11-05
Line up parse table and name base type better

duncanspumpkin:
Just a small bit of formatting changes.


jonwil (approved):
Looks good to me, approved.

PR #361, opened 2021-11-06
Various fixes to wwvegas stuff.

jonwil:
Add missing IMPLEMENT_W3D_POOL macros to Smudge, SmudgeSet and
BoxPrototypeClass.
Add some asserts to CollisionMath.
Fix W3DPrototypeClass to correctly use MemoryPoolObject instead of
W3DMPO.
Add DX8Wrapper::Clamp_Color.
Properly implement W3dUtilityClass.
Add some things to w3d_file.h
Add TextureClass::Get_Level_Description.
Remove MeshGeometryClass constructor hooks and incorrect
IMPLEMENT_W3D_POOL macro.
Add Greatest_Common_Divisor and Least_Commmon_Multiple.
Add NodeCompressedMotionStruct constructor.
Add SimpleFileFactoryClass and defaultFileFactory.
Add DX8Wrapper::Convert_Color_Clamp.
Properly implement all the HashTemplateClass member functions I can
find implementations of.
Properly implement all the HashTemplateIterator member functions.


src/w3d/lib/gcd_lcm.cpp
```diff
@@ -0,0 +1,42 @@
...
+ * @copyright Thyme is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#include "gcd_lcm.h"
+
+unsigned int Greatest_Common_Divisor(unsigned int a, unsigned int b)
```
duncanspumpkin, line 17: What are these being used for? I suggest moving them into the header and making them `constexpr` as they can be computed at compile time for any fixed values of a and b.

src/w3d/lib/gcd_lcm.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+    return d;
+}
+
+unsigned int Least_Common_Multiple(unsigned int a, unsigned int b)
+{
+    if (b) {
+        return b * a / Greatest_Common_Divisor(b, a % b);
+    } else {
+        return 0 / a;
```
duncanspumpkin: This looks wrong.

jonwil: I took another look at this in a few different binaries and re-did it.

src/w3d/lib/gcd_lcm.cpp
jonwil: Its used by the particle emitter code and the values passed in are not constants. I took another look at this in a few binaries and discovered evidence about how it actually looked so I changed it.

jonwil: The way these 2 functions are now is how they look in zero hour worldbuilder so they should be good.

tomsons26 (changes requested):
Def too much and too random for a single pr, need splitting up, at minimum to separate commits, but i'd prefer seperate prs

PR #362, opened 2021-11-06
Fix CaveSystem calls and hook ctor

duncanspumpkin:
Small little fix. Incorrect offsets were being used and also there was a circular calling due to hooking an unimplemented function. 

The SnapShot changes are required as otherwise the this pointer is at the wrong offset.


duncanspumpkin:
Confirmed fixed with a cavesystem map.

OmniBlade (approved):

PR #363, opened 2021-11-07
Implement remaining CaveSystem functions

duncanspumpkin:
Continues from #362. Can either merge that one or ignore and just merge this one.


jonwil (approved):
All checks out based on a comparison with the original. Approved.

PR #364, opened 2021-11-07
Line3DClass

jonwil:


src/w3d/renderer/line3d.cpp
```diff
@@ -0,0 +1,219 @@
...
+    Set_Transform(transform);
+}
+
+Line3DClass::Line3DClass(const Line3DClass &src) :
+    m_length(src.m_length),
+    m_thickness(src.m_thickness),
+    m_shader(src.m_shader),
+    m_color(src.m_color),
+    m_opacity(src.m_opacity)
+{
```
duncanspumpkin, line 150: Shouldn't this also copy the base class?

src/w3d/renderer/line3d.cpp
```diff
@@ -0,0 +1,219 @@
...
+#include "dx8wrapper.h"
+
+unsigned int s_indices[36] = {
+    3, 5, 1, 7, 5, 3, 1, 5, 0, 5, 4, 0, 4, 2, 0, 4, 6, 2, 7, 3, 2, 6, 7, 2, 7, 6, 5, 5, 6, 4, 2, 3, 1, 2, 1, 0
+};
+
+void Line3DClass::Render(RenderInfoClass &rinfo)
+{
+#ifdef BUILD_WITH_D3D8
+    if (Is_Not_Hidden_At_All()) {
```
duncanspumpkin: return early then we avoid all of this indent.

tomsons26 (approved):

src/w3d/renderer/line3d.cpp
```diff
@@ -0,0 +1,219 @@
...
+        m_vertices[i] *= scale;
+    }
+
+    m_length *= scale;
+    m_thickness *= scale;
+
+    Invalidate_Cached_Bounding_Volumes();
+    RenderObjClass *container = Get_Container();
+
+    if (container) {
```
duncanspumpkin, line 99: explicit nullptr check

PR #368, opened 2021-11-08
Proper reverse engineering of HashTemplateClass and HashTemplateIterator

jonwil:


src/w3d/lib/hashtemplate.h
```diff
@@ -41,249 +42,194 @@ template<typename Key, typename Value> class HashTemplateClass
...
 
     ~HashTemplateClass()
     {
-        delete[] m_index;
+        delete[] m_hash;
         delete[] m_table;
     }
 
-    Value *Get(const Key &key) const
+    Value Get(const Key &key) const
```
duncanspumpkin, line 53: Is this really a Get that performs a copy? I would have thought it would be a `const Value&`

src/w3d/lib/hashtemplate.h
```diff
@@ -41,249 +42,194 @@ template<typename Key, typename Value> class HashTemplateClass
...
     {
-        if (m_index != nullptr) {
-            for (unsigned index = m_index[Get_Hash_Val(key)]; index != -1; index = m_table[index].next) {
+        if (m_hash) {
```
duncanspumpkin: m_hash is a pointer so `nullptr` checks should still be in here.

src/w3d/lib/hashtemplate.h
```diff
@@ -41,249 +42,194 @@ template<typename Key, typename Value> class HashTemplateClass
...
         }
 
-        return nullptr;
-    }
-
-    Value Get(const Key &key, Value def_val) const
-    {
-        Value *result = Get(key);
-        return result != nullptr ? *result : def_val;
+        return 0;
```
duncanspumpkin, line 63: What happens if the Value type can't construct from `0`? This just looks wrong. The old code made sense this less so unless Value type must be pointer in which case this should be nullptr.

src/w3d/lib/hashtemplate.h
```diff
@@ -41,249 +42,194 @@ template<typename Key, typename Value> class HashTemplateClass
...
     }
 
-    bool Exists(const Key &key) const { return Get(key) != nullptr; }
+    bool Exists(const Key &key) const { return Get(key) != 0; }
```
duncanspumpkin, line 66: Again looks very wrong Value type might not be comparable to 0

src/w3d/lib/hashtemplate.h
```diff
@@ -41,249 +42,194 @@ template<typename Key, typename Value> class HashTemplateClass
...
+        m_table[index].next = m_hash[hash];
+        m_hash[hash] = index;
     }
 
     void Remove(const Key &key)
     {
-        if (m_index != nullptr) {
-            unsigned hash = Get_Hash_Val(key);
-            int *last_next_entry = &m_index[hash];
+        unsigned int hash = Get_Hash_Val(key);
```
duncanspumpkin, line 81: Suggest we revert to the old code here it made more sense and was better named

src/w3d/lib/hashtemplate.h
```diff
@@ -41,249 +42,194 @@ template<typename Key, typename Value> class HashTemplateClass
...
+        int size = m_hashTable->m_size;
 
-    void operator++() { Increment(); }
-    operator bool() const { return m_index != -1; }
-
-private:
-    void Increment()
-    {
-        m_index = m_table->m_table[m_index].next;
+        for (m_hashIndex++; m_hashIndex < size; m_hashIndex++) {
```
duncanspumpkin, line 218: no need to copy use pre increment ++m_hashIndex

src/w3d/lib/hashtemplate.h
```diff
@@ -41,249 +42,194 @@ template<typename Key, typename Value> class HashTemplateClass
...
-        Increment();
-        m_table->Remove(k);
-    }
-
-    const Key &Get_Key() { return m_table->m_table[m_index].key; }
-    Value &getValue() { return m_table->m_table[m_index].value; }
+        m_handle = m_hashTable->m_table[m_handle].next;
+        int size = m_hashTable->m_size;
 
-    void operator++() { Increment(); }
```
duncanspumpkin, line 266: Readd these are the normal interfaces for iterators.

duncanspumpkin (commented):
I don't see the reason for any of these changes apart from perhaps the name changes and the Re_Hash changes. The old code made sense the new code is now less flexible (MUST only be used with pointers for Value type) removes the common iterator interfaces and harder understand.

src/w3d/lib/hashtemplate.h
jonwil: We have specific information that this absolutely returns a Value and not a Value& or a const Value& (for example in the original game MeshModelClass::Init_For_NPatch_Rendering uses a HashTemplateClass<TriangleSide,SideIndexInfo> and I have explicitly seen a call to get for this HashTemplateClass making a copy of the SideIndexInfo rather than moving around a pointer)

Oh and the Renegade Linux server binary (which has debug symbols same as the Mac Generals binary) has explicit copies of HashTemplateClass<TriangleSide, SideIndexInfo>::Get, HashTemplateClass<StringClass,TDBObjClass *>::Get and HashTemplateClass<uint,MeshClass *>::Get that confirm that my implementation is correct.

src/w3d/lib/hashtemplate.h
jonwil: The value type doesn't need to be a pointer but all things used as a value type are things that can be converted to from int.
The SideIndexInfo type mentioned in my other comment has an explicit SideIndexInfo(int i) constructor to account for this.

src/w3d/lib/hashtemplate.h
jonwil: All the available information (including copies of HashTemplateClass<Vector3,uint>::Exists and HashTemplateClass<TriangleSide,SideIndexInfo>::Exists in the Renegade Linux server binary) confirm my implementation is correct here (SideIndexInfo has an explicit constructor from int so its fine, pointers and integers can obviously be converted from int)

src/w3d/lib/hashtemplate.h
jonwil: Available information (including an explicit copy of HashTemplateClass<StringClass, TextureClass *>::Remove in the Renegade Linux server binary) confirms my implementation.

src/w3d/lib/hashtemplate.h
jonwil: It doesn't really matter, the compiler will generate the same code either way in this case (even in a debug build)

src/w3d/lib/hashtemplate.h
jonwil: Fixed the missing nullptr checks.

src/w3d/lib/hashtemplate.h
```diff
@@ -41,249 +42,194 @@ template<typename Key, typename Value> class HashTemplateClass
...
-
-    Value Get(const Key &key, Value def_val) const
-    {
-        Value *result = Get(key);
-        return result != nullptr ? *result : def_val;
+        return 0;
     }
 
-    bool Exists(const Key &key) const { return Get(key) != nullptr; }
+    bool Exists(const Key &key) const { return Get(key) != 0; }
```
xezon, line 66: Value(0)

OmniBlade (approved):

PR #369, opened 2021-11-09
PartitionManager

duncanspumpkin:
Implemented first stage of the class. Ready for initial merge.

### DIFF
--- a/src/game/logic/system/partitionmanager.h
+++ b/src/game/logic/system/partitionmanager.h
@@ -1,0 +1,262 @@
+/**
+ * @file PartitionManager
+ *
+ * @author Duncans_Pumpkin
+ *
+ * @brief Stores information on current generals rank and purchased "sciences".
+ *
+ * @copyright Thyme is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#include "coord.h"
+#include "gametype.h"
+#include "mempoolobj.h"
+#include "snapshot.h"
+#include "subsysteminterface.h"
+#include <queue>
+
+struct FindPositionOptions;
+struct ShroudStatusStoreRestore;
+class Object;
+class GhostObject;
+class PartitionFilter;
+class SimpleObjectIterator;
+class GeometryInfo;
+class CellAndObjectIntersection;
+
+enum DistanceCalculationType : int32_t
+{
+    FROM_CENTER_2D,
+    FROM_CENTER_3D,
+    FROM_BOUNDINGSPHERE_2D,
+    FROM_BOUNDINGSPHERE_3D,
+};
+
+enum IterOrderType : int32_t
+{
+    ITER_FASTEST,
+    ITER_SORTED_NEAR_TO_FAR,
+    ITER_SORTED_FAR_TO_NEAR,
+    ITER_SORTED_CHEAP_TO_EXPENSIVE,
+    ITER_SORTED_EXPENSIVE_TO_CHEAP,
+};
+
+enum ValueOrThreat : int32_t
+{
+    VOT_THREAT,
+    VOT_VALUE,
+};
+
+enum ShroudedStatusType : int32_t
+{
+    SHROUDED_INVALID,
+    SHROUDED_NONE,
+    SHROUDED_PARTIAL,
+    SHROUDED_ALL,
+    SHROUDED_UNK4,
+};
+
+class PartitionCell : public SnapShot
+{
+public:
+    ~PartitionCell();
+
+    virtual void CRC_Snapshot(Xfer *xfer) override;
+    virtual void Xfer_Snapshot(Xfer *xfer) override;
+    virtual void Load_Post_Process() override{};
+
+private:
+    CellAndObjectIntersection *m_firstCoilInCell;
+    uint32_t unk1[16]; // some sort of 2 16bit structure
+    float unk2;
+    float unk3;
+    uint32_t m_threat[16];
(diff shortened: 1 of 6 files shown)


xezon:
Very nice. There are a few game bugs/exploits that involve this class, so will be nice to have it!

duncanspumpkin:
It's a massive class unfortunately. I think what I'll do is hook up the ctor and implement only the very simplest of functions of it. All the more complex functions will be left as calls. Then the rest of it can be done in follow on PRs. This will allow any other classes that depend on it to be implemented whilst the meat of the functions are slowly worked on.

src/game/logic/system/partitionmanager.h
```diff
@@ -0,0 +1,262 @@
...
+        float *closestDistArg,
+        Coord3D *closestDistVecArg);
+    bool Try_Position(Coord3D const *center, float dist, float angle, FindPositionOptions const *options, Coord3D *result);
+    int Iterate_Cells_Along_Line(
+        Coord3D const &pos, Coord3D const &posOther, int (*proc)(PartitionCell *, void *), void *userData);
+    int Iterate_Cells_Breadth_First(Coord3D const *pos, int (*proc)(PartitionCell *, void *), void *userData);
+    void Remove_All_Dirty_Modules();
+    void Remove_From_Dirty_Modules(PartitionData *data);
+    void Shutdown();
+    void Clear_Pending_Undo_Shroud_Reveal_Queue();
```
jonwil: This one should be named Reset_Pending_Undo_Shroud_Reveal_Queue

jonwil (changes requested):

jonwil (approved):

src/game/logic/system/partitionmanager.h
```diff
@@ -0,0 +1,262 @@
...
+    PartitionData *m_prev;
+    PartitionData *m_nextDirty;
+    PartitionData *m_prevDirty;
+    int32_t m_coiArrayCount;
+    int32_t m_coiInUseCount;
+    CellAndObjectIntersection *m_coiArray;
+    int32_t unk1;
+    int32_t m_dirtyFlag;
+    ShroudedStatusType m_shroudedness[16];
+    ShroudedStatusType m_ghostShroudedness[16];
```
tomsons26: seems this is m_shroudednessPrevious based on friend_setShroudednessPrevious setting this and end of getShroudedStatus setting it to the old value from m_shroudedness

src/game/logic/system/partitionmanager.h
```diff
@@ -0,0 +1,262 @@
...
+    bool m_unk4[16];
```
tomsons26: seems to be a is dirty or changed flag for shroudness?

src/game/logic/system/partitionmanager.h
```diff
@@ -0,0 +1,262 @@
...
+public:
+    ~PartitionCell();
+
+    virtual void CRC_Snapshot(Xfer *xfer) override;
+    virtual void Xfer_Snapshot(Xfer *xfer) override;
+    virtual void Load_Post_Process() override{};
+
+private:
+    CellAndObjectIntersection *m_firstCoilInCell;
+    uint32_t unk1[16]; // some sort of 2 16bit structure
```
tomsons26: m_shroudStatus fetched by getShroudStatusForPlayer
its weird however, its accessed as ushort in addLooker and removeShrouder, union? bitfield?
hm or short m_shroudStatus[16][2]

src/game/logic/system/partitionmanager.h
```diff
@@ -0,0 +1,262 @@
...
+    float unk2;
```
tomsons26: PartitionManager::init sets these to value in range of -1000000.0 1000000.0, mac shows a _HUGE_DIST thats 1000000.0, possibly related

src/game/logic/system/partitionmanager.h
tomsons26: 16 is a player count but not sure if we got a enum for that yet

src/game/logic/system/partitionmanager.h
```diff
@@ -0,0 +1,262 @@
...
+{
+    VOT_THREAT,
+    VOT_VALUE,
+};
+
+enum ShroudedStatusType : int32_t
+{
+    SHROUDED_INVALID,
+    SHROUDED_NONE,
+    SHROUDED_PARTIAL,
```
tomsons26: invalid and none ok, not sure if i can accept partial or all

tomsons26: Yeah looking like 4 is all but very uncertain

tomsons26:
are the args confirmed with ZH Mac? Not all symbols in rts.map match what ZH has as they changed the functions since the demo

src/game/logic/system/partitionmanager.h
duncanspumpkin: Hmm, i was looking at the setters of this and it's always after a ghost check that was my thinking.

src/game/logic/system/partitionmanager.h
duncanspumpkin: what is the convention for an unknown structure? Its definitily two uint16_t's unsure if its a struct or an array

src/game/logic/system/partitionmanager.h
duncanspumpkin: I am unsure at this time so suggest keeping it as unk. It doesn't make too much sense for dirty when there is already a dirty flag.

src/game/logic/system/partitionmanager.h
duncanspumpkin: I've gone for introducing a new struct called `ShrouderLooker` there is clearly a function that is add/remove looker and and/remove shrouder (possibly shroudGeneration)

tomsons26 (approved):

PR #373, opened 2021-11-16
Apply clang tidy to class names

duncanspumpkin:
This only applies the following clang tidy check.
  - key: readability-identifier-naming.ClassCase 
    value: 'CamelCase'

I will revert the other changes clang tidy made without asking.

### DIFF
--- a/src/hooker/setupglobals_zh.cpp
+++ b/src/hooker/setupglobals_zh.cpp
@@ -237,7 +237,7 @@ StringClass &DX8Caps::s_videoCardDetails = Make_Global<StringClass>(PICK_ADDRESS
 
 // multilist.cpp
 #include "multilist.h"
-template<> HOOK_AUTOPOOL(MultiListNodeClass, 256, PICK_ADDRESS(0x00A66288, 0x00E0BB30));
+HOOK_AUTOPOOL(MultiListNodeClass, 256, PICK_ADDRESS(0x00A66288, 0x00E0BB30));
 
 // dx8wrapper.cpp
 #include "dx8wrapper.h"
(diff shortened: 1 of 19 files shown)


src/hooker/setupglobals_zh.cpp
```diff
@@ -237,7 +237,7 @@ StringClass &DX8Caps::s_videoCardDetails = Make_Global<StringClass>(PICK_ADDRESS
 
 // multilist.cpp
 #include "multilist.h"
-HOOK_AUTOPOOL(MultiListNodeClass, 256, PICK_ADDRESS(0x00A66288, 0x00E0BB30));
+template<> HOOK_AUTOPOOL(MultiListNodeClass, 256, PICK_ADDRESS(0x00A66288, 0x00E0BB30));
```
duncanspumpkin: revert

src/w3d/lib/cpudetect.cpp
```diff
@@ -1557,7 +1557,7 @@ void CPUDetectClass::Init_Compact_Log()
 #ifdef PLATFORM_WINDOWS
     TIME_ZONE_INFORMATION time_zone;
     GetTimeZoneInformation(&time_zone);
-    COMPACT_LOG("%d\t", time_zone.Bias);
+    COMPACT_LOG("%ld\t", time_zone.Bias);
```
duncanspumpkin: revert

src/w3d/renderer/render2dsentence.h
```diff
@@ -175,7 +175,7 @@ class Render2DSentenceClass
         const unichar_t *text, int *x = nullptr, int *y = nullptr, bool reuse_surface = false);
 
 private:
-    DynamicVectorClass<SentenceDataStruct> m_sentenceData;
+    DynamicVectorClass<RendererDataStruct> m_sentenceData;
```
duncanspumpkin: revert

src/w3d/renderer/rendobj.h
```diff
@@ -322,7 +322,7 @@ class RenderObjClass : public RefCountClass, public PersistClass, public MultiLi
 #ifdef GAME_DLL
     virtual const PersistFactoryClass &Get_Factory() const;
 #endif
-    virtual bool Save(ChunkSaveClass &csave);
+    virtual bool Save(ChunkLoadClass &csave);
```
duncanspumpkin: revert

OmniBlade:
Could we do something similar to how clang-format is set up so we can run a format from a target and catch style violations on the CI. I'm happy to enforce more of the style guide automatically :D

duncanspumpkin:
Yes that is possible. It will take a little while to get in the state where this could be enforced. This GitHub action could be used https://github.com/platisd/clang-tidy-pr-comments

PR #374, opened 2021-11-16
Some INI related fixes

jonwil:


src/game/common/audio/audiomisc.cpp
```diff
@@ -66,6 +66,7 @@ FieldParse MiscAudio::s_fieldParseTable[] = { { "RadarNotifyUnitUnderAttackSound
     { "AircraftWheelScreech", &INI::Parse_Audio_Event_RTS, nullptr, offsetof(MiscAudio, m_aircraftWheelScreech) },
     { nullptr, nullptr, nullptr, 0 } };
 
+// Was originally INI::parseMiscAudio
```
duncanspumpkin, line 69: Why no definition for this one

jonwil: I don't know, ask EA :)
The original function in the game is named INI::parseMiscAudio.

src/game/common/globaldata.cpp
```diff
@@ -699,13 +699,6 @@ GlobalData::GlobalData()
     m_hardSoloAIHealthBonus = 1.0f;
```
tomsons26, line 699: this made me notice the health bonuses were arrays

tomsons26 (changes requested):

src/game/common/globaldata.cpp
duncanspumpkin: Are they? The INI parse function doesn't look to be using it as an array L319.

tomsons26 (approved):

src/game/common/globaldata.cpp
tomsons26: They are, ill pr the fix later

PR #375, opened 2021-11-19
Refactor TRANS_LIT_TEX_VERTEX into two instances

duncanspumpkin:
_TRANS_LIT_TEX_VERTEX was a problem child in the clang-tidy PR so had a closer look at it. We have two different instances of this in different forms in the codebase. I've created TransformedTexture1Vertex and TransformedTexture2Vertex to replace the two different types. I've also extracted out the D3DFVF values for each struct so its clear why they are being set.

### DIFF
--- a/src/game/client/shader/bwfilter.cpp
+++ b/src/game/client/shader/bwfilter.cpp
@@ -76,14 +76,6 @@ bool ScreenBWFilter::Pre_Render(bool &skip, CustomScenePassModes &mode)
 bool ScreenBWFilter::Post_Render(FilterModes mode, Coord2D &delta, bool &b)
 {
 #ifdef BUILD_WITH_D3D8
-    struct _TRANS_LIT_TEX_VERTEX
-    {
-        D3DXVECTOR4 p;
-        unsigned long color;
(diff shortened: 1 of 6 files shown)


duncanspumpkin:
Suggested to use https://github.com/TheAssemblyArmada/Thyme/blob/69b10921b42c6848fcd118c2fa257ec4f51c962d/src/w3d/renderer/dx8vertexbuffer.h#L45

Sadly this isn't the same this is for Vector4. Could still use the same naming scheme though.

duncanspumpkin:
We have the following vertex formats defined:
`VertexFormatXYZDUV2` within w3dwater.h
`VertexFormatXYZNDUV2` within dx8vertexbuffer.h
`VertexFormatXYZDUV1` within dx8vertexbuffer.h
This PR adds:
`VertexFormatXYZWDUV1` within dx8wrapper.h
`VertexFormatXYZWDUV2` within dx8wrapper.h

tomsons26 (changes requested):
This needs a bit of a debate, it is a bit annoying that these formats are scattered all over the place

PR #376, opened 2021-11-20
WaterRenderObjClass

jonwil:


src/game/client/water/w3dwater.h
```diff
@@ -0,0 +1,225 @@
...
+        unsigned long vertex10_diffuse;
+        unsigned long vertex11_diffuse;
+        unsigned long vertex01_diffuse;
+        unsigned long water_diffuse;
+        unsigned long transparent_water_diffuse;
+        float u_scroll_per_ms;
+        float v_scroll_per_ms;
+    };
+
+    struct SEA_PATCH_VERTEX
```
duncanspumpkin, line 87: Do we really need 6+ identical structs? Just do a: `using SeaPatchVertex = VertexFormatXYZDUV1` or whatever the other struct is called

duncanspumpkin: ahh wait this is a Vector3.

src/game/logic/map/polygontrigger.h
```diff
@@ -16,9 +16,72 @@
...
+
+    ICoord3D *Get_Point(int i) const
+    {
+        if (i < 0) {
+            i = 0;
+        }
+
+        if (i >= m_numPoints) {
+            i = m_numPoints - 1;
+        }
```
xezon, line 44: Can we reword these several lines with a `i = clamp(0, m_numPoints - 1)` ?

PR #386, opened 2021-11-29
Complete definition of Drawable class

jonwil:
Correct implementation of BitFlags
Definition of XferStatus
Implement some Thing and ThingTemplate member functions

### DIFF
--- a/src/game/common/bitflags.h
+++ b/src/game/common/bitflags.h
@@ -17,6 +17,7 @@
 #include "always.h"
 #include "asciistring.h"
 #include "ini.h"
+#include "xfer.h"
 #include <bitset>
 
 // TODO move this somewhere more appropriate?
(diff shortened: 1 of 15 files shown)


src/game/logic/scriptengine/scriptparam.cpp
```diff
@@ -425,7 +425,7 @@ Parameter *Parameter::Read_Parameter(DataChunkInput &input)
     if (param->m_type == OBJECT_STATUS) {
         for (unsigned bit = 0; bit < OBJECT_STATUS_COUNT; ++bit) {
             if (param->m_string.Compare_No_Case(BitFlags<OBJECT_STATUS_COUNT>::s_bitNamesList[bit]) == 0) {
-                param->m_objStatus.Set(bit);
+                param->m_objStatus.Set(bit, 1);
```
tomsons26: something isn't correct here in the first place

src/game/common/bitflags.h
```diff
@@ -216,20 +217,218 @@ enum ModelConditionFlagType
...
+
+    BitFlags(BogusInitType type, int flag1, int flag2, int flag3, int flag4)
+    {
+        m_bits.set(flag1);
+        m_bits.set(flag2);
+        m_bits.set(flag3);
+        m_bits.set(flag4);
+    }
+
+    BitFlags(BogusInitType type, int flag1, int flag2, int flag3, int flag4, int flag5)
```
duncanspumpkin, line 248: we could do this with a fold expression and handle any number of flags and cut down on all of this repeat code.
```cpp
template<typename... T>
BitFlags(T&&... flags){
    (m_bits.set(flags), ...);
}
```

duncanspumpkin:
Why are the bitflags changes part of this PR. They should be reviewed separate .

PR #387, opened 2021-12-01
Complete definition of BitFlags and some related changes.

jonwil:


src/game/common/bitflags.h
```diff
@@ -216,20 +217,218 @@ enum ModelConditionFlagType
...
+    {
+        BitFlags flags;
+        flags.Set(bit, 1);
+        return flags;
+    }
+
+    static BitFlags Clear_Bit(int bit)
+    {
+        BitFlags flags;
+        flags.Set(bit, 1);
```
duncanspumpkin: think this is meant to be 0

duncanspumpkin: hmm but if it was 0 then this function is useless. What is this for? It constructs an empty bitset and sets none of its fields. Very odd..

src/game/common/bitflags.h
```diff
@@ -216,20 +217,218 @@ enum ModelConditionFlagType
...
 
     void Clear() { m_bits.reset(); }
-    void Set(unsigned bit) { m_bits.set(bit); }
-    bool Get(unsigned bit) const { return m_bits.test(bit); }
+    void Set(int bit, int value) { m_bits.set(bit, value); }
+    bool Test(int bit) const { return m_bits.test(bit); }
     bool Any() const { return m_bits.any(); }
+    void Flip() { m_bits.flip(); }
+    int Count() { return m_bits.count(); }
+    int Size() { return m_bits.size(); }
```
duncanspumpkin: this can be `const` or even `constexpr` as its just returning `bits`.

src/game/common/bitflags.h
```diff
@@ -216,20 +217,218 @@ enum ModelConditionFlagType
...
     bool operator!=(BitFlags &that) const { return m_bits != that.m_bits; }
 
     void Clear() { m_bits.reset(); }
-    void Set(unsigned bit) { m_bits.set(bit); }
-    bool Get(unsigned bit) const { return m_bits.test(bit); }
+    void Set(int bit, int value) { m_bits.set(bit, value); }
+    bool Test(int bit) const { return m_bits.test(bit); }
     bool Any() const { return m_bits.any(); }
+    void Flip() { m_bits.flip(); }
+    int Count() { return m_bits.count(); }
```
duncanspumpkin: const

src/game/common/bitflags.h
```diff
@@ -216,20 +217,218 @@ enum ModelConditionFlagType
...
+    static BitFlags Clear_Bit(int bit)
+    {
+        BitFlags flags;
+        flags.Set(bit, 1);
+        return flags;
+    }
+
+    static const char *Bit_As_String(int bit)
+    {
+        if (bit < 0 || bit >= bits) {
```
duncanspumpkin, line 421: `bit >= Size()` would be a bit more descriptive.

src/game/common/bitflags.h
```diff
@@ -216,20 +217,218 @@ enum ModelConditionFlagType
...
+        if (Test(bit)) {
+            return s_bitNamesList[bit];
+        }
+
+        return nullptr;
+    }
+
+    void Get_Name_For_Bits(Utf8String *str)
+    {
+        if (str) {
```
duncanspumpkin, line 391: str != nullptr

duncanspumpkin: `str != nullptr`

src/game/common/bitflags.h
```diff
@@ -216,20 +217,218 @@ enum ModelConditionFlagType
...
+            for (int i = 0; i < Size(); i++) {
+                const char *c = Get_Name_For_Bit(i);
```
duncanspumpkin: c tends to imply character but here it is the bit name. perhaps `name` would be a better name.

src/game/common/bitflags.h
```diff
@@ -216,20 +217,218 @@ enum ModelConditionFlagType
...
+
+                if (c) {
```
duncanspumpkin: c != nullptr

src/game/common/bitflags.h
```diff
@@ -216,20 +217,218 @@ enum ModelConditionFlagType
...
+        xfer->xferVersion(&version, 1);
+
+        if (xfer->Get_Mode() == XFER_SAVE) {
+            int count = Count();
+            xfer->xferInt(&count);
+
+            for (int i = 0; i < Size(); i++) {
+                const char *str = Get_Name_For_Bit(i);
+
+                if (str) {
```
duncanspumpkin: str != nullptr

src/game/common/bitflags.h
```diff
@@ -242,77 +441,88 @@ template<int bits> void BitFlags<bits>::Parse(INI *ini, Utf8String *string)
...
 
-    if (token != nullptr) {
-        while (strcasecmp(token, "NONE") != 0) {
-            // If we have a passed in Utf8String, add the tokens to it as we parse them.
-            if (string != nullptr) {
-                if (string->Is_Not_Empty()) {
-                    *string += " ";
-                }
+    for (const char *token = ini->Get_Next_Token_Or_Null(); token; token = ini->Get_Next_Token_Or_Null()) {
+        if (string) {
```
duncanspumpkin: `!= nullptr`

src/game/common/bitflags.h
```diff
@@ -242,77 +441,88 @@ template<int bits> void BitFlags<bits>::Parse(INI *ini, Utf8String *string)
...
+        }
 
-                *string += token;
+        if (!strcasecmp(token, "NONE")) {
+            if (set || adjust) {
+                captainslog_error("you may not mix normal and +- ops in bitstring lists");
+                throw 0xDEAD0006;
             }
 
-            // If we have a plus or minus as first char, then that decides if bit is set or cleared.
```
duncanspumpkin, line 260: readd this comment

src/game/common/bitflags.h
```diff
@@ -242,77 +441,88 @@ template<int bits> void BitFlags<bits>::Parse(INI *ini, Utf8String *string)
...
+        if (*token == '+') {
+            if (set) {
+                captainslog_error("you may not mix normal and +- ops in bitstring lists");
+                throw 0xDEAD0006;
             }
 
-            token = ini->Get_Next_Token_Or_Null();
+            Set(INI::Scan_IndexList(token + 1, s_bitNamesList), 1);
+            adjust = true;
+        } else if (*token == '+') {
```
duncanspumpkin: copy paste mistake?? shouldn't this be - or the one above it be -

src/game/logic/scriptengine/scriptparam.cpp
```diff
@@ -424,19 +424,21 @@ Parameter *Parameter::Read_Parameter(DataChunkInput &input)
     // Read the object status into our bitflag struct.
     if (param->m_type == OBJECT_STATUS) {
         for (unsigned bit = 0; bit < OBJECT_STATUS_COUNT; ++bit) {
-            if (param->m_string.Compare_No_Case(BitFlags<OBJECT_STATUS_COUNT>::s_bitNamesList[bit]) == 0) {
-                param->m_objStatus.Set(bit);
+            if (param->m_string.Compare_No_Case(BitFlags<OBJECT_STATUS_COUNT>::Get_Bit_Names_List()[bit]) == 0) {
```
duncanspumpkin, line 427: would be safer to call `Bit_As_String`

duncanspumpkin:
Review based just on this code not looked at assembly.

src/game/common/bitflags.h
jonwil: I realized my clone is wrong, fixed.

src/game/common/bitflags.h
jonwil: added const to all functions that don't change m_bits.

src/game/common/bitflags.h
jonwil: Added const to all functions that don't change m_bits.

src/game/common/bitflags.h
jonwil: fixed.

src/game/common/bitflags.h
jonwil: fixed.

src/game/common/bitflags.h
jonwil: fixed.

src/game/common/bitflags.h
jonwil: fixed.

src/game/common/bitflags.h
jonwil: fixed.

src/game/common/bitflags.h
jonwil: fixed.

src/game/common/bitflags.h
jonwil: fixed.

src/game/common/bitflags.h
```diff
@@ -216,20 +217,216 @@ enum ModelConditionFlagType
...
+        m_bits.set(flag5);
+    }
+
     bool operator==(BitFlags &that) const { return m_bits == that.m_bits; }
     bool operator!=(BitFlags &that) const { return m_bits != that.m_bits; }
 
     void Clear() { m_bits.reset(); }
-    void Set(unsigned bit) { m_bits.set(bit); }
-    bool Get(unsigned bit) const { return m_bits.test(bit); }
+    void Set(int bit, int value) { m_bits.set(bit, value); }
```
duncanspumpkin, line 261: are these (set, test,)  actually signed std::bitset uses std::size_t which is unsigned. I would have thought that it would use `uint32_t`.

jonwil: The available information says that yes in fact they take int.

src/game/common/bitflags.h
```diff
@@ -242,77 +439,89 @@ template<int bits> void BitFlags<bits>::Parse(INI *ini, Utf8String *string)
         string->Clear();
     }
 
-    const char *token = ini->Get_Next_Token_Or_Null();
-    bool adjust = false;
     bool set = false;
```
duncanspumpkin, line 442: These names aren't great but i'm really struggling to think of better names. Set represents being in the no operator state and adjust represents being in the operator state. I think having some comment on this would make it more clear as at first glance you would expect Set and Clear to be the two states.

src/game/common/bitflags.h
```diff
@@ -242,77 +439,89 @@ template<int bits> void BitFlags<bits>::Parse(INI *ini, Utf8String *string)
...
+            if (adjust) {
+                captainslog_error("you may not mix normal and +- ops in bitstring lists");
+                throw 0xDEAD0006;
             }
+
+            if (!set) {
+                Clear();
+            }
+
+            Set(INI::Scan_IndexList(token + 1, s_bitNamesList), 1);
```
duncanspumpkin: token should not be incremented by 1. There is no operator so it does not need to be skipped.

src/game/common/system/xfer.cpp
```diff
@@ -307,24 +308,24 @@ void Xfer::xferKindOf(KindOfType *thing)
...
+        case XFER_LOAD: {
             xferAsciiString(&kind);
             kindc = kind.Str();
 
-            for (int i = 0; BitFlags<KINDOF_COUNT>::s_bitNamesList[i] != nullptr; ++i) {
-                if (strcasecmp(kindc, BitFlags<KINDOF_COUNT>::s_bitNamesList[i]) == 0) {
-                    *thing = (KindOfType)i;
-                    break;
-                }
+            int i = BitFlags<116>::Get_Single_Bit_From_Name(kindc);
```
duncanspumpkin: KINDOF_COUNT instead of 116

src/game/common/bitflags.h
jonwil: fixed.

src/game/common/system/xfer.cpp
jonwil: fixed.

duncanspumpkin (approved):

tomsons26 (changes requested):
I need to check this before this is merged

src/game/common/system/xfer.h
```diff
@@ -41,6 +42,28 @@ enum XferType
     XFER_CRC,
 };
 
+enum XferStatus
```
tomsons26, line 45: are these confirmed?

tomsons26 (approved):

PR #392, opened 2021-12-08
Add missing code in TextureClass destructor

### LINKED ISSUE #348: Game loses dx8 device and is unable to reset
100% Repro:
- Start game windowed
- Press CTRL+ALT+DEL
- Click "Lock" option to Lock Windows User
- Re-login Windows User

Occasionally, game loses dx8 device and is unable to reset. This happens in Windowed Mode. This is a Thyme specific issue.

In `W3D::Begin_Render` it will always enter this branch:
```c++
} else if (res == D3DERR_DEVICENOTRESET) {
    DX8Wrapper::Reset_Device(true);
    return W3D_ERROR_GENERIC;
}
```

And in `DX8Wrapper::Reset_Device` it will return false, with `IDirect3DDevice8::Reset` function returning `D3DERR_DEVICELOST`.
```c++
    HRESULT res;
    DX8CALL_HRES(Reset(&s_presentParameters), res);

    if (res) {
        return false;
    }
```

https://docs.microsoft.com/en-us/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3ddevice9-testcooperativelevel
> If the method fails, the return value can be one of the following values: D3DERR_DEVICELOST, D3DERR_DEVICENOTRESET, D3DERR_DRIVERINTERNALERROR

> periodically call IDirect3DDevice9::TestCooperativeLevel until it receives a return value of D3DERR_DEVICENOTRESET. The application may then attempt to reset the device by calling IDirect3DDevice9::Reset

https://docs.microsoft.com/en-us/previous-versions/windows/embedded/ms889709(v=msdn.10)
> Before calling the Reset method for a device, an application should release any explicit render targets, depth stencil surfaces, additional swap chains and D3DPOOL_DEFAULT resources associated with the device.

We can check if there are any resources not being Released.

This function may be worth a look. It has several conditions that prevent a `Release`.
```c++
void TextureBaseClass::Invalidate()
{
    if (!m_normalTextureLoadTask && !m_thumbnailTextureLoadTask && !m_isProcedural) {
#ifdef BUILD_WITH_D3D8
        if (m_d3dTexture != W3D_TYPE_INVALID_TEXTURE) {
            m_d3dTexture->Release();
            m_d3dTexture = nullptr;
        }
#endif
        m_initialized = false;
        m_lastAccess = W3D::Get_Sync_Time();
    }
}
```

xezon:
Fixes #348

### DIFF
--- a/src/w3d/renderer/texture.cpp
+++ b/src/w3d/renderer/texture.cpp
@@ -215,7 +215,6 @@ TextureClass::TextureClass(unsigned width,
 #endif
 
     if (pool == POOL_DEFAULT) {
-        captainslog_assert(m_pool == POOL_DEFAULT);
         m_dirty = true;
         DX8TextureManagerClass::Add(new DX8TextureTrackerClass(width, height, mip_count, this, format, render_target));
     }
@@ -342,9 +341,9 @@ void TextureClass::Init()
 void TextureClass::Apply_New_Surface(w3dbasetexture_t d3d_texture, bool initialized, bool reset)
 {
 #ifdef BUILD_WITH_D3D8
-    w3dbasetexture_t tex = Peek_Platform_Base_Texture();
-    if (tex != W3D_TYPE_INVALID_TEXTURE) {
-        tex->Release();
+    w3dbasetexture_t old = Peek_Platform_Base_Texture();
+    if (old != W3D_TYPE_INVALID_TEXTURE) {
+        old->Release();
     }
 
     m_d3dTexture = d3d_texture;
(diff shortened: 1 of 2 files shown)


src/w3d/renderer/texture.cpp
```diff
@@ -215,7 +215,6 @@ TextureClass::TextureClass(unsigned width,
 #endif
 
     if (pool == POOL_DEFAULT) {
-        captainslog_assert(m_pool == POOL_DEFAULT);
```
xezon, line 218: This assert is obsolete. It's always true with no chance being false ever.

tomsons26: it checks m_pool not pool but i guess

tomsons26 (approved):

PR #393, opened 2021-12-08
Improve minor code in w3d renderer

xezon:


src/w3d/renderer/dx8wrapper.cpp
```diff
@@ -436,7 +436,7 @@ bool DX8Wrapper::Reset_Device(bool reacquire)
     HRESULT res;
     DX8CALL_HRES(Reset(&s_presentParameters), res);
 
-    if (res) {
+    if (S_OK != res) {
```
xezon, line 439: S_OK is HRESULT(0), which is more appropriate to compare with here.

src/w3d/renderer/dx8wrapper.cpp
```diff
@@ -1782,7 +1782,7 @@ void DX8Wrapper::Set_Light_Environment(LightEnvironmentClass *light_env)
                 light.Type = D3DLIGHT_POINT;
                 light.Attenuation0 = 1.0f;
 
-                if (fabs(light_env->Get_Point_Inner_Radius(light_index) - light_env->Get_Point_Outer_Radius(light_index))
+                if (fabsf(light_env->Get_Point_Inner_Radius(light_index) - light_env->Get_Point_Outer_Radius(light_index))
```
xezon, line 1785: `fabs` is for double, but the arguments are float. `fabsf` is for float.

src/w3d/renderer/texturebase.cpp
```diff
@@ -71,7 +71,7 @@ TextureBaseClass::~TextureBaseClass()
 #ifdef BUILD_WITH_D3D8
     if (m_d3dTexture != W3D_TYPE_INVALID_TEXTURE) {
         m_d3dTexture->Release();
-        m_d3dTexture = nullptr;
+        m_d3dTexture = W3D_TYPE_INVALID_TEXTURE;
```
xezon, line 74: We use `W3D_TYPE_INVALID_TEXTURE` at some places. Might as well add more consistency here.

tomsons26 (approved):

PR #395, opened 2021-12-08
Fixes issues in Render2DSentence Allocate_New_Surface

tomsons26:
This was adding 1 again, original only adds it once
Loop was to <= 8

### DIFF
--- a/src/w3d/renderer/render2dsentence.cpp
+++ b/src/w3d/renderer/render2dsentence.cpp
@@ -686,14 +686,15 @@ void Render2DSentenceClass::Allocate_New_Surface(const unichar_t *text, bool reu
         text_width += m_font->Get_Char_Spacing(text[index]);
     }
 
-    int char_height = m_font->Get_Char_Height() + 1;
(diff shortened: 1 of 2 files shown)


src/w3d/renderer/render2dsentence.cpp
```diff
@@ -686,11 +686,11 @@ void Render2DSentenceClass::Allocate_New_Surface(const unichar_t *text, bool reu
         text_width += m_font->Get_Char_Spacing(text[index]);
     }
 
-    int char_height = m_font->Get_Char_Height() + 1;
+    int char_height = m_font->Get_Char_Height();
```
duncanspumpkin: It would make more sense to add the 1 here as per the original rather than recompute it every loop (although yeah im sure that optimiser would do that anyway). Also you can make char_height `const`

xezon: Either is fine, but yeah if it is added in the loop down below, might just add 1 here instead.

duncanspumpkin (approved):

OmniBlade (approved):

PR #403, opened 2021-12-12
Guard files with hooks from being used in standalone by mistake

xezon:


jonwil (approved):
Looks good to me, approved.

PR #405, opened 2021-12-12
Decouple strl template functions from HAVE_STD_CLAMP guard

xezon:


duncanspumpkin:
It was mentioned in the chat that perhaps the better option is to modify stl port that we have in this repo to add std::clamp. std::clamp is located in the algorithm header in the standard. I.e. move it to; https://github.com/TheAssemblyArmada/Thyme/blob/3e59017e30d644d89da1b8bb2f1cffb5ce5aaf58/deps/stlport/stl/_algo.h#L59

xezon:
That is unrelated to this change. The macros are wrong in always.h (I did the mistake originally).

xezon:
Obsolete.

PR #406, opened 2021-12-12
Allow captnlog to log without printing file name if so desired

xezon:
I would like to use captnlog without printing file names in tool for cleaner text presentation.


duncanspumpkin:
What would the use case for this be. Perhaps captainslog is inappropriate for your use case.

xezon:
Yeah I was thinking about making my own log. The lack of custom colors is for example also a problem.

xezon:
I will implement this properly.

xezon:
This Merge is superseeded by #412

PR #409, opened 2021-12-14
Implements W3DShadowGeometry classes.

tomsons26:


src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+    m_numPolygons(0),
+    m_polygons(nullptr)
+{
+}
+
+W3DShadowGeometryMesh::~W3DShadowGeometryMesh()
+{
+    Delete_Neighbors();
+
+    if (m_parentVerts) {
```
xezon: Tests could be omitted, because delete already tests null. But can also keep.

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+}
+
+PolyNeighbor *W3DShadowGeometryMesh::Get_Poly_Neighbor(int index)
+{
+    if (!m_polyNeighbors) {
+        Build_Polygon_Neighbors();
+    }
+    if (index < 0 || index >= m_numPolyNeighbors) {
+        captainslog_assert(0);
+        return nullptr;
```
xezon, line 62: This is a very odd assert, because the test is not compiled out. The original is like that?

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+}
+
+void W3DShadowGeometryMesh::Build_Polygon_Neighbors()
+{
+    Build_Polygon_Normals();
+
+    int num = Get_Num_Polygon();
+
+    if (num == 0) {
+        if (m_numPolyNeighbors) {
```
xezon: Style: `m_numPolyNeighbors != 0`

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+
+                                // TODO what is this
+                                unsigned int p0 = (p0_dist & 0x80000000) ^ (abs(p0_dist) & 2) << 30;
+                                unsigned int p1 = (p1_dist & 0x80000000) ^ (abs(p1_dist) & 2) << 30;
+
+                                if (p0 != p1) {
+
+                                    Vector3 *other_normal = Get_Polygon_Normal(j);
+
+                                    const float epsilon = 1.0f / 100.0f;
```
xezon, line 140: Compiler will compute the value on compile time right? If not, would be better to move out of loop.

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+
+bool W3DShadowGeometryMesh::Allocate_Neighbors(int num_polys)
+{
+    captainslog_assert(m_numPolyNeighbors == 0);
+    captainslog_assert(m_polyNeighbors == nullptr);
+
+    m_polyNeighbors = new PolyNeighbor[num_polys];
+
+    if (!m_polyNeighbors) {
+        captainslog_assert(0);
```
xezon, line 179: Style: `m_polyNeighbors == nullptr`
Again, this assert is odd. It asserts on well defined code path.

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+    }
+
+    m_numPolyNeighbors = num_polys;
+    return true;
+}
+
+void W3DShadowGeometryMesh::Delete_Neighbors()
+{
+
+    if (m_polyNeighbors) {
```
xezon: Style: `m_polyNeighbors != nullptr`

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+        delete[] m_polyNeighbors;
+        m_polyNeighbors = nullptr;
+        m_numPolyNeighbors = 0;
+    }
+
+    captainslog_assert(m_numPolyNeighbors == 0);
+    captainslog_assert(m_polyNeighbors == nullptr);
```
xezon: These asserts are obsolete. By design there is no chance they ever hit because of the code above.

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+}
+
+Vector3 *W3DShadowGeometryMesh::Get_Polygon_Normal(int index) const
+{
+    captainslog_assert(m_polygonNormals);
```
xezon: Style `m_polygonNormals != nullptr`

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+        }
+
+        m_polygonNormals = normals;
+    }
+}
+
+void W3DShadowGeometryMesh::Get_Polygon_Normal(int index, Vector3 *normal) const
+{
+
+    if (m_polygonNormals) {
```
xezon: Style `m_polygonNormals != nullptr`

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+    index_list[2] = m_parentVerts[v->K];
+}
+
+W3DShadowGeometry::W3DShadowGeometry() : m_name{}, m_meshCount(0), m_numTotalsVerts(0) {}
+
+void W3DShadowGeometry::Set_Name(const char *name)
+{
+    memset(m_name, 0, sizeof(m_name));
+
+    strncpy(m_name, name, sizeof(m_name) - 1);
```
xezon: Could use `strlcpy` here, unless the string really needs zero-ed out after last character.

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+}
+
+int W3DShadowGeometry::Init(RenderObjClass *robj)
+{
+    return 1;
+}
+
+int W3DShadowGeometry::Init_From_HLOD(RenderObjClass *robj)
+{
+    HLodClass *hlod = (HLodClass *)robj;
```
xezon: `static_cast` ?

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+    }
+    return m_meshCount != 0;
+}
+
+int W3DShadowGeometry::Init_From_Mesh(RenderObjClass *robj)
+{
+
+    W3DShadowGeometryMesh *geo_mesh = &m_meshList[m_meshCount];
+
+    MeshClass *mesh = (MeshClass *)robj;
```
xezon: `static_cast` ?

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+    m_meshCount = 1;
+    m_numTotalsVerts = num;
+
+    return 1;
+}
+
+W3DShadowGeometryManager::W3DShadowGeometryManager()
+{
+    m_geomPtrTable = new HashTableClass(2048);
+    m_missingGeomTable = new HashTableClass(2048);
```
xezon, line 418: `m_geomPtrTable` and `m_missingGeomTable` are not deleted anywhere?

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+        W3DShadowGeometry *v1 = it.Get_Current_Geom();
+        v1->Release_Ref();
+    }
+
+    m_geomPtrTable->Reset();
+}
+
+W3DShadowGeometry *W3DShadowGeometryManager::Peek_Geom(const char *name)
+{
+    return (W3DShadowGeometry *)m_geomPtrTable->Find(name);
```
xezon: `static_cast` ?

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,503 @@
...
+        int id = robj->Class_ID();
+
+        if (id == RenderObjClass::CLASSID_MESH) {
+            state = geo->Init_From_Mesh(robj) != 0;
+        } else if (id == RenderObjClass::CLASSID_HLOD) {
+            state = geo->Init_From_HLOD(robj) != 0;
+        }
+
+        if (state != 1) {
+            geo->Release_Ref();
```
xezon: Something is not right here. First of all if code would reach here, application would crash a little further down below, because `geo` will be released (deleted) twice.

Second, `state` looks like it should be a `bool`?

src/game/client/shadow/w3dvolumetricshadow.h
```diff
@@ -0,0 +1,157 @@
...
+    HashTableClass *m_missingGeomTable;
+};
+
+class W3DShadowGeometryManagerIterator : public HashTableIteratorClass
+{
+public:
+    W3DShadowGeometryManagerIterator(W3DShadowGeometryManager &manager) : HashTableIteratorClass(*manager.m_geomPtrTable) {}
+    virtual ~W3DShadowGeometryManagerIterator() {}
+
+    W3DShadowGeometry *Get_Current_Geom() { return (W3DShadowGeometry *)Get_Current(); }
```
xezon: `static_cast` ?

src/game/client/shadow/w3dvolumetricshadow.cpp
tomsons26: it does yeah

src/game/client/shadow/w3dvolumetricshadow.cpp
tomsons26: ah damn forgot

src/game/client/shadow/w3dvolumetricshadow.cpp
tomsons26: Yeah

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,512 @@
...
+    m_numPolygons(0),
+    m_polygons(nullptr)
+{
+}
+
+W3DShadowGeometryMesh::~W3DShadowGeometryMesh()
+{
+    Delete_Neighbors();
+
+    if (m_parentVerts != nullptr) {
```
duncanspumpkin, line 45: remove. These do not exist. When you delete[] or delete it implicitly does a nullptr check all you are doing here is adding a unneeded branch.

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,512 @@
...
+        delete[] m_parentVerts;
+    }
+
+    if (m_polygonNormals != nullptr) {
```
duncanspumpkin, line 49: same here

src/game/client/shadow/w3dvolumetricshadow.cpp
duncanspumpkin: I don't see the assert in wb. Assume its from bfme2?

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,512 @@
...
+    }
+
+    return &m_polyNeighbors[index];
+}
+
+void W3DShadowGeometryMesh::Build_Polygon_Neighbors()
+{
+    Build_Polygon_Normals();
+
+    int num = Get_Num_Polygon();
```
duncanspumpkin: may as well make this const as it never changes.

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,512 @@
...
+        if (!Allocate_Neighbors(num)) {
+            return;
+        }
+    }
+
+    for (int i = 0; i < m_numPolyNeighbors; ++i) {
+
+        m_polyNeighbors[i].myIndex = (short)i;
+
+        for (int j = 0; j < 3; ++j) {
```
duncanspumpkin, line 93: you could have alternatively gone for: ```for (auto& neighbor : m_polyNeighbors[i].neighbor) neighbor.neighborIndex = -1;``` which would have avoided the magic number of 3.

src/game/client/shadow/w3dvolumetricshadow.h
```diff
@@ -0,0 +1,157 @@
...
+enum
+{
+    MAX_SHADOW_CASTER_MESHES = 160,
+};
+
+#pragma pack(push, 2)
+
+struct NeighborEdge
+{
+    short neighborIndex;
```
duncanspumpkin, line 40: Are you sure this is correct. Everything seems to suggest that there is 16bits before this field.

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,512 @@
...
+        return false;
+    }
+
+    m_numPolyNeighbors = num_polys;
+    return true;
+}
+
+void W3DShadowGeometryMesh::Delete_Neighbors()
+{
+
```
duncanspumpkin: stray whitespace

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,512 @@
...
+        m_polygonNormals = normals;
+    }
+}
+
+void W3DShadowGeometryMesh::Get_Polygon_Normal(int index, Vector3 *normal) const
+{
+
+    if (m_polygonNormals != nullptr) {
+        *normal = m_polygonNormals[index];
+    }
```
duncanspumpkin, line 222: missing return

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,512 @@
...
+    return 1;
+}
+
+int W3DShadowGeometry::Init_From_HLOD(RenderObjClass *robj)
+{
+    HLodClass *hlod = static_cast<HLodClass *>(robj);
+
+    int lod_index = hlod->Get_LOD_Count() - 1;
+
+    W3DShadowGeometryMesh *geo_mesh = &m_meshList[m_meshCount];
```
duncanspumpkin, line 267: Urgh this is so dangerous. The L274 assert should be up here. We are possibly getting a pointer to invalid memory at this point. I would also suggest that this be changed into a reference and possibly into a function so all access's are safe. `Init_From_Mesh` for example is **not** checking if it has exhausted the mesh count!

```cpp
W3DShadowGeometryMesh& Get_Next_Mesh(){
    captainslog_dbgassert(m_meshCount < MAX_SHADOW_CASTER_MESHES, "Too many shadow sub-meshes"); 
    return m_meshList[m_meshCount];
}
```

Hmm a reference would mean reworking the below function into two so perhaps that is not preferred.

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,512 @@
...
+                                verts[j] = i;
+                                --num;
+                            }
+                        }
+                        verts[i] = i;
+                    }
+                }
+
+                geo_mesh->m_parentVerts = new unsigned short[geo_mesh->m_numVerts];
+                memcpy(geo_mesh->m_parentVerts, verts, 2 * geo_mesh->m_numVerts);
```
duncanspumpkin: `sizeof(int16_t)` instead of 2 we are copying the verts result.

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,512 @@
...
+            }
+        }
+    }
+    return m_meshCount != 0;
+}
+
+int W3DShadowGeometry::Init_From_Mesh(RenderObjClass *robj)
+{
+
+    W3DShadowGeometryMesh *geo_mesh = &m_meshList[m_meshCount];
```
duncanspumpkin, line 357: This is duplicating code almost identical to the line above suggest we deduplicate.

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,512 @@
...
+{
+    m_geomPtrTable = new HashTableClass(2048);
+    m_missingGeomTable = new HashTableClass(2048);
+}
+
+W3DShadowGeometryManager::~W3DShadowGeometryManager()
+{
+    Free_All_Geoms();
+
+    if (m_geomPtrTable != nullptr) {
```
duncanspumpkin, line 425: again not needed

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,512 @@
...
+        delete m_geomPtrTable;
+    }
+
+    if (m_missingGeomTable != nullptr) {
```
duncanspumpkin, line 429: or here

src/game/client/shadow/w3dvolumetricshadow.cpp
xezon: If it goes into bug fix territory, I would prefer it to be a separate commit for visibility (which can be part of this pull request of course).

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,512 @@
...
+
+int W3DShadowGeometryManager::Load_Geom(RenderObjClass *robj, const char *name)
+{
+    // TODO could be cleaner
+
+    int state = 0;
+
+    W3DShadowGeometry *geo = new W3DShadowGeometry;
+
+    if (geo != nullptr) {
```
duncanspumpkin: return early and avoid the indent

src/game/client/shadow/w3dvolumetricshadow.cpp
xezon: Using `ARRAY_SIZE(m_polyNeighbors[i].neighbor)` is also ok.

src/game/client/shadow/w3dvolumetricshadow.cpp
duncanspumpkin: agreed release_ref is only called once in the assembly

src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -0,0 +1,512 @@
...
+        geo->Set_Name(name);
+
+        int id = robj->Class_ID();
+
+        if (id == RenderObjClass::CLASSID_MESH) {
```
duncanspumpkin: I prefer a switch for these feel it looks cleaner.

src/game/client/shadow/w3dvolumetricshadow.h
```diff
@@ -0,0 +1,157 @@
...
+class Vector3;
+
+typedef Vector3i16 TriIndex;
+
+enum
+{
+    MAX_SHADOW_CASTER_MESHES = 160,
+};
+
+#pragma pack(push, 2)
```
duncanspumpkin, line 36: what is the packing for? Nothing looks out of the ordinary packing wise or am i mistaken.

duncanspumpkin (changes requested):
See comments one bug and one area HLOD/Mesh that i think could be a source of bugs which could be easily addressed.

jonwil (commented):
Other than W3DShadowGeometryManager::Load_Geom, all the functions seem to match up from an ASM point of view.

src/game/client/shadow/w3dvolumetricshadow.cpp
tomsons26: marking that as a todo, bfme2 does provide the how but not sure if its the same still

duncanspumpkin (approved):

PR #414, opened 2021-12-18
Move definition of std::clamp into STL Port

duncanspumpkin:
https://en.cppreference.com/w/cpp/algorithm/clamp

Clamp is located in the algorithm header. The way STL Port works _algo.h is the equivalent header.

This also bumps the min c++ version to 17 as that is the version with std::clamp.

### DIFF
--- a/deps/stlport/stl/_algo.h
+++ b/deps/stlport/stl/_algo.h
@@ -584,6 +584,20 @@ void inplace_merge(_BidirectionalIter __first,
 		   _BidirectionalIter __middle,
 		   _BidirectionalIter __last, _Compare __comp);
 
+
+// compare functions.
+
+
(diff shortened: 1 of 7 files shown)


xezon:
Closes #405

deps/stlport/stl/_algo.h
```diff
@@ -22,6 +22,9 @@
  * modified is included with the above copyright notice.
  *
  */
+/*
+ * Code has been modified to add 'clamp' function from C++ 2017. - Duncan Frost 2021
```
xezon: Did you mean to write that name here? Just asking in case you don't want name other than regular nickname in code.

duncanspumpkin: I was just trying to follow the style of the file but I'm also thinking that perhaps this isn't how best to show it anyway since it looks like this is an original sgi file so i should just add an extra line under the other authors

xezon (approved):

OmniBlade (approved):

PR #415, opened 2021-12-18
ExperieceTracker

duncanspumpkin:
Complete

### DIFF
--- a/src/game/logic/object/experiencetracker.cpp
+++ b/src/game/logic/object/experiencetracker.cpp
@@ -1,0 +1,181 @@
+/**
+ * @file
+ *
+ * @author Duncans_pumpkin
+ *
+ * @brief ExperienceTracker
+ *
+ * @copyright Thyme is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#include "experiencetracker.h"
+#include "gamelogic.h"
+#include "object.h"
+#include "xfer.h"
+
+// zh: 0x0061B790 wb: 0x008A35F9
+int32_t ExperienceTracker::Get_Experience_Value(const Object *killer) const
+{
+    if (killer->Get_Relationship(m_parent) == Relationship::ALLIES) {
+        return 0;
+    }
+
+    return m_parent->Get_Template()->Get_Experience_Value(m_currentLevel);
+}
+
+// zh: 0x0061B7D0 wb: 0x008A3637
+bool ExperienceTracker::Is_Trainable() const
+{
+    return m_parent->Get_Template()->Is_Trainable();
+}
+
+// zh: 0x0061B7E0 wb: 0x008A3654
+bool ExperienceTracker::Is_Accepting_Experience_Points() const
+{
+    return Is_Trainable() || m_experienceSink != ObjectID::OBJECT_UNK;
+}
+
+// zh: 0x0061B820 wb: 0x008A36A4
+void ExperienceTracker::Set_Min_Veterency_Level(VeterancyLevel newLevel)
+{
+    if (m_currentLevel >= newLevel) {
+        return;
+    }
+
+    captainslog_dbgassert(m_parent != nullptr, "Parent must not be nullptr!");
+
+    const auto old_level = m_currentLevel;
+    m_currentLevel = newLevel;
+
+    m_currentExperience = m_parent->Get_Template()->Get_Experience_Required(m_currentLevel);
+
+    m_parent->On_Veterancy_Level_Changed(old_level, m_currentLevel, true);
+}
+
+// zh: 0x0061B860 wb: 0x008A370D
+void ExperienceTracker::Set_Veterency_Level(VeterancyLevel newLevel, bool unk)
+{
+    if (m_currentLevel == newLevel) {
+        return;
+    }
+
+    captainslog_dbgassert(m_parent != nullptr, "Parent must not be nullptr!");
+
+    const auto old_level = m_currentLevel;
+    m_currentLevel = newLevel;
+
+    m_currentExperience = m_parent->Get_Template()->Get_Experience_Required(m_currentLevel);
+
+    m_parent->On_Veterancy_Level_Changed(old_level, m_currentLevel, true);
+}
+
+// zh: 0x0061B8A0 wb: 0x008A377A
+bool ExperienceTracker::Gain_Exp_For_Level(int32_t levelsToGain, bool applyMultiplier)
+{
+    const auto new_level = std::min<VeterancyLevel>(m_currentLevel + levelsToGain, VeterancyLevel::VETERANCY_COUNT - 1);
+
+    if (new_level <= m_currentLevel) {
+        return false;
+    }
+
+    const auto exp_for_level_gain = m_parent->Get_Template()->Get_Experience_Required(new_level) - m_currentExperience;
+
+    Add_Experience_Points(exp_for_level_gain, applyMultiplier);
+    return true;
+}
+
+// zh: 0x0061B8F0 wb: 0x008A37E4
+bool ExperienceTracker::Can_Gain_Exp_For_Level(int32_t levelsToGain) const
+{
+    return m_currentLevel != std::min<VeterancyLevel>(m_currentLevel + levelsToGain, VeterancyLevel::VETERANCY_COUNT - 1);
+}
+
+// zh: 0x0061B910 wb: 0x008A381C
+void ExperienceTracker::Add_Experience_Points(int32_t experienceGain, bool applyMultiplier)
+{
+    if (m_experienceSink != ObjectID::OBJECT_UNK) {
+        auto *sink_object = g_theGameLogic->Find_Object_By_ID(m_experienceSink);
+        if (sink_object != nullptr) {
+            sink_object->Get_Experience_Tracker()->Add_Experience_Points(experienceGain * m_experienceMultiplier, applyMultiplier);
+            return;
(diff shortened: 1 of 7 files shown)


src/game/logic/object/experiencetracker.cpp
```diff
@@ -0,0 +1,181 @@
...
+    const auto old_level = m_currentLevel;
+    m_currentLevel = newLevel;
+
+    m_currentExperience = m_parent->Get_Template()->Get_Experience_Required(m_currentLevel);
+
+    m_parent->On_Veterancy_Level_Changed(old_level, m_currentLevel, true);
+}
+
+// zh: 0x0061B860 wb: 0x008A370D
+void ExperienceTracker::Set_Veterency_Level(VeterancyLevel newLevel, bool unk)
```
xezon: `bool unk` is unused and function body of `Set_Veterency_Level` looks exact same as `Set_Min_Veterency_Level`. Is this proper?

src/game/logic/object/experiencetracker.cpp
```diff
@@ -0,0 +1,181 @@
...
+    const auto old_level = m_currentLevel;
+    m_currentLevel = newLevel;
+
+    m_currentExperience = m_parent->Get_Template()->Get_Experience_Required(m_currentLevel);
+
+    m_parent->On_Veterancy_Level_Changed(old_level, m_currentLevel, true);
+}
+
+// zh: 0x0061B8A0 wb: 0x008A377A
+bool ExperienceTracker::Gain_Exp_For_Level(int32_t levelsToGain, bool applyMultiplier)
```
xezon: Thyme Style: `levels_to_gain`, `apply_multiplier`. Same in other functions below and above.

src/game/logic/object/object.cpp
```diff
@@ -23,10 +23,28 @@ ObjectShroudStatus Object::Get_Shrouded_Status(int index) const
     return ObjectShroudStatus(0);
 }
 
+// zh: 0x005479B0 wb: 0x007D0E3D
+Relationship Object::Get_Relationship(const Object *that) const
+{
+#ifdef GAME_DLL
+    return Call_Method<Relationship, const Object, const Object *>(PICK_ADDRESS(0x005479B0, 0x007D0E3D), this, that);
+#else
+    return Relationship();
```
xezon, line 32: What about `that` argument? Should not be passed here?

src/game/logic/object/object.cpp
```diff
@@ -23,10 +23,28 @@ ObjectShroudStatus Object::Get_Shrouded_Status(int index) const
...
 #endif
     return nullptr;
 }
+
+// zh: 0x00549BD0 wb: 0x007D370C
+void Object::On_Veterancy_Level_Changed(VeterancyLevel v1, VeterancyLevel v2, bool b)
+{
+#ifdef GAME_DLL
+    Call_Method<void, Object, VeterancyLevel, VeterancyLevel, bool>(PICK_ADDRESS(0x00549BD0, 0x007D370C), this, v1, v2, b);
+#endif
```
xezon, line 49: This function is missing code for Standalone? Maybe at least add a `#pragma message` clarifying this here.

src/game/logic/object/experiencetracker.cpp
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    const auto exp_for_level_gain = m_parent->Get_Template()->Get_Experience_Required(new_level) - m_currentExperience;
+
+    Add_Experience_Points(exp_for_level_gain, applyMultiplier);
+    return true;
+}
+
+// zh: 0x0061B8F0 wb: 0x008A37E4
+bool ExperienceTracker::Can_Gain_Exp_For_Level(int32_t levelsToGain) const
```
xezon: This function is not used inside this class. So class user must call it before adding Exp. Is this proper?

src/game/logic/object/experiencetracker.cpp
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// zh: 0x0061B7D0 wb: 0x008A3637
+bool ExperienceTracker::Is_Trainable() const
+{
+    return m_parent->Get_Template()->Is_Trainable();
+}
+
+// zh: 0x0061B7E0 wb: 0x008A3654
+bool ExperienceTracker::Is_Accepting_Experience_Points() const
```
xezon, line 37: This function is not used inside this class. So class user must call it before adding EXP. Is this proper?
`Can_Gain_Exp_For_Level` is also not used. But `Is_Trainable()` is. I am asking, because if checks are missing, then EXP may be added when it should not be. But it could be of course that class user calls these originally.

src/game/logic/object/experiencetracker.cpp
```diff
@@ -0,0 +1,181 @@
...
+{
+    xfer->xferInt(&m_currentExperience);
+    xfer->xferUser(&m_currentLevel, sizeof(m_currentLevel));
+}
+
+// zh: 0x0061BAA0 wb: 0x008A3A51
+void ExperienceTracker::Xfer_Snapshot(Xfer *xfer)
+{
+    uint8_t version = 1;
+    xfer->xferVersion(&version, 1);
```
xezon, line 177: `sizeof(version)`

src/game/logic/object/experiencetracker.cpp
duncanspumpkin: Yeah unfortunately it is required for abi compatibility.

src/game/logic/object/object.cpp
duncanspumpkin: This is just to return a default value for the unimplemented function

src/game/logic/object/object.cpp
duncanspumpkin: This is the style of all functions that are unimplemented

src/game/logic/object/experiencetracker.cpp
duncanspumpkin: Just how the vanilla function is

src/game/logic/object/experiencetracker.cpp
duncanspumpkin: Again just how vanilla is

src/game/logic/object/experiencetracker.cpp
duncanspumpkin: The 1 represents the version that it should be https://github.com/TheAssemblyArmada/Thyme/blob/84e604c484f5fbd890a570a40ad4593eb1feddc1/src/game/common/system/xfer.cpp#L32

You have to remember there are two callers for this function save and load. Under save the 1st parameter, version, is set to 1 and serialised. Under load the 1st parameter, version, is deserialised and verified to match the second parameter.

xezon: Hmm ok. I wasn't sure what I am looking at here, because next line passes size.

duncanspumpkin: Yeah xfer functions are a bit weird. If you look at a few other classes xfers you'll see the general pattern.

src/game/logic/object/experiencetracker.cpp
duncanspumpkin: And it is slightly different to `Set_Min` the if statement is different.

src/game/logic/object/experiencetracker.cpp
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    captainslog_dbgassert(m_parent != nullptr, "Parent must not be nullptr!");
+
+    const auto old_level = m_currentLevel;
+    m_currentLevel = new_level;
+
+    m_currentExperience = m_parent->Get_Template()->Get_Experience_Required(m_currentLevel);
+
+    m_parent->On_Veterancy_Level_Changed(old_level, m_currentLevel, true);
```
jonwil: This is passing unk as the last parameter to On_Veterancy_Level_Changed, please fix.

src/game/logic/object/experiencetracker.cpp
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// zh: 0x0061B910 wb: 0x008A381C
+void ExperienceTracker::Add_Experience_Points(int32_t experience_gain, bool apply_multiplier)
+{
+    if (m_experienceSink != ObjectID::OBJECT_UNK) {
+        auto *sink_object = g_theGameLogic->Find_Object_By_ID(m_experienceSink);
+        if (sink_object != nullptr) {
+            sink_object->Get_Experience_Tracker()->Add_Experience_Points(
+                experience_gain * m_experienceMultiplier, apply_multiplier);
```
jonwil, line 104: experience_gain is typecast to float before the multiply happens and then back to int after the multiply. Should those casts be there?

src/game/logic/object/experiencetracker.cpp
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    if (!Is_Trainable()) {
+        return;
+    }
+
+    const auto old_level = m_currentLevel;
+    if (apply_multiplier) {
+        experience_gain *= m_experienceMultiplier;
```
jonwil, line 115: experience_gain is typecast to float before the multiply happens and then back to int after the multiply. Should those casts be there?

jonwil (changes requested):
A couple things I found.

src/game/logic/object/experiencetracker.cpp
duncanspumpkin: `experienceMultiplier` is float so the whole calculation will be done in floats and then converted back to int for the function call.

src/game/logic/object/experiencetracker.cpp
duncanspumpkin: not required it will do a promotion of the int to a float, do the multiply, then demote back to int.

jonwil (approved):
Looks good to me now.

PR #420, opened 2021-12-21
Implements FrameMetrics class.

tomsons26:


src/game/network/framemetrics.cpp
```diff
@@ -0,0 +1,114 @@
...
+    m_averageFps(0.0f),
+    m_averageLatency(0.0f),
+    m_cushionIndex(0),
+    m_fpsListIndex(0),
+    m_lastFpsTimeThing(0),
+    m_minimumCushion(0)
+{
+    m_pendingLatencies = new int[MAX_FRAMES_AHEAD];
+
+    for (int i = 0; i < MAX_FRAMES_AHEAD; ++i) {
```
xezon, line 33: You could remove this by writing `new int[MAX_FRAMES_AHEAD]()` above. Notice the parenthesis at the end.

src/game/network/framemetrics.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+void FrameMetrics::Init()
+{
+    m_averageFps = 30.0f;
+    m_averageLatency = 0.2f;
+
+    m_minimumCushion = -1;
+
+    for (int i = 0; i < g_theWriteableGlobalData->m_networkFPSHistoryLength; ++i) {
```
xezon, line 59: This does not look good. This would crash if `g_theWriteableGlobalData->m_networkFPSHistoryLength` was increased later at runtime. I suggest adding a TODO comment near the class members:

```c++
// #TODO change from C array to std::vector for robustness if ABI allows for it.
```

src/game/network/framemetrics.h
```diff
@@ -0,0 +1,54 @@
...
+
+#ifdef GAME_DLL
+    FrameMetrics *Hook_Ctor() { return new (this) FrameMetrics; }
+    void Hook_Dtor() { FrameMetrics::~FrameMetrics(); }
+
+#endif
+
+private:
+    float *m_fpsList;
+    int m_lastFpsTimeThing;
```
xezon: Is this name proper? `Thing` ? Maybe just `m_lastTime`

src/game/network/framemetrics.cpp
```diff
@@ -0,0 +1,114 @@
...
+        m_fpsList[i] = 30.0f;
+    }
+
+    m_fpsListIndex = 0;
+
+    for (int i = 0; i < g_theWriteableGlobalData->m_networkLatencyHistoryLength; ++i) {
+        m_fpsList[i] = 0.2f;
```
xezon: `m_latencyList`

src/game/network/framemetrics.cpp
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    m_cushionIndex = 0;
+}
+
+void FrameMetrics::Reset()
+{
+    Init();
```
xezon, line 74: Calling Init() will not reset the following members:
```
m_pendingLatencies
m_lastFpsTimeThing
```
Is this proper?

src/game/network/framemetrics.cpp
```diff
@@ -0,0 +1,114 @@
...
+{
+    m_pendingLatencies = new int[MAX_FRAMES_AHEAD];
+
+    for (int i = 0; i < MAX_FRAMES_AHEAD; ++i) {
+        m_pendingLatencies[i] = 0;
+    }
+
+    m_fpsList = new float[g_theWriteableGlobalData->m_networkFPSHistoryLength];
+
+    m_latencyList = new float[g_theWriteableGlobalData->m_networkLatencyHistoryLength];
```
xezon, line 39: `m_fpsList` and `m_latencyList` are not default initialized here. It seems to be no practical problem because of Init() function, but could still zero initialize if this is not a hot class.

src/game/network/framemetrics.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    m_pendingLatencies[frame % MAX_FRAMES_AHEAD] = time;
+}
+
+void FrameMetrics::Process_Latency_Response(unsigned int frame)
+{
+    int time_diff = (int)rts::Get_Time() - m_pendingLatencies[frame % MAX_FRAMES_AHEAD];
+    unsigned int index = frame % g_theWriteableGlobalData->m_networkLatencyHistoryLength;
+
+    m_averageLatency -= m_latencyList[index] / (float)(unsigned int)g_theWriteableGlobalData->m_networkLatencyHistoryLength;
```
xezon, line 109: `(float)(unsigned int)` is as designed?

tomsons26: Yeah its strange, length is read as int from ini but this implies uint so not sure

src/game/network/framemetrics.cpp
tomsons26: i prefer it how it is

jonwil (approved):
Looks good to me.

PR #424, opened 2021-12-22
Implements Energy class

tomsons26:


src/game/common/rts/energy.cpp
```diff
@@ -0,0 +1,190 @@
...
+    if (g_theGameLogic->Get_Frame() < m_frame) {
+        return false;
+    }
+
+    return m_energyProduction >= m_energyConsumption;
+}
+
+void Energy::Adjust_Power(int amount, bool positive)
+{
+    if (amount) {
```
xezon: This condition looks obsolete because of the next line.

src/game/common/rts/energy.h
```diff
@@ -0,0 +1,69 @@
...
+{
+public:
+    Energy();
+
+    virtual ~Energy() {}
+
+    // SnapShot interface
+    virtual void CRC_Snapshot(Xfer *xfer) {}
+    virtual void Xfer_Snapshot(Xfer *xfer);
+    virtual void Load_Post_Process() {}
```
xezon: Can add `override` keyword for robustness.

src/game/common/rts/energy.cpp
```diff
@@ -0,0 +1,190 @@
...
+#include "player.h"
+#include "playerlist.h"
+
+Energy::Energy() : m_energyProduction(0), m_energyConsumption(0), m_frame(0), m_player(nullptr) {}
+
+void Energy::Init(Player *player)
+{
+    m_energyProduction = 0;
+    m_energyConsumption = 0;
+    m_player = player;
```
xezon, line 29: `m_frame` is not set to 0 here? Hard to tell if it should.

src/game/common/rts/playerlist.h
```diff
@@ -44,10 +44,19 @@ class PlayerList : public SubsystemInterface, public SnapShot
         return m_local;
     }
 
+    Player *Get_Nth_Player(int i)
+    {
+        if (i < 0 || i >= 16) {
```
xezon, line 50: `i >= ARRAY_SIZE(m_players)`

src/game/common/rts/energy.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+float Energy::Get_Energy_Supply_Ratio() const
+{
+    captainslog_dbgassert(m_energyProduction >= 0 && m_energyConsumption >= 0, "Negative Energy numbers!");
+
+    if (g_theGameLogic->Get_Frame() < m_frame) {
+        return 0.0f;
+    }
+
+    if (m_energyConsumption) {
```
xezon: `m_energyConsumption != 0` ?

src/game/common/rts/playerlist.h
tomsons26: It will actually be a enum once we add it, not yet sure what to call it, will add a todo noting that tho

src/game/common/rts/energy.cpp
tomsons26: isn't no, yeah hard to tell

src/game/common/rts/energy.cpp
tomsons26: actually should be  != 0 i guess

src/game/common/rts/energy.cpp
```diff
@@ -0,0 +1,191 @@
...
+    if (g_theGameLogic->Get_Frame() < m_frame) {
+        return false;
+    }
+
+    return m_energyProduction >= m_energyConsumption;
+}
+
+void Energy::Adjust_Power(int amount, bool positive)
+{
+    if (amount != 0) {
```
xezon, line 67: `amount > 0` inherintly is `amount != 0`, therefore this line is obsolete.

tomsons26: Thats true only for unsigned, power adjustments occur with positives and negatives, assembly also has JNZ not JG so this is correct

jonwil (approved):
Looks good to me. Approved.

PR #438, opened 2021-12-28
Fix compile errors

xezon:
Not sure if correct. But does fix compile errors.

### DIFF
--- a/src/platform/w3dengine/client/w3ddisplay.cpp
+++ b/src/platform/w3dengine/client/w3ddisplay.cpp
@@ -65,8 +65,16 @@
 #endif
 #include <stdio.h>
 
+#ifdef GAME_DLL
+#include "hooker.h"
+#endif
+
 #ifdef GAME_DEBUG_STRUCTS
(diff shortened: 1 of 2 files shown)


src/platform/w3dengine/client/w3ddisplay.cpp
```diff
@@ -65,8 +65,16 @@
 #endif
 #include <stdio.h>
 
+#ifdef GAME_DLL
+#include "hooker.h"
+#endif
+
 #ifdef GAME_DEBUG_STRUCTS
+#ifdef GAME_DLL
 ICoord2D &s_leftClickReleased = Make_Global<ICoord2D>(0x00E1B118);
```
xezon, line 74: Is it supposed to be `ICoord2D` or `ICoord2D *` ?

jonwil: Its supposed to be ICoord2D.

jonwil (approved):
Looks good to me.

PR #439, opened 2021-12-28
Fix compile warning in w3d

xezon:
signed unsigned comparison mismatch


jonwil (changes requested):
I don't like this change. Would prefer it to just be a change from "int" to "unsigned int". No reason to use "auto" here and no reason to change from i-- to --i.

xezon:
Fixed.

jonwil (approved):

PR #441, opened 2021-12-28
Fix compile warnings in segline

xezon:


src/w3d/renderer/segline.cpp
```diff
@@ -73,7 +73,7 @@ void SegmentedLineClass::Set_Point_Location(unsigned int point_idx, const Vector
     Invalidate_Cached_Bounding_Volumes();
 }
 
-void SegmentedLineClass::Get_Point_Location(unsigned int point_idx, Vector3 &loc)
+void SegmentedLineClass::Get_Point_Location(int point_idx, Vector3 &loc)
 {
     if (point_idx < m_pointLocations.Count()) {
```
xezon: `m_pointLocations` (SimpleDynVecClass<>) is using `int` for indexing. Therefore `unsigned int` in this class will issue compiler warning when interacting with member class.

src/w3d/renderer/segline.cpp
```diff
@@ -388,7 +388,7 @@ float SegmentedLineClass::Get_Post_Increment_Value() const
 
     float poly = (float)Get_Num_Polys();
     float poly2 = poly + poly;
-    float p = (1.0 - 0.5 / (poly2 * poly2));
+    float p = (1.0f - 0.5f / (poly2 * poly2));
```
xezon, line 391: Compiler warning here.

jonwil:
We have clear proof that these functions take an unsigned int as a parameter and I would like to keep things matching. The compiler error could be fixed by a cast inside the function.

xezon:
Ok. I reinstated unsigned and added static_cast.

jonwil (approved):
Looks good to me.

PR #442, opened 2021-12-28
Fix compile warning in w3ddisplay with signed integer mismatch

xezon:


src/platform/w3dengine/client/w3ddisplay.cpp
```diff
@@ -486,8 +486,12 @@ void W3DDisplay::Draw()
                         int h = (int)((float)g_theDisplay->Get_Height() * 0.9f);
                         int w;
 
-                        if (display_str->Get_Width(-1) <= g_theDisplay->Get_Width()) {
-                            w = (g_theDisplay->Get_Width() - display_str->Get_Width(-1)) / 2;
+                        // #BUGFIX Cache values and cast unsigned to signed.
+                        const int str_width = display_str->Get_Width(-1);
```
duncanspumpkin, line 497: That's not a bug fix. You haven't changed anything logically apart from cacheing in which case it's not a bug fix just a change.

xezon: I agree. I wasn't sure if should label it. I was wondering if new variables in function body will confuse comparison with de-compiler.

xezon: Fixed

jonwil (approved):
This looks good to me, approved.

PR #443, opened 2021-12-28
Fix compile warning in w3ddisplay from deprecated GetVersionEx function

xezon:


jonwil:
Given that VER_PLATFORM_WIN32_WINDOWS means "Windows 9x" (at least according to available documentation) maybe its better to just remove the check entirely (given that we don't even run on Windows 9x)

xezon:
Yes that is what was done within the ifdef block. I could delete the code if so desired, but I left it for now for documentation sake.

jonwil:
No, I meant removing the whole lot including the W3D::Invalidate_Textures call (since that call is inside the "if version == Windows 9x" block)

xezon:
Hmm. Why do you think that is an appropriate action? Not invalidating textures on Device Reset will likely cause Direct X to not be able to recover.

jonwil:
The original game code is not calling Invalidate_Textures on modern windows so why do we need to do that?

xezon:
I will take a look later.

jonwil:
Its basically impossible for VersionInformation.dwPlatformId == VER_PLATFORM_WIN32_WINDOWS to be true on any version of Windows Thyme supports.

duncanspumpkin:
We already use git for source control we have absolutely no need to add unusable code branches. Just delete the code.

xezon:
Maybe comment out the whole function for now and add a TODO to remove finally later if never used?

duncanspumpkin:
Why, we have already investigated and concluded is not used. Have a comment that there used to be special win9x code if you want. If in the future you can always use git history to find the original code

xezon:
I would just like the Compiler Warning gone. Removing the function entirely will likely make it forgotten, and if ever someone wants it back, he likely re-implements it. No one will remember to check GIT history and where in time.

xezon:
Is this good to go? Review is very small so should not take long to review.

duncanspumpkin:
I would still like to see the dead code removed as it has already been investigated and if anyone for some reason wanted to view the original they can use git to find it.

xezon:
Endless discussions. I can't search Git like I can search in Visual Studio. If we can't move forward with this then we have to keep this compiler warning in the build.

jonwil:
PR solved by disabling the code via "#if 0". Closing.

PR #444, opened 2021-12-28
Fix warning in w3dshadow

xezon:
This is placeholder code, but it issues compiler warning unnecessarily.


jonwil (approved):
Looks good to me.

PR #445, opened 2021-12-28
Fix motchan types and checks

xezon:
I was looking through motchan once again because I still have another PR open on it... I saw a few things. Please check the commit section of this Pull Request to see each individual change applied and the rationale behind them.

### DIFF
--- a/src/w3d/renderer/motchan.cpp
+++ b/src/w3d/renderer/motchan.cpp
@@ -526,7 +526,7 @@ Quaternion AdaptiveDeltaMotionChannelClass::Get_Quat_Vector(float frame_idx)
 
 float AdaptiveDeltaMotionChannelClass::Get_Frame(unsigned int frame_idx, unsigned int vector_idx)
 {
-    float Dst[4];
+    float dst[4];
 
     if (frame_idx >= m_numFrames) {
         frame_idx = m_numFrames - 1;
@@ -552,8 +552,8 @@ float AdaptiveDeltaMotionChannelClass::Get_Frame(unsigned int frame_idx, unsigne
 
         captainslog_assert(m_vectorLen <= 4);
 
-        memcpy(Dst, &m_cacheData[m_vectorLen], 4 * m_vectorLen);
-        Decompress(m_cacheFrame, Dst, frame_idx, m_cacheData);
+        memcpy(dst, &m_cacheData[m_vectorLen], 4 * m_vectorLen);
+        Decompress(m_cacheFrame, dst, frame_idx, m_cacheData);
         m_cacheFrame = frame_idx;
 
         if (frame_idx != m_numFrames - 1) {
@@ -567,7 +567,6 @@ float AdaptiveDeltaMotionChannelClass::Get_Frame(unsigned int frame_idx, unsigne
 void AdaptiveDeltaMotionChannelClass::Decompress(
     unsigned int src_idx, float *srcdata, unsigned int frame_idx, float *outdata)
 {
-
     char dst[4];
 
     captainslog_assert(src_idx < frame_idx);
--- a/src/w3d/renderer/motchan.cpp
+++ b/src/w3d/renderer/motchan.cpp
@@ -297,12 +297,12 @@ void TimeCodedMotionChannelClass::Set_Identity(float *setvec) const
     }
 }
 
-unsigned long TimeCodedMotionChannelClass::Get_Index(unsigned int timecode)
+unsigned int TimeCodedMotionChannelClass::Get_Index(unsigned int timecode)
 {
     captainslog_assert(m_cachedIdx <= m_lastTimeCodeIdx);
 
     if (timecode < Get_Frame_From_Data(m_data[m_cachedIdx])) {
-        unsigned long result = TimeCodedMotionChannelClass::Binary_Search_Index(timecode);
+        unsigned int result = TimeCodedMotionChannelClass::Binary_Search_Index(timecode);
         m_cachedIdx = result;
         return result;
     }
@@ -311,7 +311,7 @@ unsigned long TimeCodedMotionChannelClass::Get_Index(unsigned int timecode)
         return m_cachedIdx;
     }
 
-    unsigned long result = m_packetSize + m_cachedIdx;
+    unsigned int result = m_packetSize + m_cachedIdx;
 
     if (timecode < Get_Frame_From_Data(m_data[m_packetSize + m_cachedIdx])) {
         return m_cachedIdx;
@@ -331,12 +331,12 @@ unsigned long TimeCodedMotionChannelClass::Get_Index(unsigned int timecode)
     return result;
 }
 
-unsigned long TimeCodedMotionChannelClass::Binary_Search_Index(unsigned int timecode) const
+unsigned int TimeCodedMotionChannelClass::Binary_Search_Index(unsigned int timecode) const
 {
     int count2;
     int count = 0;
     int rightIdx = m_numTimeCodes - 2;
-    unsigned long result = m_lastTimeCodeIdx;
+    unsigned int result = m_lastTimeCodeIdx;
 
     if (timecode < Get_Frame_From_Data(m_data[m_lastTimeCodeIdx])) {
         while (true) {
@@ -387,7 +387,7 @@ bool TimeCodedBitChannelClass::Load_W3D(ChunkLoadClass &cload)
 {
     W3dTimeCodedBitChannelStruct chan;
     Free();
-    unsigned int chunk_size = cload.Cur_Chunk_Length();
+    const unsigned int chunk_size = cload.Cur_Chunk_Length();
 
     if (cload.Read(&chan, sizeof(chan)) != sizeof(chan)) {
         return false;
@@ -398,11 +398,11 @@ bool TimeCodedBitChannelClass::Load_W3D(ChunkLoadClass &cload)
     m_pivotIdx = chan.Pivot;
     m_defaultVal = chan.DefaultVal;
     m_cachedIdx = 0;
-    int bytesleft = 4 * m_numTimeCodes - 4;
+    const int bytesleft = 4 * m_numTimeCodes - 4;
 
     captainslog_assert((sizeof(chan) + bytesleft) == chunk_size);
 
-    m_bits = new unsigned long[m_numTimeCodes];
+    m_bits = new unsigned int[m_numTimeCodes];
 
     captainslog_assert(m_bits);
 
--- a/src/w3d/renderer/motchan.cpp
+++ b/src/w3d/renderer/motchan.cpp
@@ -550,7 +550,7 @@ float AdaptiveDeltaMotionChannelClass::Get_Frame(unsigned int frame_idx, unsigne
         return m_cacheData[vector_idx + m_vectorLen];
     } else {
 
-        captainslog_assert(m_vectorLen <= 4);
+        captainslog_assert(m_vectorLen < 4);
 
         memcpy(dst, &m_cacheData[m_vectorLen], 4 * m_vectorLen);
         Decompress(m_cacheFrame, dst, frame_idx, m_cacheData);
--- a/src/w3d/renderer/motchan.cpp
+++ b/src/w3d/renderer/motchan.cpp
@@ -285,7 +285,7 @@ Quaternion TimeCodedMotionChannelClass::Get_Quat_Vector(float frame_idx)
     }
 }
 
-void TimeCodedMotionChannelClass::Set_Identity(float *setvec)
+void TimeCodedMotionChannelClass::Set_Identity(float *setvec) const
 {
     if (m_type == ANIM_CHANNEL_Q) {
         setvec[0] = 0.0f;
@@ -331,7 +331,7 @@ unsigned long TimeCodedMotionChannelClass::Get_Index(unsigned int timecode)
     return result;
 }
 
-unsigned long TimeCodedMotionChannelClass::Binary_Search_Index(unsigned int timecode)
+unsigned long TimeCodedMotionChannelClass::Binary_Search_Index(unsigned int timecode) const
 {
     int count2;
     int count = 0;
--- a/src/w3d/renderer/motchan.h
+++ b/src/w3d/renderer/motchan.h
@@ -70,7 +70,7 @@ public:
 
 private:
     unsigned int m_pivotIdx; // Could be signed, but original uses unsigned, so we do same for now.
-    unsigned long m_type;
+    unsigned int m_type; // Could be signed, but original uses unsigned, so we do same for now.
     int m_vectorLen;
     int m_unusedFloat1;
     int m_unusedFloat2;
@@ -105,7 +105,7 @@ public:
 
 private:
     unsigned int m_pivotIdx;
-    unsigned long m_type;
+    unsigned int m_type;
     int m_defaultVal;
     int m_firstFrame;
     int m_lastFrame;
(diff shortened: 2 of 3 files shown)


xezon:
This PR can be squashed when merged.

src/w3d/renderer/motchan.cpp
```diff
@@ -385,11 +385,11 @@ bool TimeCodedBitChannelClass::Load_W3D(ChunkLoadClass &cload)
...
     m_defaultVal = chan.DefaultVal;
     m_cachedIdx = 0;
-    int bytesleft = 4 * m_numTimeCodes - 4;
+    const int bytesleft = 4 * m_numTimeCodes - 4;
 
-    captainslog_assert((sizeof(chan) + bytesleft) == (unsigned)chunk_size);
+    captainslog_assert((sizeof(chan) + bytesleft) == chunk_size);
 
-    m_bits = new unsigned long[m_numTimeCodes];
+    m_bits = new unsigned int[m_numTimeCodes];
```
jonwil, line 405: Confirmed as being unsigned long in the original.

src/w3d/renderer/motchan.h
```diff
@@ -54,23 +54,23 @@ class MotionChannelClass : public W3DMPO
...
+                setvec[i] = 0.0f;
             }
         }
     }
 
 private:
-    unsigned long m_pivotIdx;
-    unsigned long m_type;
+    int m_pivotIdx;
+    int m_type;
```
jonwil: These 2 are confirmed as unsigned long in the original.

src/w3d/renderer/motchan.h
```diff
@@ -28,10 +28,10 @@ class MotionChannelClass : public W3DMPO
     void Free();
     bool Load_W3D(ChunkLoadClass &cload);
 
-    int Get_Type() { return m_type; }
-    int Get_Pivot() { return m_pivotIdx; }
+    int Get_Type() const { return m_type; }
```
jonwil, line 31: Confirmed as const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -28,10 +28,10 @@ class MotionChannelClass : public W3DMPO
...
+    int Get_Pivot() const { return m_pivotIdx; }
```
jonwil, line 32: Confirmed as const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -28,10 +28,10 @@ class MotionChannelClass : public W3DMPO
...
 
-    void Get_Vector(int frame, float *vector)
+    void Get_Vector(int frame, float *vector) const
```
jonwil, line 34: Confirmed as const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -44,7 +44,7 @@ class MotionChannelClass : public W3DMPO
         }
     }
 
-    void Get_Vector_As_Quat(int frame, Quaternion &quat)
+    void Get_Vector_As_Quat(int frame, Quaternion &quat) const
```
jonwil, line 47: Confirmed as const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -54,23 +54,23 @@ class MotionChannelClass : public W3DMPO
         }
     }
 
-    void Set_Identity(float *setvec)
+    void Set_Identity(float *setvec) const
```
jonwil, line 57: Confirmed as const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -90,10 +90,10 @@ class BitChannelClass : public W3DMPO
     void Free();
     bool Load_W3D(ChunkLoadClass &cload);
 
-    int Get_Type() { return m_type; }
-    int Get_Pivot() { return m_pivotIdx; }
+    int Get_Type() const { return m_type; }
```
jonwil, line 93: Confirmed as const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -90,10 +90,10 @@ class BitChannelClass : public W3DMPO
...
+    int Get_Pivot() const { return m_pivotIdx; }
```
jonwil, line 94: Confirmed as const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -90,10 +90,10 @@ class BitChannelClass : public W3DMPO
...
 
-    int Get_Bit(int frame)
+    int Get_Bit(int frame) const
```
jonwil, line 96: Confirmed as const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -104,8 +104,8 @@ class BitChannelClass : public W3DMPO
     }
 
 private:
-    unsigned long m_pivotIdx;
-    unsigned long m_type;
+    int m_pivotIdx;
+    int m_type;
```
jonwil: These 2 are confirmed as unsigned long in the original.

src/w3d/renderer/motchan.h
```diff
@@ -121,17 +121,17 @@ class TimeCodedMotionChannelClass : W3DMPO
     virtual ~TimeCodedMotionChannelClass() override;
     void Free();
     bool Load_W3D(ChunkLoadClass &cload);
-    int Get_Type() { return m_type; }
-    int Get_Pivot() { return m_pivotIdx; }
+    int Get_Type() const { return m_type; }
```
jonwil, line 124: This function isn't const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -121,17 +121,17 @@ class TimeCodedMotionChannelClass : W3DMPO
...
+    int Get_Pivot() const { return m_pivotIdx; }
```
jonwil, line 125: This function isn't const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -121,17 +121,17 @@ class TimeCodedMotionChannelClass : W3DMPO
...
-    unsigned long Binary_Search_Index(unsigned int timecode);
+    void Set_Identity(float *setvec) const;
+    unsigned int Get_Index(unsigned int timecode);
+    unsigned int Binary_Search_Index(unsigned int timecode) const;
 
 private:
-    unsigned long m_pivotIdx;
-    unsigned long m_type;
+    int m_pivotIdx;
+    int m_type;
```
jonwil: These 2 are confirmed as unsigned long in the original.

src/w3d/renderer/motchan.h
```diff
@@ -148,17 +148,17 @@ class TimeCodedBitChannelClass : W3DMPO
     virtual ~TimeCodedBitChannelClass() override;
     void Free();
     bool Load_W3D(ChunkLoadClass &cload);
-    int Get_Type() { return m_type; }
-    int Get_Pivot() { return m_pivotIdx; }
+    int Get_Type() const { return m_type; }
```
jonwil, line 151: This function isn't const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -148,17 +148,17 @@ class TimeCodedBitChannelClass : W3DMPO
...
+    int Get_Pivot() const { return m_pivotIdx; }
```
jonwil, line 152: This function isn't const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -148,17 +148,17 @@ class TimeCodedBitChannelClass : W3DMPO
...
     int Get_Bit(int frame);
 
 private:
-    unsigned long m_pivotIdx;
-    unsigned long m_type;
+    int m_pivotIdx;
+    int m_type;
```
jonwil: These 2 are confirmed as unsigned long in the original.

src/w3d/renderer/motchan.h
```diff
@@ -148,17 +148,17 @@ class TimeCodedBitChannelClass : W3DMPO
...
     int m_defaultVal;
-    unsigned long m_numTimeCodes;
-    unsigned long m_cachedIdx;
-    unsigned long *m_bits;
+    unsigned int m_numTimeCodes;
+    unsigned int m_cachedIdx;
+    unsigned int *m_bits;
```
jonwil, line 161: These 3 are confirmed as unsigned long in the original.

src/w3d/renderer/motchan.h
```diff
@@ -169,17 +169,17 @@ class AdaptiveDeltaMotionChannelClass : W3DMPO
...
     Quaternion Get_Quat_Vector(float frame_idx);
     float Get_Frame(unsigned int frame_idx, unsigned int vector_idx);
     void Decompress(unsigned int src_idx, float *srcdata, unsigned int frame_idx, float *outdata);
     void Decompress(unsigned int frame_idx, float *outdata);
 
 private:
-    unsigned int m_pivotIdx;
-    unsigned int m_type;
+    int m_pivotIdx;
+    int m_type;
```
jonwil: These 2 are confirmed as unsigned long in the original.

src/w3d/renderer/motchan.h
```diff
@@ -169,17 +169,17 @@ class AdaptiveDeltaMotionChannelClass : W3DMPO
     virtual ~AdaptiveDeltaMotionChannelClass() override;
     void Free();
     bool Load_W3D(ChunkLoadClass &cload);
-    int Get_Type() { return m_type; }
-    int Get_Pivot() { return m_pivotIdx; }
+    int Get_Type() const { return m_type; }
```
jonwil, line 172: This function isn't const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -169,17 +169,17 @@ class AdaptiveDeltaMotionChannelClass : W3DMPO
...
+    int Get_Pivot() const { return m_pivotIdx; }
```
jonwil, line 173: This function isn't const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -121,17 +121,17 @@ class TimeCodedMotionChannelClass : W3DMPO
...
     void Get_Vector(float frame, float *setvec);
     Quaternion Get_Quat_Vector(float frame_idx);
-    void Set_Identity(float *setvec);
-    unsigned long Get_Index(unsigned int timecode);
-    unsigned long Binary_Search_Index(unsigned int timecode);
+    void Set_Identity(float *setvec) const;
```
jonwil, line 128: This function isn't const in the original.

src/w3d/renderer/motchan.h
```diff
@@ -121,17 +121,17 @@ class TimeCodedMotionChannelClass : W3DMPO
...
+    unsigned int Get_Index(unsigned int timecode);
```
jonwil, line 129: Confirmed in the original as returning unsigned long.

src/w3d/renderer/motchan.h
```diff
@@ -121,17 +121,17 @@ class TimeCodedMotionChannelClass : W3DMPO
...
+    unsigned int Binary_Search_Index(unsigned int timecode) const;
```
jonwil, line 130: Confirmed in the original as returning unsigned long.
Also this function isn't const in the original.

jonwil (commented):
Matched some things against the original and added notes based on what I found.

tomsons26 (changes requested):
Problem with touching any of this is that means you need to regression test all animations in the game
Not something i want to deal with right now, i will check this PR later at some point and see what can be changed safely so leave it open for now

xezon:
Regarding the const functions: it is just a compiler flag here telling compiler to generate error on non-const class member access. It does not change functionality. That is safe.

Regarding the `unsigned long` vs `int` - I would like to point out that the original code uses a mess of types. See below.

### Example 1

```c++
    MotionChannelClass::unsigned long m_pivotIdx;
    MotionChannelClass::unsigned long m_type;
    
    MotionChannelClass::int Get_Type() { return m_type; }
    MotionChannelClass::int Get_Pivot() { return m_pivotIdx; }
```
Note how these types are stored as `unsigned long` in class but returned as `int`.

### Example 2

```c++
TimeCodedMotionChannelClass::unsigned int m_cachedIdx
TimeCodedBitChannelClass::unsigned long m_cachedIdx
```
One class uses `unsigned int` and the other `unsigned long` for the same kind of thing.

### Example 3

```c++
TimeCodedMotionChannelClass::unsigned long Get_Index(unsigned int timecode)
```
is used as
```c++
void TimeCodedMotionChannelClass::Get_Vector(float frame, float *setvec)
{
    int index = Get_Index(frame);
}
```
`unsigned long` is used as `int`, and `float` is used as `unsigned int`.

This odd type mess goes on and on. It's not good.

src/w3d/renderer/motchan.cpp
duncanspumpkin: ~~strongly suggest changing all these to `uint64_t` as then its much easier to read.~~

xezon:
### Regarding unsigned long m_type

It should be int, because the enum values are integer too. (enum is int by default)

```c++
enum
{
    ANIM_CHANNEL_X = 0,
    ANIM_CHANNEL_Y,
    ANIM_CHANNEL_Z,
    ANIM_CHANNEL_XR,
    ANIM_CHANNEL_YR,
    ANIM_CHANNEL_ZR,
    ANIM_CHANNEL_Q,

    ANIM_CHANNEL_TIMECODED_X,
    ANIM_CHANNEL_TIMECODED_Y,
    ANIM_CHANNEL_TIMECODED_Z,
    ANIM_CHANNEL_TIMECODED_Q,

    ANIM_CHANNEL_ADAPTIVEDELTA_X,
    ANIM_CHANNEL_ADAPTIVEDELTA_Y,
    ANIM_CHANNEL_ADAPTIVEDELTA_Z,
    ANIM_CHANNEL_ADAPTIVEDELTA_Q,
};

enum
{
    ANIM_FLAVOR_TIMECODED = 0,
    ANIM_FLAVOR_ADAPTIVE_DELTA,
    ANIM_FLAVOR_VALID
};
```

### Regarding unsigned long m_pivotIdx

It is used in int context here:

```c++
                if (channel->Get_Pivot() >= m_numNodes) {
                    delete channel;
                }
```

src/w3d/renderer/motchan.cpp
xezon: `long` is 32 bit though.

duncanspumpkin: https://github.com/TheAssemblyArmada/Thyme/wiki/Thyme-Style-Guide also as per the style guide we should be avoiding it

xezon: I am ok with making it `uint32_t`

duncanspumpkin:
I'm in agreement with xezon. We should be using the sized types for this anyway which would be `uint32_t`/`int32_t` then it is completely unambiguous.

tomsons26 (changes requested):
revert the signage changes
HRawAnimClass::Get_Transform will actually need a deeper look, need to figure out where the best info source is for this one, but at a glance what you fix is actually as it should be, there's possibly code in wrong branches or out right missing in this function

xezon:
Ok. pivot index and type changed to unsigned again.

xezon:
Submit before this one:
* #547

xezon:
I tested various units and buildings and animations looked as usual. No problems found.

xezon:
Anything else that needs fixing here?

tomsons26 (approved):

PR #449, opened 2021-12-29
Add string type aliases

xezon:
See commits for individual changes.

### DIFF
--- a/src/game/common/system/asciistring.cpp
+++ b/src/game/common/system/asciistring.cpp
@@ -44,7 +44,7 @@ Utf8String::Utf8String(const char *s) : m_data(nullptr)
 {
     if (s != nullptr) {
         // Get length of the string that was passed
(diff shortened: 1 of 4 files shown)


jonwil (commented):
Why did you remove the GAME_DEBUG_STRUCTS bits? Are they not actually present in the original WB binary? (which is what GAME_DEBUG_STRUCTS is for)

tomsons26:
> Why did you remove the GAME_DEBUG_STRUCTS bits? Are they not actually present in the original WB binary? (which is what GAME_DEBUG_STRUCTS is for)

they are and WB will crash without them

tomsons26 (dismissed):
Restore debug code

xezon:
Restored GAME_DEBUG stuff.

OmniBlade (approved):

PR #451, opened 2021-12-31
Clean up Coord2D/Coord3D/ICoord2D/ICoord3D/Region2D/Region3D/IRegion2D/IRegion3D

jonwil:


src/w3d/lib/coord.h
```diff
@@ -155,91 +117,19 @@ class Coord3D
         z = 0.0f;
     }
 
-    Coord3D &operator+=(const Coord3D &rhs)
```
duncanspumpkin, line 158: Revert all of these. Overloaded operators are much much simpler to understand and harder to mess up

src/w3d/lib/coord.h
```diff
@@ -22,104 +22,66 @@ using GameMath::Sqrt;
...
 
-        return lhs;
+        if (y >= 0.0f) {
+            return GameMath::Acos(f);
+        } else {
+            return -GameMath::Acos(f);
+        }
     }
 
-    friend Coord2D operator-(Coord2D lhs, const Coord2D &rhs)
```
duncanspumpkin, line 68: Revert

duncanspumpkin:
This appears to remove all of the operator overloads? Why would you remove the improvements?

src/w3d/lib/coord.h
```diff
@@ -22,104 +22,66 @@ using GameMath::Sqrt;
 class Coord2D
 {
 public:
-    Coord2D() : x(0.0f), y(0.0f) {}
```
duncanspumpkin, line 25: Revert this is much safer

src/w3d/lib/coord.h
```diff
@@ -22,104 +22,66 @@ using GameMath::Sqrt;
...
-    Coord2D(float x_val, float y_val) : x(x_val), y(y_val) {}
```
duncanspumpkin, line 26: Revert this is much cleaner and more performant

src/w3d/lib/coord.h
```diff
@@ -22,104 +22,66 @@ using GameMath::Sqrt;
...
-
-public:
     float x;
     float y;
 };
 
 class Coord3D
 {
 public:
-    Coord3D() : x(0.0f), y(0.0f), z(0.0f) {}
```
duncanspumpkin, line 113: Revert these are much safer

src/game/client/system/particlesystem/particle.cpp
```diff
@@ -185,7 +187,9 @@ void Particle::Compute_Color_Rate()
  */
 void Particle::Apply_Force(const Coord3D &force)
 {
-    m_accel += force;
+    m_accel.x += force.x;
+    m_accel.y += force.y;
+    m_accel.z += force.z;
```
xezon: I agree with Duncan on the operators. This specific aspect does not look like a worthwhile change, because 
```
    m_accel.x += force.x;
    m_accel.y += force.y;
    m_accel.z += force.z;
```
is a lot more to read than
```
   m_accel += force;
```
Code should be simple.

src/w3d/lib/coord.h
xezon: What is the reasoning here to remove the default constructor? Is Coord2D meant to be an aggregate type? I doubt this is for performance. Keep in mind omitting default initialization in structs always carries the risk of non initialized values and subtle bugs.

On the other hand Vector3 & co also have no default constructor, so this seems consistent.

I would like this and similar classes to have a constructor that takes a ZERO enum and then initializes to zero. This way you can easily create object like this:

`Coord3D coord(ZERO);`

src/w3d/lib/coord.h
OmniBlade: My feeling here after being bitten by a but caused by default initialisation to 0.0 being added is that leave the default constructor as uninitialised and keep the value initialising constructor so the caller can decide to intialise it to 0.0 or any other value at their discretion.

src/game/client/system/particlesystem/particlesys.cpp
```diff
@@ -65,6 +64,12 @@ ParticleSystem::ParticleSystem(const ParticleSystemTemplate *temp, ParticleSyste
     m_saveable(true),
     m_unkBool1(false)
 {
+    m_lastPos.Zero();
+    m_pos.Zero();
+    m_velCoefficient.Zero();
+    m_velCoefficient.x = 1.0;
```
OmniBlade: Use float literal 1.0f for float assignment.

OmniBlade (changes requested):

OmniBlade:
Tough call on the constructors, it does seem like its intentional that they are not defined, the user of the class was expected to initialise them as required. @xezon made a good point that its possible this was done for some percieved performance reason. Adding defaults where they didn't exist caused issues previously so I'd be inclined to go with how it looks they were coded in the original.

src/game/client/system/particlesystem/particlesys.cpp
jonwil: Fixed that.

src/w3d/lib/coord.h
jonwil: I disagree with bringing back constructors in this class and think initialization should be explicit.

src/w3d/lib/coord.h
jonwil: I disagree with bringing back constructors in this class and think initialization should be explicit.

src/w3d/lib/coord.h
jonwil: I disagree with bringing back constructors in this class and think initialization should be explicit.

src/w3d/lib/coord.h
jonwil: I added operators based on what Vector2 and Vector3 have and used them where appropriate.

src/w3d/lib/coord.h
jonwil: I added operators based on what Vector2 and Vector3 have and used them where appropriate.

src/game/client/system/particlesystem/particle.cpp
jonwil: I added operators based on what Vector2 and Vector3 have and used them where appropriate.

src/game/client/system/particlesystem/particlesys.cpp
```diff
@@ -65,6 +64,12 @@ ParticleSystem::ParticleSystem(const ParticleSystemTemplate *temp, ParticleSyste
     m_saveable(true),
     m_unkBool1(false)
 {
+    m_lastPos.Zero();
+    m_pos.Zero();
+    m_velCoefficient.Zero();
```
xezon: This Zero() appears unnecessary if x,y,z are changed to 1 one line later.

src/game/client/system/particlesystem/particlesys.cpp
```diff
@@ -413,63 +420,48 @@ Coord3D *ParticleSystem::Compute_Particle_Velocity(const Coord3D *pos)
         case EMISSION_VELOCITY_OUTWARD: {
             float speed = m_emissionVelocity.outward.outward.Get_Value();
             float other_speed = m_emissionVelocity.outward.other.Get_Value();
+            Coord3D zero;
```
xezon, line 422: zero.Zero() possible here?

src/game/logic/scriptengine/scriptparam.cpp
```diff
@@ -19,8 +19,11 @@
 #include <captainslog.h>
 
 Parameter::Parameter(ParameterType type) :
-    m_type(type), m_initialized(false), m_int(0), m_real(0.0f), m_string(), m_coord{ 0.0f, 0.0f, 0.0f }, m_objStatus()
+    m_type(type), m_initialized(false), m_int(0), m_real(0.0f), m_string(), m_objStatus()
 {
+    m_coord.x = 0.0f;
+    m_coord.y = 0.0f;
+    m_coord.z = 0.0f;
```
xezon: m_coord.Zero() ?

src/game/logic/scriptengine/scriptparam.cpp
```diff
@@ -361,6 +364,9 @@ Utf8String Parameter::Get_UI_Text()
 void Parameter::Get_Coord3D(Coord3D *coord)
 {
     captainslog_dbgassert(m_type == COORD3D, "Parameter is not a Coord3D.");
+    coord->x = 0.0f;
+    coord->y = 0.0f;
+    coord->z = 0.0f;
```
xezon: coord->z.Zero() ?

src/w3d/lib/coord.h
```diff
@@ -263,20 +304,42 @@ class ICoord3D
...
+
+    float Height() { return hi.y - lo.y; }
+    float Width() { return hi.x - lo.x; }
+    float Depth() { return hi.z - lo.z; }
+    bool Is_In_Region_No_Z(Coord3D *c) { return lo.x < c->x && c->x < hi.x && lo.y < c->y && c->y < hi.y; }
+
+    bool Is_In_Region_With_Z(Coord3D *c)
+    {
+        return lo.x < c->x && c->x < hi.x && lo.y < c->y && c->y < hi.y && lo.z < c->z && c->z < hi.z;
+    }
```
xezon, line 341: 5 const functions here.

src/w3d/lib/coord.h
```diff
@@ -263,20 +304,42 @@ class ICoord3D
...
+
     Coord2D lo;
     Coord2D hi;
 };
 
 class IRegion2D
 {
 public:
+    int Height() { return hi.y - lo.y; }
+    int Width() { return hi.x - lo.x; }
```
xezon: const

src/w3d/lib/coord.h
```diff
@@ -263,20 +304,42 @@ class ICoord3D
 class Region2D
 {
 public:
+    float Height() { return hi.y - lo.y; }
+    float Width() { return hi.x - lo.x; }
```
xezon: const

src/w3d/lib/coord.h
```diff
@@ -238,23 +233,69 @@ class Coord3D
     }
 
     bool operator==(const Coord3D &that) { return x == that.x && y == that.y && z == that.z; }
+    bool operator!=(const Coord3D &that) { return x != that.x && y != that.y && z != that.z; }
```
xezon: const

src/w3d/lib/coord.h
```diff
@@ -102,24 +88,68 @@ class Coord2D
         return *this;
     }
 
-public:
+    bool operator==(const Coord2D &that) { return x == that.x && y == that.y; }
+    bool operator!=(const Coord2D &that) { return x != that.x && y != that.y; }
```
xezon: const

src/w3d/lib/coord.h
```diff
@@ -22,75 +22,61 @@ using GameMath::Sqrt;
...
 
-        return *this;
+        if (len != 0.0f) {
+            x /= len;
+            y /= len;
+        }
     }
 
-    Coord2D &operator-=(const Coord2D &rhs)
+    float To_Angle()
```
xezon: const

src/w3d/lib/coord.h
jonwil: fixed.

src/w3d/lib/coord.h
jonwil: fixed.

src/w3d/lib/coord.h
jonwil: fixed.

src/w3d/lib/coord.h
jonwil: fixed.

src/w3d/lib/coord.h
jonwil: fixed.

src/w3d/lib/coord.h
jonwil: fixed.

src/game/client/system/particlesystem/particlesys.cpp
jonwil: fixed.

src/game/logic/scriptengine/scriptparam.cpp
jonwil: fixed.

src/game/client/system/particlesystem/particlesys.cpp
jonwil: fixed.

src/game/logic/scriptengine/scriptparam.cpp
jonwil: fixed.

xezon (approved):

OmniBlade (approved):

PR #452, opened 2021-12-31
Minor refactoring in Win32LocalFile class

xezon:


src/platform/win32localfile.cpp
```diff
@@ -189,18 +189,14 @@ bool Win32LocalFile::Scan_Int(int &integer)
...
-
-        if (bytes == 0) {
+        if (read(m_fileHandle, &tmp, sizeof(tmp)) == 0) {
             break;
         }
 
         if (!isdigit(tmp)) {
             // If we read a byte, seek back that byte for the next read
             // as we are done with the current number.
-            if (bytes) {
```
xezon, line 201: I removed this `bytes` test here, because just a few lines earlier it already tests `bytes`. And then simplified the above condition to less lines accordingly, because the `bytes` variable is no longer needed. The same applies to 2 more similar functions below.

jonwil (approved):
Looks good to me. Approved.

duncanspumpkin (approved):

PR #453, opened 2021-12-31
Implement StandardFile with buffered IO to allow speed up File operations where necessary

### LINKED ISSUE #304: Add support for buffered file write outs for File and Win32LocalFile
Every single file write issues a prompt write out to disk. Add an optional file write out mode that allows to hand over or allocate a write out buffer so that file writes are performed in chunks and not on each bytes written. Theoretically systems could implement their own write buffers when interacting with `File`, but it is more versatile and robust if `File` has support for such functionality natively.

This is a feature.

xezon:
`Win32BufferedFile` is a copy of `Win32LocalFile` for the most part. Instead of read/write it uses the more sophisticated fread/fwrite. This allows it to make use of the buffered file IO. Custom buffer size can be controlled by encoding size value into `filemode` integer.

Usage:

before:
```c++
File *file = g_theFileSystem->Open("myfile.aaa", File::READ | File::BINARY);
```

after:
```c++
File *file = g_theFileSystem->Open("myfile.aaa", File::READ | File::BINARY | File::BUFFERED);
```

with encoded custom buffer size (here: 1024)
```c++
File *file = g_theFileSystem->Open("myfile.aaa", Thyme::Encode_Buffered_File_Mode(File::READ | File::BINARY, 1024));
```

Closes #304

### DIFF
--- a/src/game/common/system/file.h
+++ b/src/game/common/system/file.h
@@ -26,6 +26,18 @@ struct FileInfo
     int write_time_low;
 };
 
+namespace Thyme
+{
+// Thyme specific
+
+// Returns integer of given mode and BUFFERED flag and encoded buffer size.
+// Buffer size min: 0, max: 4194240, in 64 byte increments.
+int Encode_Buffered_File_Mode(int mode, int buffer_size);
+
+// Decodes buffer size from given mode.
+bool Decode_Buffered_File_Mode(int mode, int &buffer_size);
+} // namespace Thyme
+
 class File : public MemoryPoolObject
 {
     IMPLEMENT_ABSTRACT_POOL(File);
@@ -48,6 +60,11 @@ public:
         TEXT = 0x20,
         BINARY = 0x40,
         STREAMING = 0x100,
+        BUFFERED = 0x8000, // Thyme specific. When set, opens file with buffered read & write as opposed to instant file IO.
+                           // Optionally, the high 16 bits of FileMode can be used to specify the buffer size, otherwise uses
+                           // default buffer size.
+
+        LASTMODE, // Just marks the enum end.
     };
 
 protected:
--- a/src/game/common/system/gamememoryinit.cpp
+++ b/src/game/common/system/gamememoryinit.cpp
@@ -648,6 +648,7 @@ static PoolSizeRec UserMemoryPools[] = {
     { "ThumbnailManagerClass", 32, 32 },
     { "SmudgeSet", 32, 32 },
     { "Smudge", 128, 32 },
+    { "StandardFile", 32, 32 }, // Thyme specific.
     { nullptr, 0, 0 } // Last entry always null.
 };
 
--- a/src/platform/win32localfilesystem.cpp
+++ b/src/platform/win32localfilesystem.cpp
@@ -13,6 +13,7 @@
  *            LICENSE
  */
 #include "win32localfilesystem.h"
+#include "standardfile.h"
 #include "win32localfile.h"
 
 // Headers needed for posix open, close, read... etc.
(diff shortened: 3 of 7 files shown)


src/platform/win32bufferedfile.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+    // r    open for reading (The file must exist)
+    // w    open for writing (creates file if it doesn't exist). Deletes content and overwrites the file.
+    // a    open for appending (creates file if it doesn't exist).
+    // r+   open for reading and writing (The file must exist).
+    // w+   open for reading and writing.
+    //      If file exists deletes content and overwrites the file, otherwise creates an empty new file.
+    // a+   open for reading and writing (append if file exists).
+
+    Utf8String modestr;
```
xezon: The `fopen` function takes a string to control the file mode as opposed to flags. Below is my attempt to build this string based on the passed flags. I am not 100% sure I got it all right, but this is catered to the documentation above. It should be close enough. If something is off, we can tweak later or fix here in review.

src/platform/win32bufferedfile.cpp
```diff
@@ -0,0 +1,290 @@
...
+            break;
+        }
+    }
+
+    integer = atoi(number.Str());
+
+    return true;
+}
+
+bool Win32BufferedFile::Scan_Real(float &real)
```
xezon: I copied all functions from `Win32LocalFile` as a base, so the algorithms are an exact match.

src/platform/win32localfilesystem.cpp
```diff
@@ -29,11 +30,19 @@
 
 File *Win32LocalFileSystem::Open_File(const char *filename, int mode)
 {
-    if (strlen(filename) <= 0) {
+    using Win32BufferedFile = Thyme::Win32BufferedFile;
+
+    if (filename == nullptr || *filename == '\0') {
```
xezon, line 33: I was so free and simplified this check here. There is no need to run an strlen if it just needs to know if first character is set. Also checks for null now as well.

src/platform/win32localfilesystem.cpp
```diff
@@ -29,11 +30,19 @@
...
     }
 
-    Win32LocalFile *file = NEW_POOL_OBJ(Win32LocalFile);
+    File *file;
+
+    if (mode & File::BUFFERED) {
+        file = NEW_POOL_OBJ(Win32BufferedFile);
+    } else {
+        file = NEW_POOL_OBJ(Win32LocalFile);
+    }
```
xezon, line 43: Only uses `Win32BufferedFile` when the flag is set, otherwise uses legacy `Win32LocalFile`.

jonwil (approved):
Looks good to me. Approved.

src/game/common/system/gamememoryinit.cpp
```diff
@@ -648,6 +648,7 @@ static PoolSizeRec UserMemoryPools[] = {
     { "ThumbnailManagerClass", 32, 32 },
     { "SmudgeSet", 32, 32 },
     { "Smudge", 128, 32 },
+    { "Win32BufferedFile", 32, 32 }, // Thyme specific.
```
OmniBlade: Can we call this something other than Win32BufferedFile. I know the original implementation is called Win32File and I only left it called that because that is what it was originally called despite it using the posix API. StandardBufferedFile or even just StandardFile would be okay given that its using stdio.h functions.

src/game/common/system/file.h
```diff
@@ -26,6 +26,18 @@ struct FileInfo
     int write_time_low;
 };
 
+namespace Thyme
```
OmniBlade, line 29: Do we really need this level of control over the buffering? Can't we just flag filesystem to provide us with a buffered file and accept the standard buffer and leave the File API untouched?

OmniBlade (changes requested):

src/game/common/system/file.h
xezon: Yes. Filesystem user can also just set the BUFFERED flag and then it will use the standard buffer size (512 bytes). The use of `Encode_Buffered_File_Mode` is optional (I use it in `GameTextFile` class for 64 kb size).

src/game/common/system/gamememoryinit.cpp
xezon: I have chosen `Win32BufferedFile` so it is very obvious that it is similar to `Win32LocalFile`. I can change name if you really desire so.

xezon: Renamed to `StandardFile`

OmniBlade (approved):

PR #454, opened 2022-01-01
Fix and refactor GameTextManager class

xezon:
A couple of minor changes to GameTextManager class. Split into 5 commits for ease of review.


jonwil (approved):
Looks good to me.

PR #455, opened 2022-01-03
Remove rts::equal_to and use std::equal_to. Switch rts::hash impl

duncanspumpkin:
`std::equal_to` is identical to `rts::equal_to` so I've removed it. Utf8String has an `==operator` that uses `Compare`.
I've swapped around the hash impl so that it can be used with any enums (used in #307).

### DIFF
--- a/src/game/rtsutils.h
+++ b/src/game/rtsutils.h
@@ -15,6 +15,7 @@
 #pragma once
 
 #include "always.h"
+#include "asciistring.h"
 #include "endiantype.h"
 #include <ctime>
 
@@ -58,14 +59,14 @@ template<typename T> struct less_than_nocase
     bool operator()(const T &left, const T &right) const { return (left.Compare_No_Case(right) < 0); }
 };
 
-template<typename T> struct equal_to
-{
-    bool operator()(const T &left, const T &right) const { return (left.Compare(right) == 0); }
-};
-
 template<typename T> struct hash
 {
-    size_t operator()(const T &object) const
+    size_t operator()(const T &object) const { return static_cast<size_t>(object); }
+};
+
+template<> struct hash<Utf8String>
+{
+    size_t operator()(const Utf8String &object) const
     {
         const char *c = reinterpret_cast<const char *>(object.Str());
         size_t hash = 0;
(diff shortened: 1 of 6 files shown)


OmniBlade (approved):

src/game/rtsutils.h
```diff
@@ -58,14 +59,14 @@ template<typename T> struct less_than_nocase
...
 {
-    bool operator()(const T &left, const T &right) const { return (left.Compare(right) == 0); }
+    size_t operator()(const T &object) const { return static_cast<size_t>(object); }
 };
 
-template<typename T> struct hash
+template<> struct hash<Utf8String>
 {
-    size_t operator()(const T &object) const
+    size_t operator()(const Utf8String &object) const
```
xezon, line 69:
```c++
        do {
            hash = *c + 5 * hash;
            ++c;
        } while (*c != '\0');
```
This looks wrong. This will read string past the end if the first character is null. This can in return cause looping well beyond the buffer and invoking access violation (crash).

The following makes more sense:
```c++
        while (*c != '\0') {
            hash = *c + 5 * hash;
            ++c;
        };
```

src/game/rtsutils.h
```diff
@@ -58,14 +59,14 @@ template<typename T> struct less_than_nocase
     bool operator()(const T &left, const T &right) const { return (left.Compare_No_Case(right) < 0); }
 };
 
-template<typename T> struct equal_to
+template<typename T> struct hash
 {
-    bool operator()(const T &left, const T &right) const { return (left.Compare(right) == 0); }
+    size_t operator()(const T &object) const { return static_cast<size_t>(object); }
```
xezon, line 64: Is this confirmed what original does?

xezon (dismissed):

src/game/rtsutils.h
duncanspumpkin: While true that isn't what I was looking at as part of the pr tbh i would rather it just called the std library https://github.com/TheAssemblyArmada/Thyme/blob/f95ea5c66e3ecd31930f36883764f14d8d287cff/deps/stlport/stl/_hash_fun.h#L42

xezon: We can do follow up change to tackle this problem indeed.

src/game/rtsutils.h
duncanspumpkin: It copies the stl version. This class is just a duplicate of the stl version but modified so that it can take a utf8string

xezon: Ok that is fine then.

xezon (approved):

PR #456, opened 2022-01-03
Fix minor errors in Render2DSentenceClass::Build_Sentence_Centered

### LINKED ISSUE #280: Tool tip text renders with corrupt characters
#279 is possibly related to this issue too but not sure, could be a different branch of the code.
This doesn't happen after unhooking FontCharsClass and Render2DSentenceClass
game dat_W4AcHA6Wyg
Game resolution was 1024x768

xezon:
Fixes #280

### DIFF
--- a/src/w3d/renderer/render2dsentence.cpp
+++ b/src/w3d/renderer/render2dsentence.cpp
@@ -752,7 +752,7 @@ void Render2DSentenceClass::Build_Sentence_Centered(const unichar_t *text, int *
         Allocate_New_Surface(text, false);
     }
 
-    bool breakout = false;
+    bool sentence_done = false;
     bool ampersand_handled = false;
 
     while (true) {
@@ -771,6 +771,10 @@ void Render2DSentenceClass::Build_Sentence_Centered(const unichar_t *text, int *
                     break;
                 }
 
+                if (test_ch == U_CHAR('\n')) {
+                    break;
+                }
+
                 if (m_processAmpersand && test_ch == U_CHAR('&')) {
                     // Spaces before the ampersand also get disregarded?
                     if (word_width != 0 && cur_text[-1] == U_CHAR(' ')) {
@@ -781,7 +785,8 @@ void Render2DSentenceClass::Build_Sentence_Centered(const unichar_t *text, int *
                     ampersand_handled = true;
                 }
 
-                unichar_t cur_ch = *cur_text++;
+                unichar_t cur_ch = *cur_text;
+                ++cur_text;
                 spacing = m_font->Get_Char_Spacing(cur_ch);
                 word_width += spacing;
                 ++word_count;
@@ -802,7 +807,7 @@ void Render2DSentenceClass::Build_Sentence_Centered(const unichar_t *text, int *
                         sentence_width += word_width - spacing;
 
                         if (*cur_text == U_CHAR('\0')) {
-                            breakout = true;
+                            sentence_done = true;
                         }
                     } else {
                         --word_count_tracker;
@@ -814,7 +819,7 @@ void Render2DSentenceClass::Build_Sentence_Centered(const unichar_t *text, int *
             if (*cur_text == U_CHAR('\0')) {
                 word_count_tracker += word_count;
                 sentence_width += word_width;
-                breakout = true;
+                sentence_done = true;
                 break;
             }
 
@@ -831,7 +836,7 @@ void Render2DSentenceClass::Build_Sentence_Centered(const unichar_t *text, int *
             sentence_width += word_width;
         }
 
-        m_cursor.X = std::max((position.X - sentence_width) * 0.5f, 0.0f);
+        m_cursor.X = (int)std::max((position.X - sentence_width) * 0.5f, 0.0f);
 
         if (ampersand_handled) {
             ampersand_handled = false;
@@ -845,7 +850,7 @@ void Render2DSentenceClass::Build_Sentence_Centered(const unichar_t *text, int *
 
             if (m_processAmpersand && cur_ch == U_CHAR('&')) {
                 unichar_t next_ch = *text;
-                if (next_ch != U_CHAR('\0') && next_ch > U_CHAR(' ')) {
+                if (next_ch != U_CHAR('\0') && next_ch > U_CHAR(' ') && next_ch != U_CHAR('\n')) {
                     cur_ch = next_ch;
                     ++text;
                     ampersand_processed = true;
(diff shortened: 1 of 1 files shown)


src/w3d/renderer/render2dsentence.cpp
```diff
@@ -845,7 +853,7 @@ void Render2DSentenceClass::Build_Sentence_Centered(const unichar_t *text, int *
 
             if (m_processAmpersand && cur_ch == U_CHAR('&')) {
                 unichar_t next_ch = *text;
-                if (next_ch != U_CHAR('\0') && next_ch > U_CHAR(' ')) {
+                if (next_ch != U_CHAR('\0') && next_ch > U_CHAR(' ') && next_ch != U_CHAR('\n')) {
```
xezon: Looks like `next_ch != U_CHAR('\n')` can be omitted because of `next_ch > U_CHAR(' ')`. It's odd this was compiled in though. Hmm.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -771,6 +771,10 @@ void Render2DSentenceClass::Build_Sentence_Centered(const unichar_t *text, int *
                     break;
                 }
 
+                if (test_ch == U_CHAR('\n')) {
```
xezon: This looks like it can be omitted because of the test above. It's odd it was compiled in though.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -831,7 +836,10 @@ void Render2DSentenceClass::Build_Sentence_Centered(const unichar_t *text, int *
             sentence_width += word_width;
         }
 
-        m_cursor.X = std::max((position.X - sentence_width) * 0.5f, 0.0f);
+        m_cursor.X = (float)(int)((position.X - sentence_width) * 0.5f);
+        if (m_cursor.X < 0.0f) {
+            m_cursor.X = 0.0f;
+        }
```
xezon: This change fixes the screwed character positions.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -781,7 +785,8 @@ void Render2DSentenceClass::Build_Sentence_Centered(const unichar_t *text, int *
                     ampersand_handled = true;
                 }
 
-                unichar_t cur_ch = *cur_text++;
+                unichar_t cur_ch = *cur_text;
+                ++cur_text;
```
duncanspumpkin, line 785: This is identical. Why change?

src/w3d/renderer/render2dsentence.cpp
duncanspumpkin: You could still use max but make it cast to int to force the truncation

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -862,7 +870,7 @@ void Render2DSentenceClass::Build_Sentence_Centered(const unichar_t *text, int *
 
                 if (cur_ch == U_CHAR(' ')) {
                     m_cursor.X += char_spacing;
-                } else if (cur_ch == U_CHAR('\n') || cur_ch == U_CHAR('\0')) {
+                } else if (cur_ch == U_CHAR('\0') || cur_ch == U_CHAR('\n')) {
```
duncanspumpkin, line 866: Why has this been swapped

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -1005,7 +1012,8 @@ Vector2 Render2DSentenceClass::Build_Sentence_Not_Centered(const unichar_t *text
                                 ++tmp_text;
                             }
 
-                            spacing += m_font->Get_Char_Spacing(*tmp_text++);
+                            spacing += m_font->Get_Char_Spacing(*tmp_text);
+                            ++tmp_text;
```
duncanspumpkin, line 1011: Why this change its identical

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -916,10 +924,7 @@ void Render2DSentenceClass::Build_Sentence_Centered(const unichar_t *text, int *
 
     if (x != nullptr) {
         *x = final_x_pos;
-
-        if (y != nullptr) {
-            *y = 0;
-        }
+        *y = 0;
```
duncanspumpkin, line 922: Revert, should keep the safety check

src/w3d/renderer/render2dsentence.cpp
xezon: It's easier on the eye in this case and compares with the decompile better.

src/w3d/renderer/render2dsentence.cpp
xezon: In the decompile it is in this order and therefore easier to compare against.

src/w3d/renderer/render2dsentence.cpp
xezon: It's easier on the eye in this case and compares with the decompile better.

src/w3d/renderer/render2dsentence.cpp
xezon: No. It should be different.

src/w3d/renderer/render2dsentence.cpp
duncanspumpkin: It will compile to exactly the same and is the normal convention for this sort of function

src/w3d/renderer/render2dsentence.cpp
xezon: Yeah, but this compares better with the decompile, so can leave as is. Important is that it works correctly.

src/w3d/renderer/render2dsentence.cpp
duncanspumpkin: Why? Do you have some reason why `y` can never be nullptr?

src/w3d/renderer/render2dsentence.cpp
duncanspumpkin: It is much harder to read and reason with when you add unnecessary casts at least drop the float cast there is no need to do that twice.

src/w3d/renderer/render2dsentence.cpp
xezon: Fixed.

src/w3d/renderer/render2dsentence.cpp
xezon: Fixed.

src/w3d/renderer/render2dsentence.cpp
xezon: I added proper checks.

src/w3d/renderer/render2dsentence.cpp
xezon: Ok. Changed and tested.

PR #458, opened 2022-01-04
Implement SightingInfo

duncanspumpkin:
Not much to say about this class. All is pretty self explanatory.

### DIFF
--- a/src/game/logic/system/partitionmanager.h
+++ b/src/game/logic/system/partitionmanager.h
@@ -18,7 +18,6 @@
 #include "mempoolobj.h"
 #include "snapshot.h"
 #include "subsysteminterface.h"
-#include <new>
 #include <queue>
 
 struct FindPositionOptions;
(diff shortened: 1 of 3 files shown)


src/game/logic/system/partitionmanager.h
```diff
@@ -119,10 +119,28 @@ class SightingInfo : public MemoryPoolObject, public SnapShot
 {
     IMPLEMENT_POOL(SightingInfo);
 
+    void *operator new(size_t size, void *dst) { return dst; }
+    void operator delete(void *p, void *q) {}
```
xezon: Is this proper? Functions don't do anything.

duncanspumpkin: its just so that i can hook it

tomsons26: use `#include <new>`

src/game/logic/system/partitionmanager.h
```diff
@@ -119,10 +120,28 @@ class SightingInfo : public MemoryPoolObject, public SnapShot
 {
     IMPLEMENT_POOL(SightingInfo);
 
+    void *operator new(size_t size, void *dst) { return dst; }
```
jonwil: Remove these, these are not needed with the #include <new> in there.

jonwil (dismissed):

src/game/logic/system/partitionmanager.h
duncanspumpkin: Seems to be needed still. Getting a compile error otherwise. Tbh i don't think we need to hook the ctor I might just remove that.

tomsons26 (approved):

PR #462, opened 2022-01-05
Fix crash in LaserUpdate::Update_Start_Pos

xezon:
Not 100% this is correct as I have not seen code, but logically this looks right to begin with.

### DIFF
--- a/src/game/logic/object/update/laserupdate.cpp
+++ b/src/game/logic/object/update/laserupdate.cpp
@@ -310,13 +310,13 @@ void LaserUpdate::Update_Start_Pos()
                 m_startPos.x = m.Get_X_Translation();
                 m_startPos.y = m.Get_Y_Translation();
                 m_startPos.z = m.Get_Z_Translation();
+            } else {
+                m_startPos = *drawable->Get_Position();
             }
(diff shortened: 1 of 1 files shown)


src/game/logic/object/update/laserupdate.cpp
```diff
@@ -310,9 +310,9 @@ void LaserUpdate::Update_Start_Pos()
...
                 m_startPos.y = m.Get_Y_Translation();
                 m_startPos.z = m.Get_Z_Translation();
+            } else {
+                m_startPos = *drawable->Get_Position();
             }
-        } else {
-            m_startPos = *drawable->Get_Position();
         }
 
         if (!(m_startPos == oldpos)) {
```
jonwil: Move this if statement inside the if (drawable) as well. Looks good otherwise.

jonwil (changes requested):

src/game/logic/object/update/laserupdate.cpp
xezon: ok

jonwil (approved):
Looks good to me now, approved.

PR #463, opened 2022-01-06
Partially implements GameWindow.

tomsons26:


src/game/client/gui/gamewindow.h
```diff
@@ -0,0 +1,149 @@
...
+enum WindowMsgHandledType
+{
+    MSG_IGNORED,
+    MSG_HANDLED,
+};
+
+typedef WindowMsgHandledType(__cdecl *window_callback_func)(
+    GameWindow *window, unsigned int message, unsigned int data_1, unsigned int data_2);
+
+typedef void(__cdecl *window_draw_func)(GameWindow *window, WinInstanceData *instance);
```
duncanspumpkin: I don't think there is an entry in the style guide for function pointers but I would expect them to follow either the casing of classes or the casing of functions and not the style of variables.

src/game/client/gui/gamewindow.cpp
```diff
@@ -0,0 +1,407 @@
...
+    if (m_inputData != nullptr) {
+        delete m_inputData;
+    }
+    m_inputData = nullptr;
+
+    if (m_editData != nullptr) {
+        delete m_editData;
+    }
+
+    m_editData = nullptr;
```
xezon: Destructor can be simplified to just
```
delete m_inputData;
delete m_editData;
```

src/game/client/gui/gamewindow.cpp
```diff
@@ -0,0 +1,407 @@
...
+}
+
+void GameWindow::Normalize_Window_Region()
+{
+    if (m_region.lo.x > m_region.hi.x) {
+        int tmp = m_region.lo.x;
+        m_region.lo.x = m_region.hi.x;
```
xezon: Can be simplified to `std::swap(m_region.lo.x, m_region.hi.x)`

src/game/client/gui/gamewindow.cpp
```diff
@@ -0,0 +1,407 @@
...
+        m_region.hi.x = tmp;
+    }
+
+    if (m_region.lo.y > m_region.hi.y) {
+        int tmp = m_region.lo.y;
+        m_region.lo.y = m_region.hi.y;
```
xezon: Can be simplified to `std::swap(m_region.lo.y, m_region.hi.y)`

src/game/client/gui/gamewindow.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+    Normalize_Window_Region();
+
+    return 0;
+}
+
+int GameWindow::Win_Get_Position(int *x, int *y)
+{
+    if (!x || !y) {
+        return -3;
```
xezon, line 81: What does this mysterious -3 mean?

src/game/client/gui/gamewindow.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+    return 0;
+}
+
+int GameWindow::Win_Get_Screen_Position(int *x, int *y)
+{
+    GameWindow *parent = m_parent;
+
+    *x = m_region.lo.x;
+    *y = m_region.lo.y;
```
xezon, line 101: In above function, `x` and `y` are tested, but here not. Is this proper?

src/game/client/gui/gamewindow.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+    *width = m_size.x;
+    *height = m_size.y;
+
+    return 0;
+}
+
+int GameWindow::Win_Enable(bool enable)
+{
+    int status;
```
xezon: This temporary can be removed. Just work on `m_status` directly.
```c++
    if (enable) {
        m_status |= WIN_STATUS_ENABLED;
    } else {
        m_status &= ~WIN_STATUS_ENABLED;
    }
```

src/game/client/gui/gamewindow.cpp
```diff
@@ -0,0 +1,407 @@
...
+        int r_x = i->m_region.lo.x;
+        int r_y = i->m_region.lo.y;
+
+        for (GameWindow *j = i->m_parent; j != nullptr; j = j->m_parent) {
+            r_x += j->m_region.lo.x;
+            r_y += j->m_region.lo.y;
+        }
+
+        if (x >= r_x && x <= i->m_size.x + r_x && y >= r_y && y <= i->m_size.y + r_y
+            && (ignore_hidden != 1 || !(i->m_status & WIN_STATUS_HIDDEN))) {
```
xezon: `!ignore_hidden` or `ignore_hidden == false`

src/game/client/gui/gamewindow.cpp
```diff
@@ -0,0 +1,407 @@
...
+        }
+    }
+
+    return this;
+}
+
+int GameWindow::Win_Set_Enabled_Image(int index, Image *image)
+{
+    if (index < 0 || index >= WinInstanceData::MAX_DRAW_DATA) {
+        captainslog_debug("Set enabled image, index out of range '%d'\n", index);
```
xezon: `\n` char at the end can be omitted in all logs.

src/game/client/gui/gamewindow.cpp
```diff
@@ -0,0 +1,407 @@
...
+int GameWindow::Win_Set_Tooltip_Func(window_tooltip_func tooltip)
+{
+    m_tooltip = tooltip;
+
+    return 0;
+}
+
+GameWindow *GameWindow::Win_Point_In_Child(int x, int y, bool ignore_enable_check, bool audio_event)
+{
+    for (GameWindow *i = m_child; i != nullptr; i = i->m_next) {
```
xezon: i = window

src/game/client/gui/gamewindow.cpp
```diff
@@ -0,0 +1,407 @@
...
+        int r_x = i->m_region.lo.x;
+        int r_y = i->m_region.lo.y;
+
+        for (GameWindow *j = i->Win_Get_Parent(); j != nullptr; j = j->m_parent) {
```
xezon: j = parent_window

Same in next function.

src/game/client/gui/gamewindow.h
OmniBlade: I'm not sure if I've done any function pointers as of yet, I've probably done it as one of the type styles, callback_func_t or CallbackFunc. I might be inclined to drop the "func" when callback is also mentioned as well, callback implies it. So shall we go with WindowCallback and WindowDrawFunc?

PR #465, opened 2022-01-06
Improve game memory pool setup

xezon:
Pool object can now be created with new operator. Calling `NEW_POOL_OBJ` is no longer necessary.

Using `delete MyPoolObject` does not compile. Not sure why, but it's good as is.

The reason delete operator must be used with this setup, is the fact that the virtual table is deleted before code in delete function is run. Therefore virtual function to access Memory Pool cannot be called anymore.

See this test code:
https://godbolt.org/z/qWc9TEbze

Therefore only `Delete_Instance` should be called.

An `operator delete` still needs to be there because `operator new` requires the pairing. But given the `MagicEnum`, no one practically calls it by mistake.

### DIFF
--- a/src/game/common/system/mempoolobj.h
+++ b/src/game/common/system/mempoolobj.h
@@ -41,8 +41,8 @@ protected:
     virtual ~MemoryPoolObject() {}
     virtual MemoryPool *Get_Object_Pool() = 0;
 
-    void *operator new(size_t);
-    void operator delete(void *) { captainslog_debug("This should be impossible"); }
+    void *operator new(size_t) = delete;
+    void operator delete(void *) { captainslog_dbgassert(0, "This should be impossible to call"); }
 
     // Class implementing Get_Object_Pool needs to provide these
     // use macros below to generated them.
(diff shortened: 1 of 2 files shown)


src/game/common/system/mempoolobj.h
```diff
@@ -76,45 +76,44 @@ private: \
     virtual MemoryPool *Get_Object_Pool() override { return Get_Class_Pool(); } \
 \
 public: \
-    enum classname##MagicEnum{ classname##_GLUE_NOT_IMPLEMENTED }; \
+    enum classname##MagicEnum{ classname##_GLUE_NOT_IMPLEMENTED = 0 }; \
 \
-    void *operator new(size_t size); \
-    void *operator new(size_t size, classname##MagicEnum) \
+    void *operator new(size_t size, classname##MagicEnum = classname##_GLUE_NOT_IMPLEMENTED) \
```
xezon, line 81: `MagicEnum` argument now has default value here. This allows writing `new MyPoolObject`.

src/game/common/system/mempoolobj.h
```diff
@@ -76,45 +76,44 @@ private: \
...
 #define IMPLEMENT_ABSTRACT_POOL(classname) \
 private: \
     virtual MemoryPool *Get_Object_Pool() override { throw CODE_01; } \
-\
-    enum classname##MagicEnum{ classname##_GLUE_NOT_IMPLEMENTED }; \
-\
-protected: \
-    void *operator new(size_t size); \
-    void *operator new(size_t size, classname##MagicEnum); \
-    void operator delete(void *ptr) { captainslog_debug("this should be impossible to call (abstract base class)"); } \
```
xezon, line 107: I have seen this error message in the decompile, but it's useless to keep. Best to just delete the function and let the compiler give an error straight.

src/game/common/system/mempoolobj.h
```diff
@@ -76,45 +76,44 @@ private: \
...
             "The wrong operator new is being called; ensure all objects in the hierarchy have MemoryPoolGlue set up " \
             "correctly"); \
         return Get_Class_Pool()->Allocate_Block(); \
     } \
-    void operator delete(void *ptr) \
+    void operator delete(void *ptr) = delete; \
+    void operator delete(void *ptr, classname##MagicEnum) \
     { \
-        captainslog_debug("Please call deleteInstance instead of delete."); \
+        captainslog_dbgassert(0, "Please call Delete_Instance instead of delete."); \
```
xezon, line 95: I promoted this to an `assert`, because really it should be treated as an error.

xezon:
Somehow there is compile error in `ClientUpdateModule` now. Needs figuring out.

OmniBlade (approved):

src/game/logic/object/update/clientupdatemodule.h
```diff
@@ -19,9 +19,13 @@
 
 class ClientUpdateModule : public DrawableModule
 {
+    IMPLEMENT_ABSTRACT_POOL(ClientUpdateModule)
```
tomsons26, line 22: seems it was DrawableModule that had the pool

tomsons26 (changes requested):

tomsons26 (approved):

PR #468, opened 2022-01-07
Set correct size of ThingTemplate class

xezon:
IDA shows pack size of 2 with ThingTemplate class. In World Builder its size 4. No idea if correct. But thought I put this here for review.


jonwil (commented):
This PR is not necessary, I can confirm from the binaries (both WB and game) that packing of 4 is correct and the correct size of ThingTemplate (as confirmed by the "operator new" call in ThingFactory::newTemplate in both binaries) is 0x23C. IDA may have miss-detected the padding size.

PR #469, opened 2022-01-07
Fix ThingTemplate initialization order and missing members

xezon:
While debugging I saw constructor of ThingTemplate is wild. Ordering is not correct. Also a pod members had no value initialization. So I just set some reasonable ones for them.

### DIFF
--- a/src/game/common/thing/thingtemplate.cpp
+++ b/src/game/common/thing/thingtemplate.cpp
@@ -224,52 +224,77 @@ AudioArray &AudioArray::operator=(const AudioArray &that)
 
 ThingTemplate::ThingTemplate() :
     m_geometryInfo(GEOMETRY_SPHERE, false, 1.0f, 1.0f, 1.0f),
-    m_moduleParseState(0),
-    m_originalSkinTemplate(nullptr),
-    m_radarPriority(RADAR_PRIORITY_NOT_ON_RADAR),
-    m_nextThingTemplate(nullptr),
-    m_transportSlotCount(0),
-    m_fenceWidth(0.0f),
-    m_fenceXOffset(0.0f),
-    m_visionRange(0.0f),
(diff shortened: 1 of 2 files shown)


tomsons26 (changes requested):
Don't change order, it matches existing binaries, its easier to cross reference then.
for missing members do this at end
```
    // BUGFIX Intialize all members
    m_buildable(0),
    m_assetScale(1.0f),
    m_displayColor(0xFFFFFFFF),
    m_editorSorting(0),
    m_isForbidden(false),
    m_armorCopiedFromDefault(false),
    m_weaponsCopiedFromDefault(false),
    m_isBridge(false)
```
since we are messing about with these files in header add a new line after `unsigned char m_crushableLevel;` so there's a gap between static and members

duncanspumpkin:
In general it's more performant for the order to match the declaration order. Plus you get the benefit of making it easier to confirm if items are missing.

 I really think we are making it much harder on ourselves striving for similarity to the binary these latest bugs are further showing that written code is not being understood when written. Perhaps if people were focusing on understanding rather than trying to make identical this wouldn't occur.

xezon:
Technically the order of initialization will always match the declaration order. So it will behave the same. I disagree that comparability with Decompiled is a strong argument to leave it in the wrong order. It should follow CPP guidelines and be properly written in the correct order.

This makes
a) Easier to review declaration and initialization
b) Easier to debug, because the debug position will follow initialization order and not jump all over the place when stepping into constructor
c) Avoid compiler warnings due wrong initialization order on some compilers

duncanspumpkin:
> Technically the order of initialization will always match the declaration order. So it will behave the same.

Hmm the more you learn. I'm sure I read something that said the opposite of this but seems I was mistaken on looking into it again.

xezon:
Done.

tomsons26 (approved):

PR #470, opened 2022-01-07
Fix ModuleFactory::Find_Module_Interface_Mask function name

xezon:
Shows as **findModuleInterfaceMask** in Decompiler. So I adjusted name accordingly.

### DIFF
--- a/src/game/common/thing/thingtemplate.cpp
+++ b/src/game/common/thing/thingtemplate.cpp
@@ -474,9 +474,9 @@ void ThingTemplate::Init_For_LTA(const Utf8String &name)
     strncpy(buffer, name, ARRAY_SIZE(buffer));
     int i;
 
-    for (i = 0; buffer[i]; i++) {
+    for (i = 0; buffer[i]; ++i) {
         if (buffer[i] == '/') {
-            i++;
+            ++i;
             break;
         }
     }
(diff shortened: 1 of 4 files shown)


src/game/common/thing/thingtemplate.cpp
```diff
@@ -474,9 +474,9 @@ void ThingTemplate::Init_For_LTA(const Utf8String &name)
     strncpy(buffer, name, ARRAY_SIZE(buffer));
     int i;
 
-    for (i = 0; buffer[i]; i++) {
+    for (i = 0; buffer[i]; ++i) {
```
jonwil, line 477: Why make this unrelated change?

src/game/common/thing/thingtemplate.cpp
```diff
@@ -474,9 +474,9 @@ void ThingTemplate::Init_For_LTA(const Utf8String &name)
...
         if (buffer[i] == '/') {
-            i++;
+            ++i;
```
jonwil, line 479: Why make this unrelated change?

jonwil (changes requested):
Other than the unnecessary unrelated changes, this looks OK to me.

src/game/common/thing/thingtemplate.cpp
xezon: Decompiler shows like so. Plus Pre-increment is always the default that should be used unless post-increment is explicitly required.

jonwil (approved):
Looks good to me.

tomsons26 (approved):

PR #471, opened 2022-01-07
Minor tweaks in ThingFactory and ModuleFactory classes

xezon:
A few tweaks.

### DIFF
--- a/src/game/common/modules/modulefactory.cpp
+++ b/src/game/common/modules/modulefactory.cpp
@@ -79,6 +79,7 @@ const ModuleFactory::ModuleTemplate *ModuleFactory::Find_Module_Template(const U
         return &it->second;
     }
 
+    captainslog_dbgassert(0, "Module name '%s' not found", name.Str());
     return nullptr;
 }
 
@@ -130,7 +131,7 @@ ModuleData *ModuleFactory::New_Module_Data_From_INI(INI *ini, const Utf8String &
 
         if (temp != nullptr) {
             ModuleData *data = temp->create_data_proc(ini);
-            data->m_tagKey = g_theNameKeyGenerator->Name_To_Key(tag);
+            data->m_tagKey = g_theNameKeyGenerator->Name_To_Key(tag.Str());
             m_moduleDataList.push_back(data);
 
             return data;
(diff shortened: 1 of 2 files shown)


src/game/common/modules/modulefactory.cpp
```diff
@@ -79,6 +79,7 @@ const ModuleFactory::ModuleTemplate *ModuleFactory::Find_Module_Template(const U
         return &it->second;
     }
 
+    captainslog_dbgassert(0, "Module name '%s' not found", name.Str());
```
xezon, line 82: I saw this in one of the binaries.

src/game/common/thing/thingfactory.cpp
```diff
@@ -52,9 +52,7 @@ ThingTemplate *ThingFactory::New_Template(const Utf8String &name)
     new_template->Friend_Set_Template_ID(m_nextTemplateID);
     m_nextTemplateID++;
 
-    if (m_nextTemplateID == 0) {
-        captainslog_dbgassert(0, "m_nextTemplateID wrapped to zero");
-    }
+    captainslog_dbgassert(m_nextTemplateID != 0, "m_nextTemplateID wrapped to zero");
```
xezon, line 55: Condition can be part of assert itself.

src/game/common/thing/thingfactory.cpp
```diff
@@ -68,12 +66,12 @@ ThingTemplate *ThingFactory::New_Override(ThingTemplate *thing_template)
         "Thing template '%s' not in master list",
         thing_template->Get_Name().Str());
 
+    ThingTemplate *override_template = static_cast<ThingTemplate *>(thing_template->Friend_Get_Final_Override());
```
xezon, line 69: Looks like it was `Friend_Get_Final_Override`, which is same as `Get_Final_Override` though. But it allows for `static_cast` here, which is nicer.

src/game/common/thing/thingfactory.cpp
```diff
@@ -217,15 +215,15 @@ void ThingFactory::Parse_Object_Definition(INI *ini, Utf8String name, Utf8String
     }
 
     if (reskin_from.Is_Not_Empty()) {
-        ThingTemplate *that = g_theThingFactory->Find_Template(reskin_from, true);
+        ThingTemplate *that = g_theThingFactory->Find_Template_Internal(reskin_from, true);
```
xezon: Non-World Builder shows `Find_Template_Internal` here, which looks more logical too.

jonwil: WorldBuilder confirms the original code used Find_Template. Find_Template just happens to be inlined into the other binaries.

jonwil (changes requested):
Other than the Find_Template thing, this looks good to me.

src/game/common/thing/thingfactory.cpp
xezon: reverted.

jonwil (approved):
Looks good to me now. Approved.

tomsons26 (approved):

PR #472, opened 2022-01-07
Fix ThingFactory::Parse_Object_Definition function arguments

xezon:
As seen in code.


jonwil (approved):
Looks correct to me, approved.

tomsons26 (approved):

PR #477, opened 2022-01-08
Guard hooker.h with compile error if used in non GAME_DLL

xezon:


duncanspumpkin (approved):
This looks sensible i don't think it needs @OmniBlade review

PR #478, opened 2022-01-08
Fix ExperienceTracker incorrectly calculating Vet Level

duncanspumpkin:
Off by 1 error as it was trying to loop


duncanspumpkin:
Confirmed fixes promotions of hackers and pilots.

xezon (approved):

PR #479, opened 2022-01-08
Fix bounding sphere calc

duncanspumpkin:
Calc was missing m_majorRadius


tomsons26:
since we touch geoinfo
```
float GeometryInfo::Get_Max_Height_Below_Position() const
{
    switch (m_type) {
        case GEOMETRY_SPHERE:
            return m_majorRadius;
        case GEOMETRY_CYLINDER: // Fallthrough
        case GEOMETRY_BOX:
            return 0.0f;
        default:
            captainslog_error("GeometryInfo::Get_Max_Height_Below_Position - should never get here");
            break;
    }

    return 0.0f;
}
```
noticed based on pseudo it should be like that, functionally same but logical

tomsons26 (approved):

PR #480, opened 2022-01-08
Fix infinite loop in W3DDisplay::Take_ScreenShot

xezon:


src/game/common/globaldata.h
```diff
@@ -63,6 +63,8 @@ class GlobalData : public SubsystemInterface
 
     bool Set_Time_Of_Day(TimeOfDayType time);
 
+    const char *Get_Path_User_Data() const { return m_userDataDirectory.Str(); }
```
xezon: Seen in World Builder

src/platform/w3dengine/client/w3ddisplay.cpp
```diff
@@ -1434,19 +1434,18 @@ void Create_Bmp_File(const char *filename, const char *data, int width, int heig
...
+    while (true) {
         sprintf(buf, "%s%.3d.bmp", "sshot", currentShot);
         currentShot++;
-        strcpy(fname, g_theWriteableGlobalData->m_userDataDirectory);
+        strcpy(fname, g_theWriteableGlobalData->Get_Path_User_Data());
         strcat(fname, buf);
 
-        if (_access(fname, 0) != -1) {
-            found = true;
+        if (_access(fname, 0) == -1) {
```
duncanspumpkin, line 1447: you could just make this a do while and skip the while(true).

xezon: I think it's fine as is. Compiler will know what to do.

src/game/common/globaldata.h
tomsons26: Should actually return AsciiString instance

src/platform/w3dengine/client/w3ddisplay.cpp
tomsons26: found shouldn't be removed, however "!= -1" should be "== -1" yeah

src/game/common/globaldata.h
xezon: Very well.

src/game/common/globaldata.h
```diff
@@ -63,6 +63,8 @@ class GlobalData : public SubsystemInterface
 
     bool Set_Time_Of_Day(TimeOfDayType time);
 
+    Utf8String Get_Path_User_Data() const { return m_userDataDirectory.Str(); }
```
jonwil: This seems wrong, it shouldn't be calling .Str() if its returning Utf8String.
Fix this and it should be good.

jonwil (changes requested):

src/game/common/globaldata.h
xezon: Fixed.

jonwil (approved):

PR #487, opened 2022-01-09
Further fixes to GeometryInfo.

tomsons26:


src/game/common/system/geometry.cpp
```diff
@@ -334,6 +350,11 @@ void GeometryInfo::Clip_Point_To_Footprint(const Coord3D &pos, Coord3D &point) c
     }
 }
 
+inline bool Is_Point_In_Range(float min, float val, float max)
+{
+    return min <= val && val <= max;
```
duncanspumpkin, line 352: That's a very weird way of writing this. Could you make it val > min && val <= max. Also are you sure it's exclusive of min and inclusive of max. Again a bit weird.

src/game/common/system/geometry.cpp
```diff
@@ -346,8 +367,9 @@ bool GeometryInfo::Is_Point_In_Footprint(const Coord3D &pos, const Coord3D &poin
         case GEOMETRY_CYLINDER:
             return Sqrt(Square(point.y - pos.y) + Square(point.x - pos.x)) <= m_majorRadius;
         case GEOMETRY_BOX:
-            return pos.x - m_majorRadius <= point.x && point.x <= pos.x + m_majorRadius && pos.y - m_minorRadius <= point.y
-                && point.y <= m_minorRadius + pos.y;
+            return Is_Point_In_Range(pos.x - m_majorRadius, point.x, pos.x + m_majorRadius)
+                && (Is_Point_In_Range(pos.y - m_minorRadius, point.y, pos.y + m_minorRadius));
```
duncanspumpkin: Drop the superfluous () around the second call

src/game/common/system/geometry.cpp
```diff
@@ -248,10 +248,19 @@ float GeometryInfo::Is_Intersected_By_Line_Segment(Coord3D &loc, Coord3D &from,
...
-    float xy = Sqrt(Square(coord2.y - coord1.y) + Square(coord2.x - coord1.x));
-    f2 = Atan2(xy, float(coord2.z + info.Get_Max_Height_Above_Position()) - z);
-    f1 = Atan2(xy, float(coord2.z - info.Get_Max_Height_Below_Position()) - z);
+    float x;
+    float y;
+
+    Coord3D coord3;
+    Get_Center_Position(coord1, coord3);
+
+    x = Sqrt(Square(coord2.y - coord3.y) + Square(coord2.x - coord3.x));
```
duncanspumpkin: ~~Rename to hyp would make this clearer since it has nothing to do with x~~
Scratch that it's the adj which is the hyp of x/y.

src/game/common/system/geometry.cpp
```diff
@@ -248,10 +248,19 @@ float GeometryInfo::Is_Intersected_By_Line_Segment(Coord3D &loc, Coord3D &from,
...
+
+    y = coord2.z + info.Get_Max_Height_Above_Position() - coord3.z;
```
duncanspumpkin: can we keep this as z or opp

src/game/common/system/geometry.cpp
```diff
@@ -248,10 +248,19 @@ float GeometryInfo::Is_Intersected_By_Line_Segment(Coord3D &loc, Coord3D &from,
...
+    f2 = Atan2(y, x);
+
+    y = coord2.z - info.Get_Max_Height_Below_Position() - coord3.z;
```
duncanspumpkin: I feel this function would make more sense if on the first line we did:
```cpp
const auto coord_diff = coord2 - coord3;
```
Then all the sums look much simpler:
```cpp
const auto adj = Sqrt(Square(coord_diff.x) + Square(coord_diff.y));
const auto opp_below = coord_diff.z - info.Get_Max_Height_Below_Position();
const auto opp_above = coord_diff.z - info.Get_Max_Height_Above_Position();
f1 = Atan2(opp_below, adj);
f2 = Atan2(opp_above, adj);

File: src/game/common/system/geometry.cpp [hunk 6]
```diff
@@ -270,30 +279,37 @@ void GeometryInfo::Get_2D_Bounds(const Coord3D &pos, float angle, Region2D &regi
...
+
             float opp_major = sin_theta * m_majorRadius;
             float opp_minor = sin_theta * m_minorRadius;
-            region.lo.x = region.hi.x = float(pos.x - adj_major) - opp_minor;
-            region.lo.y = region.hi.y = float(pos.y + adj_minor) - opp_major;
 
-            float tmp_x = float(adj_major + pos.x) - opp_minor;
-            float tmp_y = float(opp_major + pos.y) + adj_minor;
+            float tmp_x = float(pos.x - adj_major) - opp_minor;
+            float tmp_y = float(pos.y + adj_minor) - opp_major;
```
duncanspumpkin: These float casts are pointless all these calcs are with float's and floats are always converted to.

src/game/common/system/geometry.cpp
```diff
@@ -270,30 +279,37 @@ void GeometryInfo::Get_2D_Bounds(const Coord3D &pos, float angle, Region2D &regi
...
+            tmp_x = float(pos.x + adj_major) + opp_minor;
             tmp_y = float(pos.y - adj_minor) + opp_major;
             region.lo.x = std::min(region.lo.x, tmp_x);
             region.lo.y = std::min(region.lo.y, tmp_y);
             region.hi.x = std::max(region.hi.x, tmp_x);
             region.hi.y = std::max(region.hi.y, tmp_y);
 
-            tmp_x = float(opp_minor + pos.x) - adj_major;
+            tmp_x = float(pos.x - adj_major) + opp_minor;
             tmp_y = float(pos.y - adj_minor) - opp_major;
```
duncanspumpkin: I see this function as calculating the 4 x points and the 4 y points and using the min/max of the set.
```cpp
float points_x[4] = {
    pos.x - adj_major - opp_minor, 
    pos.x + adj_major - opp_minor, 
    pos.x + adj_major + opp_minor,
    pos.x - adj_major + opp_minor,
};
region.lo.x = std::min_element(std::begin(points_x), std::end(points_x));
region.hi.x = std::max_element(std::begin(points_x), std::end(points_x));
// repeat for y
```
And when you have it like that you can see that pos.x is common across this so you could remove that from the table calc and perform the additon only once! Simplifying the code, making it even more readable and speeding it up by reducing calcs and branches. But perhaps a bit too radicle.

duncanspumpkin (approved):

PR #488, opened 2022-01-10
Set unique ID for water polygon trigger

xezon:
I am 100% sure this is logically the right thing. In current implementation, the Water Trigger ID will either be 0 (invalid ID) or share the same ID as another Polygon Trigger. The pre-increment fixes that.

I don't know how this effectively can be tested in game, but I did load map and stepped through the code, and the trigger ID's looked good with this change.


tomsons26 (changes requested):
Has to be carefully tested with a lot of SP maps and approved by a experienced mapper so putting a block on this until then

PR #489, opened 2022-01-10
Fix polygon trigger radius

xezon:
OmniBlade on behalf of OpenSAGE developer:

> I found one little thing you might be interested in - while parsing .sav files, specifically the state for PolygonTrigger, I think I found a bug in the original game. I found a value that looks like it's supposed to be the radius of the polygon trigger, but it's wrong. Maybe it's something you can fix in Thyme? The value stored in .sav files is what you get if you do this:

```c++
width = (bottomRight.X - topLeft.X) * 0.5
height = (bottomRight.Y + topLeft.Y) * 0.5
radius = sqrt(width * width + height * height)
```

> ... but that's wrong. The height should be bottomRight.Y - topLeft.Y, not bottomRight.Y + topLeft.Y. If this is indeed a bug in the original game, the radius would be way too large, and would presumably lead to many false-positives when doing the radius check, which would be bad for performance. Presumably there's then a more precise check that uses the actual polygon, so this bug wouldn't actually change anything... it would just make it slower than it needs to be.

> I added a comment about this buggy radius here: https://github.com/OpenSAGE/OpenSAGE/blob/899b380545a7ca17aa9cae936974a3ce367ec018/src/OpenSage.Game/Data/Map/PolygonTrigger.cs#L185-L199

### DIFF
--- a/src/game/logic/map/polygontrigger.cpp
+++ b/src/game/logic/map/polygontrigger.cpp
@@ -39,7 +39,7 @@ PolygonTrigger::PolygonTrigger(int initial_allocation) :
     m_isShownInLayer(true),
     m_isSelected(false),
     m_bounds{},
-    m_radius(0),
+    m_radius(0.0f),
     m_boundsNeedsUpdate(false)
 {
     if (initial_allocation < 2) {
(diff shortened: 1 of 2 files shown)


xezon:
I don't know how to test this fix. But should be possible to test this in some sort of test map. I will ask kABUSE.

tomsons26 (changes requested):
Has to be carefully tested with a lot of SP maps and approved by a experienced mapper so putting a block on this until then

xezon:
PolygonTrigger refers to water area. That is the thing with the blue line in World Builder. Tournament Desert (2) map has 2 of these polygon triggers for example (ID 7 and 9).

src/game/logic/map/polygontrigger.cpp
```diff
@@ -128,8 +128,11 @@ void PolygonTrigger::Update_Bounds() const
     }
 
     m_boundsNeedsUpdate = false;
-    float x = (float)(m_bounds.hi.x - m_bounds.lo.x) / 2.0f;
-    float y = (float)(m_bounds.lo.y + m_bounds.hi.y) / 2.0f;
+
+    // #BUGFIX Calculate correct height value here.
+    float x = (float)(m_bounds.Width()) / 2.0f;
```
duncanspumpkin: you can drop the casts. Make the comment `Bugfix Vanilla calculated midpoint for y instead of height`

xezon: Ok no problem. I reworded the comment better to explain what is happening and refactored the code so it is easier to understand. Removed C casts.

PR #491, opened 2022-01-10
Rename and add comments to SparseMatchFinder Find_Best_Info_Slow

duncanspumpkin:
Whilst investigating the flashbang issue I ended up renaming bits of this.


jonwil (approved):
Looks good to me as this just renames some things and adds some comments and the names and comments make sense.

Issue #482, opened 2022-01-08
Ranger Flashbang is not usable


xezon:
Ranger's Flashbang weapon cannot be used after research completed.

Repro:
- Load match with USA
- Build Barracks with Dozer
- Build Ranger with Barracks
- Build Flashbang Grenade with Barracks

Result:
Flashbang is not usable with Ranger.

![sshot001](https://user-images.githubusercontent.com/4720891/148655861-24e2f5ea-0a16-4528-9bfb-8c156cfc8376.jpg)

![sshot002](https://user-images.githubusercontent.com/4720891/148655864-e054bb71-8d85-4b7b-a19f-9b4255a0d3eb.jpg)

PR #492, opened 2022-01-10
Fix flashbang issue caused by incorrect Map comparator

duncanspumpkin:


src/game/common/sparsematchfinder.h
```diff
@@ -22,7 +22,7 @@ template<typename Type, typename Key> class SparseMatchFinder
     class MapHelper
     {
     public:
-        bool operator()(const Key &left, const Key &right) const { return left == right; }
+        bool operator()(const Key &left, const Key &right) const { return left != right; }
```
jonwil: This is not correct, we should implement this the way the original did it (see function 006F1520 in ZH worldbuilder) to avoid any risk of bugs.

jonwil (changes requested):

src/game/common/sparsematchfinder.h
```diff
@@ -22,7 +22,23 @@ template<typename Type, typename Key> class SparseMatchFinder
...
+        // This class implements a < operation for BitFlags
+        bool operator()(const Key &left, const Key &right) const
+        {
+            if (left.Size() < right.Size()) {
+                return true;
+            }
+            for (auto bit = 0; bit < left.Size(); ++bit) {
+                const auto left_bit = left.Test(bit);
+                const auto right_bit = right.Test(bit);
+                if ((left_bit && right_bit) || (!left_bit && !right_bit)) {
```
duncanspumpkin, line 35: This could be simplified to `!(left_bit ^ right_bit)`

xezon:
Fixes #482

jonwil (approved):
Looks good now, approved.

PR #493, opened 2022-01-11
Properly document the Thyme change in Is_Resolution_Valid function

xezon:
Adding FEATURE tag and deactivating code with `#if 0` helps discoverability.


jonwil (approved):
Looks good to me, approved.

PR #494, opened 2022-01-11
Add missing float literals in various files

xezon:
These are simple float literals added. None of these were worth verifying against decomp.

### DIFF
--- a/src/w3d/lib/iniclass.cpp
+++ b/src/w3d/lib/iniclass.cpp
@@ -419,7 +419,7 @@ float INIClass::Get_Float(const char *section, const char *entry, float defvalue
 
         // Is this actually a percentage? if so, divide it by 100
         if (strchr(entryptr->Get_Value(), '%')) {
-            return defvalue / (float)100.0;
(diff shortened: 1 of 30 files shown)


src/w3d/lib/iniclass.cpp
```diff
@@ -446,7 +446,7 @@ double INIClass::Get_Double(const char *section, const char *entry, double defva
         // Is this value really a percentage? if so, divide it by 100.
         //
         if (strchr(entryptr->Get_Value(), '%')) {
-            return defvalue / 100.0;
+            return defvalue / 100.0f;
```
jonwil: This should probably not be changed given that its operating on a double value.

jonwil (changes requested):

src/w3d/lib/iniclass.cpp
xezon: Ups. Reverted.

jonwil (approved):
Looks good to me, I don't see anywhere else that's specifically working with doubles. Approved.

PR #495, opened 2022-01-11
Implement PlayerList

duncanspumpkin:
Implemented as far as possible

### DIFF
--- a/src/game/common/rts/player.h
+++ b/src/game/common/rts/player.h
@@ -18,22 +18,60 @@
 #include "snapshot.h"
 #include "unicodestring.h"
 
+class PlayerTemplate;
+class Team;
+class SpecialPowerTemplate;
+class Object;
+class Waypoint;
+enum Relationship;
+
 // TODO this is just a skeleton to support other objects at the moment.
 class Player : public SnapShot
 {
 public:
+    Player(int32_t player_index);
+    virtual ~Player();
+
+    // SnapShot interface
+    virtual void CRC_Snapshot(Xfer *xfer) override;
+    virtual void Xfer_Snapshot(Xfer *xfer) override;
+    virtual void Load_Post_Process() override {}
+
+    virtual void Compute_Superweapon_Target(const SpecialPowerTemplate *sp_template, Coord3D *loc, int32_t unk1, float unk2);
+    virtual bool Check_Bridges(Object *obj, Waypoint *waypoint);
+    virtual bool Get_Ai_Base_Center(Coord3D *center);
+    virtual void Repair_Structure(ObjectID obj_id);
+
+    NameKeyType Get_Player_NameKey() { return m_playerNameKey; }
     int Get_Player_Index() { return m_playerIndex; }
+    uint16_t Get_Player_Mask() { return 1ULL << m_playerIndex; }
     int Get_Color() { return m_playerColor; }
+    // wb: 0x006DA8F0
+    Team *Get_Default_Team()
+    {
+        captainslog_dbgassert(m_defaultTeam != nullptr, "Default team is null.");
+        return m_defaultTeam;
+    }
+    Relationship Get_Relationship(Team *team);
 
     void On_Power_Brown_Out_Change(bool b);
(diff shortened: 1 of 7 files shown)


src/game/common/rts/playerlist.cpp
```diff
@@ -18,3 +18,234 @@
...
+    for (auto i = 0; i < MAX_PLAYER_COUNT; ++i) {
+        m_players[i] = new Player(i);
+    }
+    Init();
+}
+
+// zh: 0x0045A8D0 wb: 0x007BEEFD
+PlayerList::~PlayerList()
+{
+    Init();
```
tomsons26: Add a todo to investigate as this is quite illogical

src/game/common/rts/playerlist.cpp
```diff
@@ -18,3 +18,234 @@
...
+    for (auto i = 0; i < Get_Player_Count(); ++i) {
+        auto *player = Get_Nth_Player(i);
+        if (player == nullptr || player == target_player) {
+            continue;
+        }
+
+        const auto relationship = target_player->Get_Relationship(player->Get_Default_Team());
+        switch (relationship) {
+            case Relationship::ENEMIES:
+                if (relationship_mask & (1ULL << 2)) {
```
tomsons26: why is this more preferable than just & 4?

duncanspumpkin: These are flags. I was going to switch them to an enum but forgot. Will add that shortly.

src/game/common/rts/playerlist.cpp
duncanspumpkin: Init clears out a lot of things and unloads the PlayerTemplate I think that is the reason for it. Init seems to get called a lot.

src/game/common/rts/playerlist.h
```diff
@@ -18,46 +18,71 @@
 #include "snapshot.h"
 #include "subsysteminterface.h"
 
+class Team;
+
+constexpr int32_t MAX_PLAYER_COUNT = 16;
```
tomsons26: should be a in a more global header, its gonna be used a lot

duncanspumpkin: Any suggestions as to which one?

duncanspumpkin: Suggested gametypes.h

src/game/common/rts/playerlist.h
```diff
@@ -48,7 +56,8 @@ class PlayerList : public SubsystemInterface, public SnapShot
     void Set_Local_Player(Player *player);
     Player *Get_Player_From_Mask(uint16_t mask);
     Player *Get_Each_Player_From_Mask(uint16_t &mask);
-    uint16_t Get_Players_With_Relationship(int32_t player_idx, uint16_t relationship_mask);
+    uint16_t Get_Players_With_Relationship(
+        int32_t player_idx, uint16_t relationship_mask); // Uses PlayerRelationshipFlags for mask
```
xezon: You could get rid of the comment by creating a `using relationship_flags_t = uint16_t` or similar nearby the flag definition.

src/game/common/rts/player.h
```diff
@@ -18,22 +18,60 @@
...
+    NameKeyType m_playerNameKey;
     int m_playerIndex;
-    char unk[0xFC];
+    char unk4[0xFC];
     int m_playerColor;
-};
\ No newline at end of file
+    uint8_t unk5[0x160 - 0x128];
+    Team *m_defaultTeam;
+    uint8_t unk6[0x44C - 0x164]; // There is a size 4 field specific to WB in here
```
xezon: WorldBuilder specific bytes should not be compiled into non-`GAME_DEBUG_STRUCTS` afaik.

```c++
#ifdef GAME_DEBUG_STRUCTS
uint8_t unk6[0x44C - 0x164];
#else
uint8_t unk6[0x44C - 0x164 - 4];
#endif
```

tomsons26 (approved):

PR #496, opened 2022-01-11
Avoid nullptr crash in W3DDisplay::Gather_Debug_Stats

xezon:
Just a few lines above, `object` is tested, but when getting its physics, it was untested.

### DIFF
--- a/src/platform/w3dengine/client/w3ddisplay.cpp
+++ b/src/platform/w3dengine/client/w3ddisplay.cpp
@@ -1925,7 +1925,8 @@ void W3DDisplay::Gather_Debug_Stats()
                 drawable->Get_Position()->y,
                 drawable->Get_Position()->z);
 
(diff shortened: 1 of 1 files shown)


src/platform/w3dengine/client/w3ddisplay.cpp
```diff
@@ -1925,7 +1925,8 @@ void W3DDisplay::Gather_Debug_Stats()
                 drawable->Get_Position()->y,
                 drawable->Get_Position()->z);
 
-            const PhysicsBehavior *phys = object->Get_Physics();
+            // #BUGFIX Test object before getting physics.
```
duncanspumpkin, line 1928: We don't really add a comment like this when it's just a nullptr check

xezon: It deviates from original, so comment should be ok, no? It also helps compiling change lists later to say what has changed in Thyme.

PR #498, opened 2022-01-12
Prevent bad pointer access in hasher

duncanspumpkin:
I've also added notes as to why it should exist.

### DIFF
--- a/src/game/rtsutils.h
+++ b/src/game/rtsutils.h
@@ -68,13 +68,13 @@ template<> struct hash<Utf8String>
 {
     size_t operator()(const Utf8String &object) const
     {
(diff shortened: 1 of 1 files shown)


xezon (approved):

src/game/rtsutils.h
```diff
@@ -69,12 +69,12 @@ template<> struct hash<Utf8String>
     size_t operator()(const Utf8String &object) const
     {
         const char *c = reinterpret_cast<const char *>(object.Str());
```
xezon: The reinterpret_cast here can be removed.

xezon (changes requested):

xezon (approved):

PR #484, opened 2022-01-09
Fix type operator post-increment, shift and unsigned types

xezon:

PR #499, opened 2022-01-12
Implement new GameTextFile, GameTextManager classes and gametextcompiler tool

xezon:
I setup 5 separate commits to organize this change better.

The new utility classes and function are used in the new feature code. Some of the functionality is redundant to c++11 to c++20 library features, but due the existence of STL Port had to be implemented for Thyme.

The `FileRef` class is an optional wrapper for `File*` to allow using game files more comfortably without worrying about leaks.

The `GameTextFile` class contains all new code to read and write CSF, STR and Multi STR files. Read result is identical to original implementation for the most part, but can also read UTF8 STR, which the original cannot. Performance likely is much better, mainly because of using new Buffered File IO. Read and Write have been designed with performance in mind.

The `GameTextManager` class is an alternative implementation to the legacy `GameTextManager` class using the new `GameTextFile`. It provides the same functionality as the original, and a bit more sugar on top. Memory footprint is a little smaller than original, due absence of buffers as class members.

The `gametextcompiler` supports simple command line aka `-LOAD_STR generals.csf` but also more powerful function like syntax which can perform more complex operations.

### New Utilities Overview ###

**rts::array<>** : Wants to be the same as `std::array` (c++11). Does suffer from non-constexpr `std::reverse_iterator` due STL Port.

**rts::array_view<>** : Wants to be the same as `std::span` (c++20). Does suffer from non-constexpr `std::reverse_iterator` due STL Port. Constructors are not quite the same as `std::span`.

**rts array util functions** : Functions to build `rts::array_view<>`. Has custom functions catered to `Utf8String` and `Utf16String`. **Has no STL equivalent as far as I know.**

**rts::ebitflags<>** : Capsulates enum or enum class with bit values into object with some functionality for bit operations. **Has no equivalent in STL.**

**rts::enumflags<>** : Capsulates enum or enum class with incremental values into object with some functionality for bit operations. `std::bitset` is similar, but is not catered to `enum` types, so they cannot be used interchangeably. **Has no equivalent in STL.**

**rts::enumerator<>** : Capsulates `enum` or `enum class` with incremental values into object with some functionality for enumeration operations. **Has no equivalent in STL.**

**rts file utility functions** : Provides some helper functions to read and write data with game files. **Has no equivalent in STL.**

**rts::intrusive_ptr<>** : Wraps any reference counted object and provides strong guarantee that the contained object will be deleted when all references are destroyed. Is compatible with DirectX classes. **Has no equivalent in STL.**

**rts:shared_ptr<>** : Wraps any heap allocated object and provides strong guarantee that the contained object will be deleted when all references are destroyed. Is similar to `std::shared_ptr<>`, though not quite as sophisticated. E.g. it does not support automatic `delete[]` when an array pointer was assigned. Array deleter has to be explicitly set when needed. Also locality of the allocated reference counter is not optimal in `rts::shared_ptr<>`. As far as I know in `std::shared_ptr<>` it somehow allocates reference counter close to object when using `std::make_shared<>`. `rts:shared_ptr<>` does allow for non-atomic reference counter. In any case, user code of `rts::shared_ptr<>` could be swapped for `std::shared_ptr<>`.

**rts ref counter** : has functions and classes that implement atomic and non-atomic reference counters. They are fit to be used with `rts::intrusive_ptr<>` and `rts::shared_ptr<>`. **Has no equivalent in STL.**

**rts atomicop functions** : Interlocked functions for atomic integer operations. Is used due the absence of `std::atomic` in STL Port.

**deleter functions** : Deleters for smart pointers. **Has no equivalent in STL.**

**rts::sized integer classes** : Will be removed once this code is in typesize.h. See: #484

**rts util functions** : Just some helper functions for STL. **Has no equivalent in STL.**

### DIFF
--- a/src/tools/gametextcompiler/src/processor.cpp
+++ b/src/tools/gametextcompiler/src/processor.cpp
@@ -1,0 +1,713 @@
+/**
+ * @file
+ *
+ * @author xezon
+ *
+ * @brief Game Text Compiler Processor. (Thyme Feature)
+ *
+ * @copyright Thyme is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#include "processor.h"
+#include "commands.h"
+#include <algorithm>
+#include <utility/stringutil.h>
+
+namespace Thyme
+{
+namespace
+{
+constexpr const char *const s_result_names[] = {
+    "SUCCESS",
+    "INVALID_COMMAND_ACTION",
+    "INVALID_COMMAND_ARGUMENT",
+    "INVALID_LANGUAGE_VALUE",
+    "INVALID_OPTION_VALUE",
+    "INVALID_FILE_ID_ARGUMENT",
+    "MISSING_FILE_PATH_ARGUMENT",
+    "MISSING_LANGUAGE_ARGUMENT",
+    "EXECUTION_ERROR",
+};
+
+static_assert(ARRAY_SIZE(s_result_names) == size_t(Processor::ResultId::COUNT));
+} // namespace
+
+const char *Processor::Get_Result_Name(ResultId id)
+{
+    return s_result_names[size_t(id)];
+}
+
+const Processor::FileId Processor::s_defaultFileId = { 0 };
+
+Processor::Processor() : m_fileMap(), m_commands() {}
+
+Processor::Result Processor::Parse_Commands(const CommandTexts &command_texts)
+{
+    if (Has_Simple_Command(command_texts)) {
+        return Parse_Simple_Commands(command_texts);
+    } else {
+        return Parse_Function_Commands(command_texts);
+    }
+}
+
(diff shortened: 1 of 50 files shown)


src/tools/gametextcompiler/src/processor.cpp
```diff
@@ -0,0 +1,713 @@
...
+        auto *file = new GameTextFile();
+        file->Set_Options(GameTextOptions::Value::NONE);
+        file_map.emplace(file_id, file);
+    }
+}
+
+GameTextFilePtr Processor::Get_File_Ptr(const FileMap &file_map, FileId file_id)
+{
+    const FileMap::const_iterator it = file_map.find(file_id);
+    captainslog_assert(it != file_map.end());
```
xezon, line 660: captainslog assert will not hit in gametextcompiler. Needs fixing.

xezon: Fixed.

xezon:
Code updated to latest Thyme.

STR comment bug fixed: Now recognizes `//` comment correctly.

Captains Log now works correctly in gametextcompiler tool.

Next steps: 
1. Replace rts template classes with std template classes in gametextcompiler where applicable
2. Replace rts template classes with std template classes in GameTextFile class where applicable

PR #500, opened 2022-01-13
Fix housecolor crash and improve readability of relevant code in GameAssetManager

xezon:
Loading **Tournament Desert** map crashes the game.

### DIFF
--- a/src/w3d/renderer/assetmgr.cpp
+++ b/src/w3d/renderer/assetmgr.cpp
@@ -767,13 +767,12 @@ bool Get_Mesh_Color_Methods(MeshClass *mesh, bool *housecolor, bool *zhc)
         matinfo->Release_Ref();
     }
 
-    const char *str = strchr(mesh->Get_Name(), '.');
-    const char *str2 = str == nullptr ? str + 1 : mesh->Get_Name();
+    captainslog_assert(mesh->Get_Name() != nullptr);
+    const char *word = strchr(mesh->Get_Name(), '.');
+    word = (word != nullptr) ? word + 1 : mesh->Get_Name();
 
-    if (str2 != nullptr) {
-        if (!strncasecmp(str2, "HOUSECOLOR", 10)) {
-            *housecolor = true;
-        }
+    if (!strncasecmp(word, "HOUSECOLOR", 10)) {
+        *housecolor = true;
     }
 
     return *housecolor || *zhc;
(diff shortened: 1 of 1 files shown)


duncanspumpkin (changes requested):
Can you just change the nullptr check to the correct way round it would be much cleaner.
`const char *str2 = str != nullptr ? str + 1 : mesh->Get_Name();`

src/w3d/renderer/assetmgr.cpp
```diff
@@ -758,7 +758,7 @@ bool Get_Mesh_Color_Methods(MeshClass *mesh, bool *housecolor, bool *zhc)
     if (matinfo != nullptr) {
         for (int i = 0; i < matinfo->Texture_Count(); i++) {
 
-            if (!strncasecmp(matinfo->Peek_Texture(i)->Get_Name(), "ZHC", 3)) {
+            if (strncasecmp(matinfo->Peek_Texture(i)->Get_Name(), "ZHC", 3) == 0) {
```
duncanspumpkin: unrelated revert

src/w3d/renderer/assetmgr.cpp
```diff
@@ -767,11 +767,12 @@ bool Get_Mesh_Color_Methods(MeshClass *mesh, bool *housecolor, bool *zhc)
...
-    const char *str = strchr(mesh->Get_Name(), '.');
-    const char *str2 = str == nullptr ? str + 1 : mesh->Get_Name();
+    captainslog_assert(mesh->Get_Name() != nullptr);
+    const char *word = strchr(mesh->Get_Name(), '.');
+    word = (word != nullptr) ? word + 1 : mesh->Get_Name();
 
-    if (str2 != nullptr) {
-        if (!strncasecmp(str2, "HOUSECOLOR", 10)) {
+    if (word != nullptr) {
+        if (strncasecmp(word, "HOUSECOLOR", 10) == 0) {
```
duncanspumpkin: unrelated revert

src/w3d/renderer/assetmgr.cpp
```diff
@@ -866,7 +869,7 @@ bool GameAssetManager::Recolor_Mesh(RenderObjClass *robj, uint32_t colour)
 
     for (int i = 0; i < matinfo->Texture_Count(); i++) {
         TextureClass *texture = matinfo->Peek_Texture(i);
-        if (!strncasecmp(texture->Get_Name(), "ZHC", 3)) {
+        if (strncasecmp(texture->Get_Name(), "ZHC", 3) == 0) {
```
duncanspumpkin: revert

xezon:
I updated the title of the change. That should do (Squash and merge). No need to revert it all.

src/w3d/renderer/assetmgr.cpp
xezon: fixed.

duncanspumpkin (approved):
I personally dont see this as more readable but its fine.

xezon:
On second look, it looks as if the `if (word != nullptr)` branch could be removed, no? At this point word is not to be expected null.

duncanspumpkin:
It can only be nullptr if get name is nullptr. I don't really want to think through all possibilities

xezon:
If `Get_Name()` was null, then it would crash 2 lines before already at `strchr(mesh->Get_Name(), '.');`. That's why I placed the assert just before that.

duncanspumpkin:
I don't think it is guaranteed to crash in that function standard doesn't specify if strchr is nullptr safe.

xezon:
I check later. If `strchr` was checking null, then the `assert` would be obsolete. If `strchr` is not checking null, then the assert is ok and the condition is obsolete.

src/w3d/renderer/assetmgr.cpp
xezon: reverted

PR #503, opened 2022-01-14
Force x87 floating point instructions

duncanspumpkin:


OmniBlade:
Probably want to add `/fp:fast` as well to match the optimisations used in the original binary.

duncanspumpkin:
I was having a read into /fp:fast and it appears that modifying the status registers is not allowed in /fp:fast.

xezon (approved):

PR #506, opened 2022-01-15
Fix motchan filter table math

xezon:
This appears to be what game does. As seen in MAC.

The DEG_TO_RAD is a double with the PI value that is now added to gamemath header.

### DIFF
--- a/src/w3d/renderer/motchan.cpp
+++ b/src/w3d/renderer/motchan.cpp
@@ -445,14 +445,15 @@ AdaptiveDeltaMotionChannelClass::AdaptiveDeltaMotionChannelClass() :
     m_type(0),
     m_vectorLen(0),
     m_numFrames(0),
-    m_scale(0),
+    m_scale(0.0f),
     m_data(nullptr),
     m_cacheFrame(0),
     m_cacheData(nullptr)
 {
     if (!g_tableValid) {
+        const double quarter_rot = DEG_TO_RAD(90.0);
         for (int i = 0; i < 240; ++i) {
-            g_filterTable[i + 16] = 1.0 - GameMath::Sin(i / 240.0 * DEG_TO_RAD(90.0));
+            g_filterTable[i + 16] = 1.0f - GameMath::Sin((i / 240.0f) * quarter_rot);
         }
         g_tableValid = true;
     }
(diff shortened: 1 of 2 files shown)


src/w3d/renderer/motchan.cpp
```diff
@@ -432,14 +432,15 @@ AdaptiveDeltaMotionChannelClass::AdaptiveDeltaMotionChannelClass() :
...
+    m_scale(0.0f),
     m_data(nullptr),
     m_cacheFrame(0),
     m_cacheData(nullptr)
 {
     if (!g_tableValid) {
+        const double rad = DEG_TO_RAD(90.0);
         for (int i = 0; i < 240; ++i) {
-            g_filterTable[i + 16] = 1.0 - GameMath::Sin(i / 240.0 * DEG_TO_RAD(90.0));
+            g_filterTable[i + 16] = 1.0f - GameMath::Sin(((float)i / 240.0f) * rad);
```
duncanspumpkin: there isn't any point adding the cast and just clutters the code:

src/w3d/renderer/motchan.cpp
```diff
@@ -432,14 +432,15 @@ AdaptiveDeltaMotionChannelClass::AdaptiveDeltaMotionChannelClass() :
...
     m_vectorLen(0),
     m_numFrames(0),
-    m_scale(0),
+    m_scale(0.0f),
     m_data(nullptr),
     m_cacheFrame(0),
     m_cacheData(nullptr)
 {
     if (!g_tableValid) {
+        const double rad = DEG_TO_RAD(90.0);
```
duncanspumpkin: can you use a better name perhaps `half_pi`.

src/w3d/renderer/motchan.cpp
xezon: Is i / 240f treated correctly as float division though? It looks as if it could result in integer math.

duncanspumpkin: yes. It always promotes. 
https://en.cppreference.com/w/cpp/language/operator_arithmetic#Conversions

> Otherwise, if either operand is float, the other operand is converted to float

xezon: ok

tomsons26 (approved):

PR #507, opened 2022-01-15
Fix obsolete casts in matrix functions

xezon:
Just removed a bunch of obsolete casts. All of these operations are all float values.


jonwil (approved):
Looks good to me, approved.

PR #508, opened 2022-01-15
Add missing float literals and game math functions in some functions

xezon:
As seen in World Builder.

### DIFF
--- a/src/w3d/math/quaternion.cpp
+++ b/src/w3d/math/quaternion.cpp
@@ -26,14 +26,13 @@ namespace
 // Utility function used in Trackball
 float Project_To_Sphere(float r, float x, float y)
 {
-    const float SQRT2 = 1.41421356f;
     float t, z;
     float d = GameMath::Sqrt(x * x + y * y);
 
-    if (d < r * (SQRT2 / (2.0f))) { // inside sphere
+    if (d < r * (GAMEMATH_SQRT2 / (2.0f))) { // inside sphere
         z = GameMath::Sqrt(r * r - d * d);
     } else { // on hyperbola
-        t = r / SQRT2;
+        t = r / GAMEMATH_SQRT2;
         z = t * t / d;
     }
 
@@ -54,17 +53,17 @@ Quaternion::Quaternion(const Vector3 &axis, float angle)
 
 void Quaternion::Rotate_X(float theta)
 {
-    *this = (*this) * Quaternion(Vector3(1, 0, 0), theta);
+    *this = (*this) * Quaternion(Vector3(1.0f, 0.0f, 0.0f), theta);
 }
 
 void Quaternion::Rotate_Y(float theta)
 {
-    *this = (*this) * Quaternion(Vector3(0, 1, 0), theta);
+    *this = (*this) * Quaternion(Vector3(0.0f, 1.0f, 0.0f), theta);
 }
 
 void Quaternion::Rotate_Z(float theta)
 {
-    *this = (*this) * Quaternion(Vector3(0, 0, 1), theta);
+    *this = (*this) * Quaternion(Vector3(0.0f, 0.0f, 1.0f), theta);
 }
 
 void Quaternion::Normalize()
@@ -299,7 +298,6 @@ Quaternion Build_Quaternion(const Matrix4 &mat)
     return q;
 }
 
-#define SLERP_EPSILON 0.001f
 void Slerp_Setup(const Quaternion &p, const Quaternion &q, SlerpInfoStruct *slerpinfo)
 {
     float cos_t;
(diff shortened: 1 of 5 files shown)


src/w3d/math/quaternion.cpp
```diff
@@ -313,7 +311,9 @@ void Slerp_Setup(const Quaternion &p, const Quaternion &q, SlerpInfoStruct *sler
         slerpinfo->flip = false;
     }
 
-    if (1.0f - cos_t < SLERP_EPSILON) {
+    constexpr double slerp_epsilon = 0.001;
```
xezon, line 314: This is a `double` in World Builder. Whether or not that makes any practical difference, it does not hurt to leave it original.

src/w3d/math/quaternion.cpp
```diff
@@ -26,14 +26,13 @@ namespace
...
-    const float SQRT2 = 1.41421356f;
     float t, z;
     float d = GameMath::Sqrt(x * x + y * y);
 
-    if (d < r * (SQRT2 / (2.0f))) { // inside sphere
+    if (d < r * (GAMEMATH_SQRT2 / (2.0f))) { // inside sphere
         z = GameMath::Sqrt(r * r - d * d);
     } else { // on hyperbola
-        t = r / SQRT2;
+        t = r / GAMEMATH_SQRT2;
```
xezon, line 35: GAMEMATH_SQRT2 is 1.41421356f

src/w3d/renderer/camera.cpp
```diff
@@ -371,7 +371,7 @@ void CameraClass::Get_D3D_Projection_Matrix(Matrix4 *set_tm)
 
     *set_tm = m_projectionTransform;
 
-    float oozdiff = 1.0 / (m_zFar - m_zNear);
+    float oozdiff = double(1.0) / (m_zFar - m_zNear);
```
duncanspumpkin: what are you trying to achieve here. double cast to double???

src/w3d/math/quaternion.cpp
```diff
@@ -54,17 +53,17 @@ Quaternion::Quaternion(const Vector3 &axis, float angle)
...
 void Quaternion::Rotate_Y(float theta)
 {
-    *this = (*this) * Quaternion(Vector3(0, 1, 0), theta);
+    *this = (*this) * Quaternion(Vector3(0.f, 1.f, 0.f), theta);
 }
 
 void Quaternion::Rotate_Z(float theta)
 {
-    *this = (*this) * Quaternion(Vector3(0, 0, 1), theta);
+    *this = (*this) * Quaternion(Vector3(0.f, 0.f, 1.f), theta);
```
duncanspumpkin: it looks weird not using one unit after the decimal point for a floating point this comment applies to all of this pr

src/w3d/renderer/camera.cpp
xezon: The purpose here is to mark this as double explicitly, so there is no question whether or not this is float. For compile this makes no difference.

src/w3d/math/quaternion.cpp
xezon: It is valid code though. Can append a 0 if so desired for style.

src/w3d/renderer/camera.cpp
duncanspumpkin: For reading it makes no difference either. Remove

src/w3d/math/quaternion.cpp
duncanspumpkin: I think that would look better

src/w3d/renderer/camera.cpp
xezon: It does. When looking at it as is, one may think it should be a float (which it would be if we wrote it).

duncanspumpkin: So you are suggesting every time we use a double we add `double()`?

xezon: It does help to signal "Hey, this is a double, although you would expect a float here."

If a double is used in double context, it would be not necessary to highlight explicit double. But here double is used in float context, so marking explicit double helps understanding that this is meant to be double. Alternatively we can add a comment instead.

duncanspumpkin: We already have excessive amounts of casts.

This doesn't make any difference to the sum in any context i see no reason in doing this. 1.0f and 1.0 are exactly the same. The only difference is which load instruction is used and wasting slightly more storage with a double.

We set the x87 flag to 32bit float calculations so all doubles are floats when calculating (except we don't actually use x87 maths so the flag does nothing for now so actually it's the reverse all floating point calculations are performed at long double precision).

src/w3d/math/quaternion.cpp
xezon: Zeros added.

src/w3d/renderer/camera.cpp
xezon: Replaced explicit double keyword with comment.

jonwil (approved):
Looks good to me, approved.

PR #509, opened 2022-01-15
Refactor matrix math to be closer to what the assembler does

xezon:
In assembler, it does not multiply value by 2, but adds it to become twice. I tested if there is any difference between 2.0f * x and x + x, but there appeared no difference in the calculated results. Anyway, I thought it can't hurt to write it the way it is calculated in assembler for best representation of calculation.

### DIFF
--- a/src/w3d/math/gamemath.h
+++ b/src/w3d/math/gamemath.h
@@ -56,6 +56,8 @@
 #define DEG_TO_RADF(x) (((float)x) * GAMEMATH_PI / 180.0f)
 #endif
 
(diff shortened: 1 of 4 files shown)


duncanspumpkin:
I disagree with doing this the optimiser will decide this and will likely undo this

OmniBlade:
I agree with @duncanspumpkin this is a micro optimisation that is best left to the optimiser to worry about. If we have issues with determinism on different platforms down the line then it will be something to look at then, but I'm happy to leave the maths as it is for now.

PR #510, opened 2022-01-15
Implement SoundFXNugget

duncanspumpkin:


src/game/client/fxlist.cpp
```diff
@@ -68,3 +72,54 @@ void FXList::Parse(INI *ini, void *, void *store, const void *)
...
+        secondary_pos = secondary->Get_Position();
+    }
+    Do_FX_Pos(primary_pos, primary_mtx, 0.0f, secondary_pos, 0.0f);
+}
+
+// zh: 0x004C9F20 wb: 0x0075F800
+void SoundFXNugget::Do_FX_Pos(
+    const Coord3D *primary, const Matrix3D *primary_mtx, float primary_speed, const Coord3D *secondary, float radius)
+{
+    AudioEventRTS audio_event{ m_soundName };
```
xezon, line 105: Why do you use curly braces here? If you mean to call `AudioEventRTS(const Utf8String &name);` then please use round braces.

duncanspumpkin: It's just the standard way of calling the constructor in modern c++

xezon: Not really. It is different to () and is not a replacement for it. It will promote to std::initializer list if it can and may not call the function you intended.

duncanspumpkin: It's considerably more safe to use {} initialisation that is why it is recommended.

duncanspumpkin: For reference https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es23-prefer-the--initializer-syntax

tomsons26 (approved):

PR #517, opened 2022-01-22
Implements DX8Wrapper::Set_Light, corrects float check in Set_Light_Environment

tomsons26:


duncanspumpkin:
Is this the source of the tree shading bug? Are there any bugs in the issue list this fixes?

src/w3d/renderer/dx8wrapper.cpp
```diff
@@ -1791,10 +1863,10 @@ void DX8Wrapper::Set_Light_Environment(LightEnvironmentClass *light_env)
                 light.Attenuation0 = 1.0f;
 
                 if (fabsf(light_env->Get_Point_Inner_Radius(light_index) - light_env->Get_Point_Outer_Radius(light_index))
-                    >= 0.00001f) {
-                    light.Attenuation1 = 0.1f / light_env->Get_Point_Inner_Radius(light_index);
-                } else {
+                    < 1e-5) {
```
duncanspumpkin, line 1866: Why invert the logic?

tomsons26: so it follows binary flow

tomsons26:
> Is this the source of the tree shading bug? Are there any bugs in the issue list this fixes?

nah, just cleaning out unmerged things from old branches, this function afict isn't called in game, likely is in w3dviewer i'd guess so yeah needed for later

PR #520, opened 2022-01-24
StateMachine

jonwil:


src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+    return m_machine->Get_Goal_Object();
+}
+
+StateMachine *State::Get_Machine()
+{
+    return m_machine;
+}
+
+void State::Friend_On_Condition(
+    bool (*test)(State *, void *), unsigned int to_state_id, void *user_data, char const *description)
```
xezon, line 61: Please create an alias for this function pointer. It is used in a few places and would increase readability.

`using TestFunctionPtr = bool (*)(State *, void *);`

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+{
+    TransitionInfo info(test, to_state_id, user_data, description);
+    m_transitions.push_back(info);
+}
+
+std::vector<unsigned int> *State::Get_ID_Vector() const
```
xezon, line 67: Please add a comment near function

`// #TODO Refactor ownership of allocated object between caller and callee when possible.`

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+            vector->push_back(info.to_state_id);
+        }
+    }
+
+    return vector;
+}
+
+struct StIncrementer
+{
+    int *value;
```
xezon, line 82: value can be `private`.

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+    m_transitions.push_back(info);
+}
+
+std::vector<unsigned int> *State::Get_ID_Vector() const
+{
+    std::vector<unsigned int> *vector = new std::vector<unsigned int>;
+    vector->push_back(m_successStateID);
+    vector->push_back(m_failureStateID);
+
+    if (!m_transitions.empty()) {
```
xezon: This test is obsolete. The for-range loop below is enough. Even if the original did it like this, remove it.

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+    static int checkfortransitionsnum;
+    StIncrementer incrementer(&checkfortransitionsnum);
+
+    if (checkfortransitionsnum >= 20) {
+        captainslog_dbgassert(checkfortransitionsnum < 20, "checkforsleeptransitionsnum is > 20");
+        return STATE_FAILURE;
+    } else {
+        captainslog_dbgassert(status > STATE_CONTINUE, "Please only pass sleep states here");
+
+        if (!m_transitions.empty()) {
```
xezon: This test is obsolete. The for-range loop below is enough. Even if the original did it like this, remove it.

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+            } else if (m_successStateID == EXIT_MACHINE_WITH_FAILURE) {
+                Get_Machine()->Internal_Set_State(MACHINE_DONE_STATE_ID);
+                return STATE_FAILURE;
+            } else {
+                return Get_Machine()->Internal_Set_State(m_successStateID);
+            }
+        } break;
+
+        case STATE_CONTINUE: {
+            if (!!m_transitions.empty()) {
```
xezon: Condition is double inverse, so its wrong, but also obsolete. Can be removed.

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+    return STATE_CONTINUE;
+}
+
+StateMachine::StateMachine(Object *owner, Utf8String name) :
+    m_owner(owner),
+    m_updateFrame(0),
+    m_defaultStateID(INVALID_STATE_ID),
+    m_defaultStateInited(false),
+    m_currentState(nullptr),
+    m_locked(0)
```
xezon: false

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+#ifdef GAME_DEBUG_STRUCTS
+    ,
```
xezon, line 215: Uh. Is this how clang-format formats this? Looks really ugly.

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+            m_currentState->On_Exit(EXIT_RESET);
+            m_currentState = nullptr;
+        }
+
+        Internal_Clear();
+        StateReturnType state = Internal_Set_State(m_defaultStateID);
+        captainslog_dbgassert(state != STATE_FAILURE, "StateMachine::Reset_To_Default_State() Error setting default state");
+        return state;
+    } else {
+        captainslog_debug("you may not callcall Reset_To_Default_State before Init_Default_State");
```
xezon: callcall

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+            }
+        } else {
+            captainslog_debug("State machine has no current state -- did you remember to call Init_Default_State?");
+            return STATE_FAILURE;
+        }
+    }
+}
+
+void StateMachine::Define_State(
+    unsigned int id, State *state, unsigned int success_id, unsigned int failure_id, StateConditionInfo const *conditions)
```
xezon, line 372: `unsigned int state_id` would be better, because there are 2 other ids too.

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+        state->Friend_On_Condition(conditions->test, conditions->to_state_id, conditions->user_data, nullptr);
+        conditions++;
+    }
+
+    if (m_defaultStateID == INVALID_STATE_ID) {
+        m_defaultStateID = id;
+    }
+}
+
+State *StateMachine::Internal_Get_State(unsigned int id)
```
xezon, line 392: `unsigned int state_id`

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+
+StateReturnType StateMachine::Init_Default_State()
+{
+#ifdef GAME_DEBUG_STRUCTS
+    for (auto &iter : m_stateMap) {
+        State *state = iter.second;
+        unsigned int id = state->Get_ID();
+        std::vector<unsigned int> *vector = state->Get_ID_Vector();
+
+        if (!vector->empty()) {
```
xezon: This condition is obsolete and can be removed.

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+    m_updateFrame = state + frame;
+    return m_currentState->Friend_Check_For_Sleep_Transitions((StateReturnType)(m_updateFrame - frame));
+}
+
+StateReturnType StateMachine::Init_Default_State()
+{
+#ifdef GAME_DEBUG_STRUCTS
+    for (auto &iter : m_stateMap) {
+        State *state = iter.second;
+        unsigned int id = state->Get_ID();
```
xezon: `state_id`

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+        std::vector<unsigned int> *vector = state->Get_ID_Vector();
+
+        if (!vector->empty()) {
+            for (auto &sid : *vector) {
```
xezon: `unsigned int transition_state_id`

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+        for (auto &iter : m_stateMap) {
+            State *state = iter.second;
+            unsigned int state_id = state->Get_ID();
+            xfer->xferUnsignedInt(&state_id);
+
+            captainslog_dbgassert(
+                state_id == state->Get_ID(), "State ID mismatch - %d expected, %d read", state->Get_ID(), state_id);
+
+            if (state == nullptr) {
+                captainslog_dbgassert(state != nullptr, "state was NULL on xfer, trying to heal...");
```
xezon: Assert condition can be set 0 here, because `state` was already tested above.

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+                state = Internal_Get_State(m_defaultStateID);
+            }
+
+            state->Xfer_Snapshot(xfer);
+        }
+    } else {
+        if (m_currentState == nullptr) {
+            captainslog_dbgassert(m_currentState != nullptr, "currentState was NULL on xfer, trying to heal...");
```
xezon: Assert condition can be set 0 here, because `state` was already tested above.

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+
+    xfer->xferObjectID(&m_goalObjectID);
+    xfer->xferCoord3D(&m_goalPosition);
+    xfer->xferBool(&m_locked);
+    xfer->xferBool(&m_defaultStateInited);
+}
+
+unsigned int StateMachine::Get_Current_State_ID() const
+{
+    if (m_currentState) {
```
xezon: != INVALID_STATE_ID

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+    xfer->xferUnsignedInt(&m_updateFrame);
+    xfer->xferUnsignedInt(&m_defaultStateID);
+    unsigned int id = Get_Current_State_ID();
+    xfer->xferUnsignedInt(&id);
+
+    if (xfer->Get_Mode() == XFER_LOAD) {
+        m_currentState = Internal_Get_State(id);
+    }
+
+    bool b = false;
```
xezon: `bool has_state`

src/game/common/statemachine.cpp
```diff
@@ -0,0 +1,716 @@
...
+}
+
+bool StateMachine::Is_In_Attack_State() const
+{
+    return m_currentState == nullptr || m_currentState->Is_Attack();
+}
+
+bool StateMachine::Is_In_Guard_Idle_State() const
+{
+    return m_currentState == nullptr || m_currentState->Is_Guard_Idle();
```
xezon: Hmm. This condition looks very odd. The way its written right now would mean if current state is null, then both Attack State and Guard Idle State are true.

```c++
return m_currentState ? m_currentState->Is_Guard_Idle() : false;
```
would be more logical here.

jonwil: Its how the original code did it.

src/game/common/statemachine.cpp
jonwil: fixed.

src/game/common/statemachine.cpp
jonwil: fixed.

src/game/common/statemachine.cpp
jonwil: fixed.

src/game/common/statemachine.cpp
jonwil: fixed.

src/game/common/statemachine.cpp
jonwil: fixed.

src/game/common/statemachine.cpp
jonwil: fixed.

src/game/common/statemachine.cpp
jonwil: fixed.

src/game/common/statemachine.cpp
jonwil: Yeah that's how its formatted.

src/game/common/statemachine.cpp
jonwil: fixed.

src/game/common/statemachine.cpp
jonwil: fixed.

src/game/common/statemachine.cpp
jonwil: fixed.

src/game/common/statemachine.cpp
jonwil: fixed.

src/game/common/statemachine.cpp
xezon: Please add a comment to investigate.

jonwil: I took another look and it turns out Is_In_Guard_Idle_State was wrong so I fixed that but Is_In_Attack_State is correct so there is nothing to investigate.

tomsons26 (approved):

PR #524, opened 2022-01-30
Implement UpdateModule, BehaviorModule

xezon:


src/game/common/modules/updatemodule.cpp
```diff
@@ -0,0 +1,170 @@
...
+void UpdateModule::Set_Wake_Frame(Object *object, unsigned int frame)
+{
+    unsigned int cur_frame = g_theGameLogic->Get_Frame();
+    g_theGameLogic->Friend_Awaken_Update_Module(object, this, frame + cur_frame);
+}
+
+unsigned int UpdateModule::Frame_To_Sleep_Time(
+    unsigned int frame, unsigned int max_frame1, unsigned int max_frame2, unsigned int max_frame3)
+{
+    captainslog_dbgassert(frame != 0 && max_frame1 != 0 && max_frame2 != 0 && max_frame3 != 0,
```
tomsons26: BFME2 has this assert, it exposes arg names `frame1 != 0 && frame2 != 0 && frame3 != 0 && frame4 != 0`

src/game/common/modules/updatemodule.cpp
```diff
@@ -0,0 +1,170 @@
...
+{
+    return m_updatePhase >> 2;
+}
+
+unsigned int UpdateModule::Get_Wake_Frame() const
+{
+    unsigned int cur_frame = g_theGameLogic->Get_Frame();
+    unsigned int wake_frame = Decode_Frame();
+
+    if (cur_frame >= wake_frame) {
```
tomsons26: flip if to match asm
this also eliminates the else

src/game/common/modules/updatemodule.cpp
```diff
@@ -0,0 +1,170 @@
...
+    m_updatePhase = Get_Update_Phase();
+    m_updatePhase |= (frame << 2);
+}
+
+unsigned int UpdateModule::Decode_Frame() const
+{
+    return m_updatePhase >> 2;
+}
+
+unsigned int UpdateModule::Get_Wake_Frame() const
```
tomsons26: this technically returns UpdateSleepTime, i guess to prevent misuse they used a enum

src/game/common/modules/updatemodule.cpp
```diff
@@ -0,0 +1,170 @@
...
+    unsigned int wake_frame = Decode_Frame();
+
+    if (cur_frame >= wake_frame) {
+        return 1;
+    } else {
+        return wake_frame - cur_frame;
+    }
+}
+
+void UpdateModule::Set_Wake_Frame(Object *object, unsigned int frame)
```
tomsons26: uint was actually UpdateSleepTime

src/game/common/modules/updatemodule.cpp
```diff
@@ -0,0 +1,171 @@
...
+}
+
+SleepyUpdatePhase UpdateModule::Get_Update_Phase() const
+{
+    return SLEEPY_UPDATE_PHASE_2;
+}
+
+BitFlags<DISABLED_TYPE_COUNT> UpdateModule::Get_Disabled_Types_To_Process() const
+{
+    return BitFlags<DISABLED_TYPE_COUNT>();
```
tomsons26: should return DISABLEDMASK_NONE, consult Jonwil if unclear how as he did this kind of thing in a PR before

src/game/common/modules/updatemodule.cpp
xezon: Fixed.

src/game/common/modules/updatemodule.cpp
xezon: Adjusted.

src/game/common/modules/updatemodule.cpp
xezon: Fixed.

src/game/common/modules/updatemodule.cpp
xezon: Added.

src/game/common/modules/updatemodule.h
```diff
@@ -32,14 +33,15 @@ class UpdateModuleData : public BehaviorModuleData
 {
 public:
     UpdateModuleData() {}
-    virtual ~UpdateModuleData() {}
+    virtual ~UpdateModuleData() override {}
 
     static void Build_Field_Parse(MultiIniFieldParse &p) {}
```
tomsons26, line 38: needs to be implemented, uses UpgradeMuxData field parse table

src/game/common/modules/behaviormodule.h
```diff
@@ -116,18 +116,17 @@ class BehaviorModule : public ObjectModule, public BehaviorModuleInterface
     IMPLEMENT_ABSTRACT_POOL(BehaviorModule)
 
 protected:
+    BehaviorModule(Thing *thing, const ModuleData *module_data);
```
tomsons26, line 119: ctors were public

src/game/common/modules/module.h
```diff
@@ -43,18 +43,20 @@ class ModuleData : public SnapShot
     friend class ModuleFactory;
 
 public:
-    ModuleData() {}
+    ModuleData() : m_tagKey(NAMEKEY_INVALID) {}
```
tomsons26: comment as bugfix

src/game/common/modules/behaviormodule.h
xezon: I think it is better to keep these protected unless explicitly required. These modules don't look like they want to be created and destroyed just like that, but are created through the `Friend_New_Module_Instance` functions. See `SwayClientUpdate::Friend_New_Module_Instance` for example.

src/game/common/modules/module.h
xezon: Ok.

src/game/common/modules/updatemodule.cpp
xezon: Added.

xezon:
This review is ready for review, but must not be submitted before #543

src/game/common/bitflags.cpp
```diff
@@ -65,24 +65,6 @@ const char *BitFlags<OBJECT_STATUS_COUNT>::s_bitNamesList[] = {
     nullptr,
 };
 
-template<>
-const char *BitFlags<DISABLED_TYPE_COUNT>::s_bitNamesList[] = {
-    "DEFAULT",
```
duncanspumpkin, line 70: Why is this being moved. All the other instances are in this file

xezon: tomsons26 refered to the existence of this file "DisabledTypes" from the original. So I moved all related things there. Note that some other BitFlag types, such as KindOf, also have their own file. I think that makes sense. They should not be cramped all into BitFlags file. Is not tidy.

src/game/common/modules/updatemodule.cpp
```diff
@@ -0,0 +1,170 @@
...
+    return UpdateSleepTime(1u);
+}
+
+void UpdateModule::Set_Wake_Frame(Object *object, UpdateSleepTime frame)
+{
+    unsigned int cur_frame = g_theGameLogic->Get_Frame();
+    g_theGameLogic->Friend_Awaken_Update_Module(object, this, frame + cur_frame);
+}
+
+unsigned int UpdateModule::Frame_To_Sleep_Time(
```
tomsons26: returns UpdateSleepTime

src/game/logic/system/gamelogic.cpp
```diff
@@ -62,40 +62,117 @@ bool GameLogic::Is_Intro_Movie_Playing()
...
 {
-    captainslog_dbgassert(update, "You may not pass null for sleepy update info");
-    m_sleepingUpdateModules.push_back(update);
-    update->Set_Index_In_Logic(m_sleepingUpdateModules.size() - 1);
+    captainslog_dbgassert(index >= 0 && (unsigned)index < m_sleepingUpdateModules.size(), "bad sleepy idx");
+
+    int next_index = 2 * index + 1;
+    UpdateModule **curr = &m_sleepingUpdateModules[index];
+    UpdateModule **next = &m_sleepingUpdateModules[next_index];
+    UpdateModule **end = &m_sleepingUpdateModules[0] + m_sleepingUpdateModules.size();
```
tomsons26, line 102: m_sleepingUpdateModules[m_sleepingUpdateModules.size()]?

src/game/logic/system/gamelogic.cpp
```diff
@@ -62,40 +62,117 @@ bool GameLogic::Is_Intro_Movie_Playing()
...
+int GameLogic::Rebalance_Parent_Sleepy_Update(int index)
 {
-    captainslog_dbgassert(update >= 0 && (unsigned)update < m_sleepingUpdateModules.size(), "bad sleepy idx");
+    captainslog_dbgassert(index >= 0 && (unsigned)index < m_sleepingUpdateModules.size(), "bad sleepy idx");
 
-    for (int i = ((update + 1) >> 1) - 1; i >= 0; i = ((i + 1) >> 1) - 1) {
-        if (!Compare_Update_Modules(m_sleepingUpdateModules[i], m_sleepingUpdateModules[update])) {
+    for (int i = ((index + 1) >> 1) - 1; i >= 0; i = ((i + 1) >> 1) - 1) {
+
+        UpdateModule *a = m_sleepingUpdateModules[i];
```
tomsons26, line 73: not sure if this will be functionally equal, a is fetched by index in original b is by i

src/game/logic/system/gamelogic.cpp
xezon: That won't work, because that would access element past end and be undefined behaviour.

src/game/logic/system/gamelogic.cpp
xezon: It was already implemented like that. I will double check later.

src/game/common/system/disabledtypes.cpp
```diff
@@ -0,0 +1,38 @@
...
+    "DISABLED_UNMANNED",
+    "DISABLED_UNDERPOWERED",
+    "DISABLED_FREEFALL",
+    "DISABLED_AWESTRUCK",
+    "DISABLED_BRAINWASHED",
+    "DISABLED_SUBDUED",
+    "DISABLED_SCRIPT_DISABLED",
+    "DISABLED_SCRIPT_UNDERPOWERED",
+    nullptr,
+};
```
xezon, line 35: Should probably involve `DISABLED_TYPE_COUNT` here.

src/game/common/modules/behaviormodule.h
```diff
@@ -116,18 +116,20 @@ class BehaviorModule : public ObjectModule, public BehaviorModuleInterface
...
-    BehaviorModule(Thing *thing, ModuleData *module_data);
-
     virtual StealthUpdate *Get_Steath();
     virtual SpyVisionUpdate *Get_Spy_Vision_Update();
 
+    // Snapshot
     virtual void CRC_Snapshot(Xfer *xfer) override;
     virtual void Xfer_Snapshot(Xfer *xfer) override;
     virtual void Load_Post_Process() override;
+    //~Snapshot
```
duncanspumpkin, line 130: as ~ClassName implies destructor I would advise against using this to mark areas. Just use `#pragma region Snapshot` then at least it can be collapsed.

xezon: ~ just marks the end here as a comment. No need to overcomplicate I think.

src/game/logic/system/gamelogic.cpp
xezon: I can check if I can utilize iterators maybe. `m_sleepingUpdateModules[0]` does not look very safe. If `m_sleepingUpdateModules` was empty, this would crash. On new STD we could use `vector.data()` here.

src/game/logic/system/gamelogic.cpp
xezon: I checked. Implementation looks correct. The ordering is changed, but the indexing and pointers used is the exact same.

src/game/common/system/disabledtypes.cpp
xezon: Fixed.

src/game/common/modules/updatemodule.cpp
xezon: Fixed.

src/game/logic/system/gamelogic.cpp
xezon: See fixup commit added to this PR with iterator approach commit. Maybe that is better to look at.

src/game/logic/system/gamelogic.cpp
```diff
@@ -62,40 +62,124 @@ bool GameLogic::Is_Intro_Movie_Playing()
...
+            "Set_Wake_Frame() should not be called from inside the Update(), because it will be ignored, in favor "
+            "of the return code from update");
+
+    } else if (wakeup_frame != module->Decode_Frame()
+        && (0 == cur_frame || module->Decode_Frame() != cur_frame || wakeup_frame != cur_frame + 1)) {
+
+        int index = module->Get_Index_In_Logic();
+
+        if (object->Is_In_List(&m_objList)) {
+            if (index >= 0 && static_cast<size_t>(index) < m_sleepingUpdateModules.size()) {
```
tomsons26: flip

src/game/logic/system/gamelogic.cpp
```diff
@@ -62,40 +62,124 @@ bool GameLogic::Is_Intro_Movie_Playing()
...
+                if (m_sleepingUpdateModules[index] == module) {
+                    module->Encode_Frame(wakeup_frame);
+                    Rebalance_Sleepy_Update(index);
+                } else {
```
tomsons26, line 171: flip

src/game/logic/system/gamelogic.cpp
```diff
@@ -62,40 +62,124 @@ bool GameLogic::Is_Intro_Movie_Playing()
...
+            "of the return code from update");
+
+    } else if (wakeup_frame != module->Decode_Frame()
+        && (0 == cur_frame || module->Decode_Frame() != cur_frame || wakeup_frame != cur_frame + 1)) {
+
+        int index = module->Get_Index_In_Logic();
+
+        if (object->Is_In_List(&m_objList)) {
+            if (index >= 0 && static_cast<size_t>(index) < m_sleepingUpdateModules.size()) {
+                if (m_sleepingUpdateModules[index] == module) {
```
tomsons26: flip

src/game/common/modules/updatemodule.h
```diff
@@ -42,51 +43,60 @@ class UpdateModuleInterface
...
 
+    enum : uint32_t
+    {
+        UPDATE_PHASE_FRAME_MASK = (~(1u << SLEEPY_UPDATE_PHASE_2) + 1u)
+    };
+
 protected:
-    virtual ~UpdateModule() override;
+    UpdateModule(Thing *thing, const ModuleData *module_data);
+    virtual ~UpdateModule() override {}
```
xezon: Implement destructor.

src/game/logic/system/gamelogic.cpp
xezon: Flipped.

src/game/logic/system/gamelogic.cpp
xezon: Flipped.

src/game/common/modules/updatemodule.h
xezon: Assert added.

tomsons26 (approved):

src/game/common/modules/behaviormodule.h
duncanspumpkin: It goes against the style of the rest of the codebase and can cause confusion.

PR #538, opened 2022-02-01
Hook PhysicsBehavior::Get_Forward_Speed_2D function

xezon:


src/game/common/modules/updatemodule.h
```diff
@@ -17,6 +17,7 @@
 #include "always.h"
 #include "behaviormodule.h"
 #include "bitflags.h"
+#include "moduleinfo.h"
```
xezon, line 20: This must be in this review, otherwise compilation fails.

tomsons26 (approved):

PR #539, opened 2022-02-02
Fix outstanding MSVC compiler warnings

xezon:
Closes #443

### DIFF
--- a/src/platform/w3dengine/client/w3ddisplay.cpp
+++ b/src/platform/w3dengine/client/w3ddisplay.cpp
@@ -2161,13 +2161,19 @@ void Reset_D3D_Device(bool restore_assets)
                         true,
                         true,
                         true);
-                    OSVERSIONINFOA VersionInformation;
 
-                    if (GetVersionEx(&VersionInformation)) {
+#if 0
+                    // Original code uses deprecated function.
+                    // VER_PLATFORM_WIN32_WINDOWS refers to Windows 95, 98, ME - none of which we support.
+                    OSVERSIONINFOA VersionInformation;
+                    ZeroMemory(&VersionInformation, sizeof(OSVERSIONINFOA));
+                    VersionInformation.dwOSVersionInfoSize = sizeof(OSVERSIONINFOA);
+                    if (GetVersionExA(&VersionInformation)) {
                         if (VersionInformation.dwPlatformId == VER_PLATFORM_WIN32_WINDOWS) {
                             W3D::Invalidate_Textures();
                         }
                     }
+#endif
                 } else {
                     W3D::Set_Render_Device(W3D::Get_Render_Device(),
                         g_theDisplay->Get_Width(),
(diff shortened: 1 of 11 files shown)


src/platform/w3dengine/client/w3ddisplay.cpp
```diff
@@ -2155,13 +2161,23 @@ void Reset_D3D_Device(bool restore_assets)
...
+                    ZeroMemory(&VersionInformation, sizeof(OSVERSIONINFOA));
+                    VersionInformation.dwOSVersionInfoSize = sizeof(OSVERSIONINFOA);
+                    if (GetVersionExA(&VersionInformation)) {
                         if (VersionInformation.dwPlatformId == VER_PLATFORM_WIN32_WINDOWS) {
                             W3D::Invalidate_Textures();
                         }
                     }
+#else
+                    // Replacement code uses supported Windows API. Available with Windows 2000 and higher.
+                    if (!IsWindowsXPOrGreater()) {
```
xezon: Actually, this is stupid, because it would not compile on earlier windows to begin with.

xezon: Fixed.

src/platform/w3dengine/client/w3ddisplay.cpp
```diff
@@ -2155,13 +2155,19 @@ void Reset_D3D_Device(bool restore_assets)
                         true,
                         true,
                         true);
-                    OSVERSIONINFOA VersionInformation;
 
-                    if (GetVersionEx(&VersionInformation)) {
+#if 0
```
jonwil, line 2165: I still say we don't even need this code at all.
It says "if OS = an os Thyme will never run on, do x". So lets just ditch the whole lot instead of keeping it around for no good reason.

xezon: For now, I would like to keep it in place to test it later again. We still have a critical issue with ALT + Tab in the game, and anything related to Direct X Reset needs to be looked at in that regard. The deactivation of the code is appropriate for now.

Note that this function is called in game.

src/w3d/renderer/meshgeometry.cpp
```diff
@@ -262,7 +262,7 @@ void MeshGeometryClass::Set_Name(const char *newname)
     }
 
     if (newname) {
-        m_meshName = new ShareBufferClass<char>(strlen(newname) + 1);
+        m_meshName = new ShareBufferClass<char>((int)strlen(newname) + 1);
         strcpy(m_meshName->Get_Array(), newname);
```
duncanspumpkin, line 267: Use static casts

xezon: Fixed.

xezon:
https://stackoverflow.com/questions/45995784/how-to-set-compiler-options-with-cmake-in-visual-studio-2017

jonwil:
This PR is obsolete. Current code builds with no warnings (at least none that matter). Closing.

PR #547, opened 2022-02-06
Fix several animation issues and improve related code

### LINKED ISSUE #535: China Red Guard Infantry does not go into firing animation when attacking from idle stance
China Red Guard Infantry does not go into firing animation when attacking from idle stance. It still shoots however.

shot_20220131_223154_4

xezon:
Fixes #535
Fixes memory leaks
Fixes uninitialized variables
Refactors code to look closer to original and better to read

### DIFF
--- a/src/w3d/renderer/motchan.cpp
+++ b/src/w3d/renderer/motchan.cpp
@@ -196,14 +196,14 @@ bool TimeCodedMotionChannelClass::Load_W3D(ChunkLoadClass &cload)
         return false;
     }
 
-    m_numTimeCodes = chan.NumTimeCodes;
-    m_vectorLen = chan.VectorLen;
     m_type = chan.Flags;
+    m_vectorLen = chan.VectorLen;
+    m_numTimeCodes = chan.NumTimeCodes;
+    m_packetSize = chan.VectorLen + 1;
+    m_lastTimeCodeIdx = (chan.VectorLen + 1) * (m_numTimeCodes - 1);
     m_pivotIdx = chan.Pivot;
-    m_packetSize = m_vectorLen + 1;
     m_cachedIdx = 0;
-    m_lastTimeCodeIdx = m_packetSize * (m_numTimeCodes - 1);
-    m_data = new unsigned int[((chunk_size / 4) + 1)];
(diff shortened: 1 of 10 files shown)


xezon:
I tested animations on a wide array of infantry, vehicles and buildings. Other than the SCUD Launcher which is still broken, it looked all good.

src/w3d/renderer/motchan.cpp
```diff
@@ -196,14 +196,14 @@ bool TimeCodedMotionChannelClass::Load_W3D(ChunkLoadClass &cload)
         return false;
     }
 
-    m_numTimeCodes = chan.NumTimeCodes;
-    m_vectorLen = chan.VectorLen;
     m_type = chan.Flags;
+    m_vectorLen = chan.VectorLen;
+    m_numTimeCodes = chan.NumTimeCodes;
+    m_packetSize = chan.VectorLen + 1;
+    m_lastTimeCodeIdx = (chan.VectorLen + 1) * (m_numTimeCodes - 1);
```
tomsons26, line 203: can remain m_packetSize * (m_numTimeCodes - 1)

src/w3d/renderer/motchan.cpp
```diff
@@ -287,23 +287,22 @@ Quaternion TimeCodedMotionChannelClass::Get_Quat_Vector(float frame_idx)
...
         setvec[0] = 0.0f;
         setvec[1] = 0.0f;
         setvec[2] = 0.0f;
         setvec[3] = 1.0f;
+    } else {
+        setvec[0] = 0.0f;
     }
 }
 
 unsigned long TimeCodedMotionChannelClass::Get_Index(unsigned int timecode)
```
tomsons26, line 300: unsigned long should be purged from these files

src/w3d/renderer/motchan.cpp
xezon: Yes. But I think writing it this way is much cleaner, because this way it only uses the chan struct to set class members, and not other class members to set class members, thus removing any kind of order dependencies.

Compiler can optimize it any way it wants, and it will likely make it one `chan.VectorLen + 1` only.

tomsons26 (approved):

PR #549, opened 2022-02-07
Fix shroud bug in W3DModelDraw

### LINKED ISSUE #542: Civilian structures do not draw in shroud
Civilian structures do not draw in shroud. Tested on map Tournament Desert (2).

shot_20220203_235211_1

xezon:
Fixes #542

### DIFF
--- a/src/game/client/draw/w3dmodeldraw.cpp
+++ b/src/game/client/draw/w3dmodeldraw.cpp
@@ -2596,9 +2596,10 @@ int W3DModelDraw::Get_Current_Bone_Positions(
 void W3DModelDraw::React_To_Transform_Change(const Matrix3D *matrix, const Coord3D *pos, float angle)
 {
     if (m_renderObject) {
-        Matrix3D m(Get_Drawable()->Get_Transform_Matrix());
-        Adjust_Transform_Mtx(m);
-        m_renderObject->Set_Transform(m);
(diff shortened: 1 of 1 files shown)


jonwil (approved):
Fix looks correct to me, approved.

src/game/client/draw/w3dmodeldraw.cpp
```diff
@@ -2596,9 +2596,10 @@ int W3DModelDraw::Get_Current_Bone_Positions(
 void W3DModelDraw::React_To_Transform_Change(const Matrix3D *matrix, const Coord3D *pos, float angle)
 {
     if (m_renderObject) {
-        Matrix3D m(Get_Drawable()->Get_Transform_Matrix());
-        Adjust_Transform_Mtx(m);
-        m_renderObject->Set_Transform(m);
+        const Matrix3D *dtm = Get_Drawable()->Get_Transform_Matrix();
+        Matrix3D rtm(*dtm);
```
duncanspumpkin, line 2600: how did this fix it?

duncanspumpkin: I take it there was a different constructor being called when passed a pointer instead of a reference? That suggests there could easily be other issues elsewhere in the codebase and a massive footgun

jonwil: We already checked for other possible similar issues and found none. And we are looking at warnings that will prevent such an issue in the future.

PR #550, opened 2022-02-07
Forces warning 4800 as a error. Fixes cases where it occurs.

tomsons26:
Issues just like and similar to #542 shouldn't be possible with this set.


xezon:
Any reason for the C style casts as opposed to static casts?

xezon (approved):

PR #558, opened 2022-02-11
Implement more of drawable

jonwil:


src/game/client/draw/drawable.cpp
```diff
@@ -3069,3 +3083,211 @@ void Drawable::Apply_Physics_Xform(Matrix3D *mtx)
...
+
+    Clear_Status_Bit(DRAWABLE_STATUS_AMBIENT_LIGHT_LOCKED);
+}
+
+void Drawable::Draw(View *view)
+{
+    if (!Check_Draw_Bit(DRAW_BIT_WEAPON_BONUS)) {
+        if (Get_Object() != nullptr && Get_Object()->Is_Effectively_Dead()) {
+            m_stealthEmissiveScale = 0.0f;
+        } else if (m_stealthEmissiveScale <= 0.001f) {
```
tomsons26: flip

src/game/client/draw/drawable.cpp
```diff
@@ -3069,3 +3083,211 @@ void Drawable::Apply_Physics_Xform(Matrix3D *mtx)
...
+{
+    if (m_tintColorEnvelope == nullptr) {
+        m_tintColorEnvelope = new TintEnvelope();
+    }
+
+    if (color) {
+        m_tintColorEnvelope->Play(color, attack_frames, decay_frames, peak_frames);
+    } else {
+        RGBColor c;
+        c.Set_From_Int(0xFFFFFFFF);
```
tomsons26, line 3227: this is a ctor based on GlobalData ctor invoking it

src/game/client/draw/drawable.cpp
```diff
@@ -3069,3 +3083,211 @@ void Drawable::Apply_Physics_Xform(Matrix3D *mtx)
         }
     }
 }
+
+void Drawable::Set_Fully_Obscured_By_Shroud(bool fully_obscured)
+{
+    for (DrawModule **i = Get_Draw_Modules(); *i != nullptr; i++) {
```
tomsons26: needs to check m_fullyObscuredByShroud != fully_obscured before

src/game/client/draw/drawable.cpp
```diff
@@ -3069,3 +3083,211 @@ void Drawable::Apply_Physics_Xform(Matrix3D *mtx)
...
+        m_terrainDecalOpacity = 0.0f;
+    } else {
+        DrawModule **modules = Get_Draw_Modules();
+
+        if (*modules != nullptr) {
+            if (m_terrainDecalFadeTarget2 != 0.0f) {
+                (*modules)->Set_Terrain_Decal_Opacity(m_terrainDecalOpacity);
+            }
+
+            if (m_terrainDecalFadeTarget2 >= 0.0f || m_terrainDecalOpacity > 0.0f) {
```
tomsons26: flip

src/game/client/draw/drawable.cpp
```diff
@@ -3069,3 +3083,211 @@ void Drawable::Apply_Physics_Xform(Matrix3D *mtx)
...
+                m_terrainDecalFadeTarget2 = 0.0f;
+                m_terrainDecalOpacity = 0.0f;
+                Set_Terrain_Decal(TERRAIN_DECAL_8);
+            }
+        }
+    }
+
+    if (m_expirationDate && frame >= m_expirationDate) {
+        captainslog_dbgassert(object == nullptr, "Drawables with Objects should not have expiration dates!");
+        g_theGameClient->Destroy_Drawable(this);
```
tomsons26, line 3143: early return was here, thus eliminating the else below

src/game/client/draw/drawable.h
```diff
@@ -167,6 +167,15 @@ class Drawable : public Thing, public SnapShot
         float m_totalZ;
     };
 
+    enum DrawBits
+    {
+        DRAW_BIT_DISABLED = 1,
```
tomsons26, line 172: these are based on?

src/game/client/draw/drawable.cpp
```diff
@@ -70,6 +71,11 @@ static const char *s_theDrawableIconNames[] = { "DefaultHeal",
     "CarBomb",
     nullptr };
 
+const RGBColor s_darkGreyDisabledColor = RGBColor{ -0.5f, -0.5f, -0.5f };
+const RGBColor s_subdualDamageColor = RGBColor{ -0.2f, -0.2f, 0.80000001f };
+const RGBColor s_frenzyColor = RGBColor{ 0.2f, -0.2f, -0.2f };
+const RGBColor s_frenzyColorInfantry = RGBColor{ 0.0f, -0.69999999f, -0.69999999f };
```
xezon, line 77: How are these floats supposed to be read?

src/game/client/draw/drawable.cpp
```diff
@@ -174,11 +180,17 @@ void Drawable::Set_Position(Coord3D const *pos)
 
 ClientUpdateModule *Drawable::Find_Client_Update_Module(NameKeyType key)
 {
-#ifdef GAME_DLL
-    return Call_Method<ClientUpdateModule *, const Drawable, NameKeyType>(PICK_ADDRESS(0x00477A80, 0x007C9F40), this, key);
-#else
+    ClientUpdateModule **modules = Get_Client_Update_Modules();
+
+    if (modules != nullptr) {
+        while (*modules != nullptr) {
```
xezon, line 186: It looks like this loop never advances, thus would get stuck if keys do not match. Is this meant to be a `if` instead?

src/game/client/draw/drawable.cpp
```diff
@@ -3069,3 +3083,211 @@ void Drawable::Apply_Physics_Xform(Matrix3D *mtx)
...
+        }
+
+        if (m_selectionColorEnvelope != nullptr) {
+            m_selectionColorEnvelope->Update();
+        }
+
+        if (m_ambientSound != nullptr) {
+            if (m_ambientSoundEnabled) {
+                if (m_ambientSoundFromScriptEnabled) {
+                    if (!m_ambientSound->m_event.Get_Event_Name().Is_Empty()
```
xezon: Maybe put a few more conditions into same `if` line ?

src/game/client/draw/drawable.cpp
jonwil: Its not, confirmed that.

src/game/client/draw/drawable.cpp
jonwil: fixed.

src/game/client/draw/drawable.cpp
jonwil: fixed.

src/game/client/draw/drawable.cpp
jonwil: fixed.

src/game/client/draw/drawable.cpp
jonwil: fixed.

src/game/client/draw/drawable.cpp
jonwil: The original code looks like this. Definitely a while loop and definitely no increment.

src/game/client/draw/drawable.cpp
jonwil: Since RGBColor has no constructor, this code will initialize it via an initializer list and fill its 3 variables with the 3 floats.

src/game/client/draw/drawable.h
jonwil: The places that I saw calling clearDrawBit and setDrawBit.

src/game/client/draw/drawable.cpp
jonwil: fixed.

src/game/client/draw/drawable.cpp
xezon: Please put a `#TODO investigate infinite loop` for this. It looks very broken.

jonwil: Unless we have evidence to suggest that what is being done is a bug (rather than something intentional) we should not diverge from how the original did it IMO.

jonwil: After checking BFME2 and seeing the issue fixed there, its clear the way it works in ZH is unintentional and its been fixed.

xezon:
Looks good in game.

xezon (approved):

PR #560, opened 2022-02-12
Fix linux compile warnings

xezon:
Fixes a few linux warnings

warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
warning: comparison between ‘enum DrawableID’ and ‘enum ObjectID’ [-Wenum-compare]

### DIFF
--- a/src/game/client/draw/w3dmodeldraw.cpp
+++ b/src/game/client/draw/w3dmodeldraw.cpp
@@ -1879,7 +1879,7 @@ bool W3DModelDraw::Update_Bones_For_Client_Particle_Systems()
     return true;
 }
 
-char *s_terrainDecalTextureName[] = { "TERRAIN_DECAL_DEMORALIZED_OBSOLETE",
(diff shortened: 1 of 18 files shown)


src/game/client/draw/w3dmodeldraw.cpp
```diff
@@ -1879,7 +1879,7 @@ bool W3DModelDraw::Update_Bones_For_Client_Particle_Systems()
     return true;
 }
 
-char *s_terrainDecalTextureName[] = { "TERRAIN_DECAL_DEMORALIZED_OBSOLETE",
+const char *const s_terrainDecalTextureName[] = { "TERRAIN_DECAL_DEMORALIZED_OBSOLETE",
```
duncanspumpkin: Could make this constexpr to further cement it's fixed value ness

xezon: As far as I am aware `constexpr` or not makes no practical difference here.

duncanspumpkin (approved):

PR #567, opened 2022-02-14
Fix linux compilation issues

feliwir:
- [x] Use const char* where required
- [x] Add missing cctype header for `isspace` & `isdigit` usage
- [x] Remove usage of undefined/unavailable `_isnan` function

### DIFF
--- a/src/game/client/draw/w3dmodeldraw.cpp
+++ b/src/game/client/draw/w3dmodeldraw.cpp
@@ -1879,7 +1879,7 @@ bool W3DModelDraw::Update_Bones_For_Client_Particle_Systems()
     return true;
 }
 
-const char *s_terrainDecalTextureName[] = { "TERRAIN_DECAL_DEMORALIZED_OBSOLETE",
+const char *const s_terrainDecalTextureName[] = { "TERRAIN_DECAL_DEMORALIZED_OBSOLETE",
     "EXHorde",
     "EXHorde_UP",
     "EXHordeB",
--- a/src/game/client/draw/w3dmodeldraw.cpp
+++ b/src/game/client/draw/w3dmodeldraw.cpp
@@ -1879,7 +1879,7 @@ bool W3DModelDraw::Update_Bones_For_Client_Particle_Systems()
     return true;
 }
 
-char *s_terrainDecalTextureName[] = { "TERRAIN_DECAL_DEMORALIZED_OBSOLETE",
+const char *s_terrainDecalTextureName[] = { "TERRAIN_DECAL_DEMORALIZED_OBSOLETE",
     "EXHorde",
     "EXHorde_UP",
     "EXHordeB",
--- a/src/game/common/globaldata.cpp
+++ b/src/game/common/globaldata.cpp
@@ -728,22 +728,24 @@ GlobalData::GlobalData()
     }
 
     // crc the skirmish and multiplayer scripts.
-    File *scriptfile = g_theFileSystem->Open("Data/Scripts/SkirmishScripts.scb", File::BINARY | File::READ);
-    uint8_t buffer[CRC_BUFFER_SIZE];
     int read = 0;
+    uint8_t buffer[CRC_BUFFER_SIZE];
+    File *scriptfile = g_theFileSystem->Open("Data/Scripts/SkirmishScripts.scb", File::BINARY | File::READ);
+    if (scriptfile) {
+        while ((read = scriptfile->Read(buffer, sizeof(buffer))) != 0) {
+            crc.Compute_CRC(buffer, read);
+        }
 
-    while ((read = scriptfile->Read(buffer, sizeof(buffer))) != 0) {
-        crc.Compute_CRC(buffer, read);
+        scriptfile->Close();
     }
-
-    scriptfile->Close();
     scriptfile = g_theFileSystem->Open("Data/Scripts/MultiplayerScripts.scb", File::BINARY | File::READ);
+    if (scriptfile) {
+        while ((read = scriptfile->Read(buffer, sizeof(buffer))) != 0) {
+            crc.Compute_CRC(buffer, read);
+        }
 
-    while ((read = scriptfile->Read(buffer, sizeof(buffer))) != 0) {
-        crc.Compute_CRC(buffer, read);
+        scriptfile->Close();
     }
-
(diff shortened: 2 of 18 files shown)


duncanspumpkin:
FYI #560 covers a few of these.

src/game/common/globaldata.cpp
```diff
@@ -728,23 +728,25 @@ GlobalData::GlobalData()
     }
 
     // crc the skirmish and multiplayer scripts.
-    File *scriptfile = g_theFileSystem->Open("Data/Scripts/SkirmishScripts.scb", File::BINARY | File::READ);
-    uint8_t buffer[CRC_BUFFER_SIZE];
     int read = 0;
+    uint8_t buffer[CRC_BUFFER_SIZE];
+    File *scriptfile = g_theFileSystem->Open("Data/Scripts/SkirmishScripts.scb", File::BINARY | File::READ);
+    if (scriptfile) {
```
tomsons26, line 734: != nullptr for all, looks good otherwise
these checks were in original but we missed them somehow so good catch

src/game/common/system/mempoolobj.h
```diff
@@ -104,10 +104,11 @@ public: \
 #define IMPLEMENT_POOL(classname) IMPLEMENT_NAMED_POOL(classname, classname);
 
 #define IMPLEMENT_ABSTRACT_POOL(classname) \
-private: \
+protected: \
```
tomsons26, line 107: whats the reason for this?
getObjectMemoryPool is confirmed as private by a linker map left in Generals Demo, and others we chose to make private

feliwir: Well, it is no valid C++. A delete operator needs to be visible to the compiler.
It results in errors like these: `error: ‘static void Thing::operator delete(void*)’ is private within this context`

The only reason this compiles under WIndows is probably MSVC being a non-compliant C++ compiler :)

feliwir:
> FYI #560 covers a few of these.

Oh nice, didn't see these. However i also compiled without `-DSTANDALONE` catching a few more errors

src/game/common/globaldata.cpp
feliwir: Fixed.
I catched this by running the thyme executable under Linux - I expected it to crash - however crashes due to missing files can & should be avoided.

xezon: This file should always be there on a typical install. I wonder, is AI broken when the file is absent?

Comment this with `// #BUGFIX` if this is no original behavior.

src/game/common/system/mempoolobj.h
xezon: protected looks good.

src/game/common/globaldata.cpp
tomsons26: its original behavior, we just missed the ifs

src/game/common/system/mempoolobj.h
duncanspumpkin: Nothing should be constructing an abstract pool though. Isn't this an indication that something else is wrong

xezon: @feliwir Where did the error originate from?

feliwir: From every class that inherits from a class using that macro e.g.:
```
Object -> Thing
AIUpdateInterface -> UpdateModule
etc.
```

duncanspumpkin: My understanding is that it's perfectly legal to have a private operator be overloaded to public in a derived. I even made a minimal test case and gcc/clang appears to compile it fine. I'm just scared that this change will lead to cases where previously there was a compiler error and now there is only a runtime error.
https://godbolt.org/z/Wz15edWfd

feliwir: @duncanspumpkin  you got the error wrong. The operator is not overloaded by the derived class at all. Resulting in exactly the error i got: https://godbolt.org/z/bGvabq91K

duncanspumpkin: But no derived should exist that don't have a non abstract pool macro unless they are also an abstract pool. So this does suggest that there are other issues. Can you list all the types it errors on. UpdateModule/Thing for example I don't think should ever be created except through a derived.

duncanspumpkin: Fwiw msvc also errors on hidden private in godbolt in your test so i think it's a problem in some ifdef code.

feliwir: Yes, in theory there would be some place where the delete operator is called (which would not be allowed), however GCC does not show that place. I'm trying to compile with `clang` to get a better error description, however there's another chain of issues when trying to compile with clang

xezon: Could you test if that linux error still happens with this change
* #537

Note I changed `void *operator new(size_t size)` operator there, which may help prevent the error.

feliwir: @xezon  it doesn't work with these changes. I isolated the issue down to the virtual destructors:
https://godbolt.org/z/dsMrerv78

This compiles for MSVC, but not for GCC/Clang. GCC/Clang try to calling the destructor/delete operator of the base class (it assumes that delete would be used for that)

See https://stackoverflow.com/questions/51083176/gcc-produces-hidden-delete-operator-call-despite-the-constructor-being-marked-a

MSVC seems to generate these calls to the delete operator aswell - however it doesn't care about visibility

src/game/client/draw/w3dmodeldraw.cpp
```diff
@@ -1879,7 +1879,7 @@ bool W3DModelDraw::Update_Bones_For_Client_Particle_Systems()
     return true;
 }
 
-char *s_terrainDecalTextureName[] = { "TERRAIN_DECAL_DEMORALIZED_OBSOLETE",
+const char *s_terrainDecalTextureName[] = { "TERRAIN_DECAL_DEMORALIZED_OBSOLETE",
```
xezon: `const char *const`

src/game/common/system/mempoolobj.h
duncanspumpkin: So in conclusion not much we can do about it apart from switch it to protected.

xezon (approved):

PR #568, opened 2022-02-15
Fixes up CommandList and GameMessageList.

tomsons26:


xezon (approved):
Logically looks good.

PR #572, opened 2022-02-16
SDL2 backend

feliwir:
Use SDL2 to show a window (required for *nix systems)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.19.0)
+cmake_minimum_required(VERSION 3.18.0)
```
xezon: 3.19 was set for interface library feature. It is used for MSVC. Did you test this?
https://cmake.org/cmake/help/latest/command/add_library.html#interface-libraries

CMakeLists.txt
```diff
@@ -223,6 +226,18 @@ if(USE_GAMEMATH)
     endif()
 endif()
 
+if(USE_SDL2)
+    include(FetchContent)
+
+    FetchContent_Declare(
+        SDL2
+        GIT_REPOSITORY https://github.com/libsdl-org/SDL.git
```
xezon: What happens if `libsdl-org/SDL.git` goes offline? Will Thyme break? If yes, what can we do to not have external dependency?

src/game/main.cpp
```diff
@@ -150,24 +154,59 @@ void Check_Windowed(int argc, char *argv[])
...
             ++i;
             g_yPos = atoi(argv[i]);
-            g_yPos = std::clamp(g_yPos, 0, (int)(Res.bottom - 600)); // Prevent negative values
         }
     }
+#ifdef PLATFORM_WINDOWS
+    RECT Res;
+    GetWindowRect(GetDesktopWindow(), &Res);
+    g_xPos = std::clamp(g_xPos, 0, (int)(Res.right - WIN_WIDTH)); // Prevent negative values
+    g_yPos = std::clamp(g_yPos, 0, (int)(Res.bottom - WIN_HEIGHT)); // Prevent negative values
```
xezon: Now the clamping is Windows specific behaviour, and this code will not run on other platform. This seems not ok.

src/game/main.cpp
```diff
@@ -196,8 +234,8 @@ void Create_Window()
 
     RegisterClassA(&WndClass);
 
-    Rect.right = 800;
-    Rect.bottom = 600;
+    Rect.right = window_width;
+    Rect.bottom = window_height;
```
xezon: Does this compile? I don't see declaration of these.

src/game/main.cpp
```diff
@@ -253,7 +291,6 @@ void Create_Window()
     ShowWindow(app_hwnd, show_cmd);
     UpdateWindow(app_hwnd);
 
-    g_applicationHInstance = app_hinstance;
```
xezon: Moving the original variables around will make it more difficult to compare function against original.

src/game/main.cpp
```diff
@@ -150,24 +154,59 @@ void Check_Windowed(int argc, char *argv[])
...
     }
+#ifdef PLATFORM_WINDOWS
+    RECT Res;
+    GetWindowRect(GetDesktopWindow(), &Res);
+    g_xPos = std::clamp(g_xPos, 0, (int)(Res.right - WIN_WIDTH)); // Prevent negative values
+    g_yPos = std::clamp(g_yPos, 0, (int)(Res.bottom - WIN_HEIGHT)); // Prevent negative values
 #endif
 }
 
 void Create_Window()
```
xezon: May I suggest to leave `Create_Window` as is and simply create a new function `Create_Window_SDL2` that is called instead of this one? This way we do not need to touch this function at all and leave it in pristine state.

CMakeLists.txt
feliwir: Didn't mean to downgrade - will revert (i only had 3.18 locally)

CMakeLists.txt
feliwir: GitHub & SDL2 are both big organization and i doub't they'll just go "offline". We already rely on GitHub repos from AssemblyArmada. When organizations are renamed GitHub will auto redirect. 

As an alternative the repo could be forked, but as i said i don't see a requirement for that

src/game/main.cpp
feliwir: We currently don't use these 2 options for SDL. I just center the window always (SDL_WINDOWPOS_CENTERED). I'll add an else block with a TODO. 

However I'm afraid all of this will require more tinkering for multimonitor setups in SDL2 - which screen to spawn the window on? Which monitor to use for clamping?

That's why i was always centering the window atm

src/game/main.cpp
feliwir: Yes, that would maybe help readability. Will do so

src/game/main.cpp
feliwir: just fixed this - no it didn't :)

src/game/main.cpp
feliwir: Yes, will be reverted with a 2nd function

CMakeLists.txt
xezon: TheAssemblyArmada dependency is ok, because that belongs to us.

Maybe we should fork SDL2 into TheAssemblyArmada and then link to that? This way our project is self contained and has no external dependency. And if we need newer SDL2, then we simply merge down the new changes.

src/game/main.cpp
xezon: Maybe same here, don't edit original function, and just create a new one for SDL2.

src/game/main.cpp
```diff
@@ -269,6 +276,43 @@ void Create_Window()
 #endif
 }
 
+void Create_Window_SDL2()
+{
+#ifdef BUILD_WITH_SDL2
```
xezon: How many SDL2 function do we expect to be created from here on? I think this is fine for now, but depending on the SDL2 volume in the future, we may want to not have every SDL2 function have this `#ifdef BUILD_WITH_SDL2`.

src/game/main.cpp
```diff
@@ -269,6 +276,43 @@ void Create_Window()
...
+    bool is_windowed = g_gameIsWindowed;
+    SDL_Window *window = NULL;
```
xezon: You can move this declaration down to the first line it is used.

src/game/main.cpp
```diff
@@ -269,6 +276,43 @@ void Create_Window()
...
+
+    if (SDL_Init(SDL_INIT_VIDEO) < 0) {
+        captainslog_error("SDL could not initialize!");
+        return;
```
xezon: Shouldn't this function return false and then trickle up to terminate the game process? Or is the process expected to continue after this error?

src/game/main.cpp
```diff
@@ -150,15 +154,18 @@ void Check_Windowed(int argc, char *argv[])
         if (strcasecmp(argv[i], "-xpos") == 0) {
             ++i;
             g_xPos = atoi(argv[i]);
-            g_xPos = std::clamp(g_xPos, 0, (int)(Res.right - 800)); // Prevent negative values
         }
 
         if (strcasecmp(argv[i], "-ypos") == 0) {
             ++i;
             g_yPos = atoi(argv[i]);
-            g_yPos = std::clamp(g_yPos, 0, (int)(Res.bottom - 600)); // Prevent negative values
```
xezon, line 170: Did you test this change on windows? I suspect the window positioning now behaves differently than it did before. Before it would clamp only when -xpos or -ypos command line arguments were set. Now it always does it. This looks like it goes beyond the intent of this change, as you change Windows Window behavior and not just add SDL2 Window.

src/game/main.cpp
feliwir: I'd probably want to put these into different cpp files at some point and compile these conditionally. However since there isn't too much WinAPI code in Thyme yet i also can't replace that much

src/game/main.cpp
feliwir: Yeah, i just noticed it's not set to invalidPos after this anymore - I'll revert this change

src/game/main.cpp
```diff
@@ -154,19 +158,19 @@ void Check_Windowed(int argc, char *argv[])
         if (strcasecmp(argv[i], "-xpos") == 0) {
             ++i;
             g_xPos = atoi(argv[i]);
+#ifdef MES_PLATFORM_WINDOWS
```
xezon: `MES_PLATFORM_WINDOWS` ?

src/game/main.cpp
```diff
@@ -154,19 +158,19 @@ void Check_Windowed(int argc, char *argv[])
...
+            g_xPos = std::clamp(g_xPos, 0, (int)(Res.right - WIN_WIDTH)); // Prevent negative values
+#endif
         }
 
         if (strcasecmp(argv[i], "-ypos") == 0) {
             ++i;
             g_yPos = atoi(argv[i]);
+            #ifdef PLATFORM_WINDOWS
```
xezon: Remove indentation.

src/game/main.cpp
feliwir: Wops, i did use macros from work by accident :)

src/game/main.cpp
feliwir: there is no indent for me locally

src/game/main.cpp
```diff
@@ -269,6 +280,44 @@ void Create_Window()
...
+    HINSTANCE app_hinstance = GetModuleHandle(nullptr);
+    g_applicationHInstance = app_hinstance;
+    // Get native window handle
+    SDL_SysWMinfo wmInfo;
+    SDL_VERSION(&wmInfo.version);
+    SDL_GetWindowWMInfo(window, &wmInfo);
+    g_applicationHWnd = wmInfo.info.win.window;
+#endif
+#endif
+    return true;
```
xezon, line 562: This should probably return false when not `BUILD_WITH_SDL2`

xezon (dismissed):
Code looks ok to me.

src/platform/input/sdl2mouse.h
```diff
@@ -0,0 +1,51 @@
...
+#include "mouse.h"
+
+#include <SDL.h>
+
+class SDL2Mouse : public Mouse
+{
+public:
+    SDL2Mouse();
+
+    virtual ~SDL2Mouse(){};
```
xezon: override

src/platform/input/sdl2mouse.cpp
```diff
@@ -0,0 +1,186 @@
...
+    std::memset(s_loadedCursors, 0, sizeof(s_loadedCursors));
+}
+
+void SDL2Mouse::Load_Cursor(const char *file_location, uint32_t load_index, uint32_t load_direction) const
+{
+    if (s_loadedCursors[load_index][load_direction] != NULL) {
+        // Cursor at this position already loaded. Ignoring subsequent load.
+        return;
+    }
+    // TODO: this should parse the ANI file and load the cursor correctly
```
xezon: What is this comment for? Did you test this? Does it work?

src/platform/input/sdl2mouse.cpp
```diff
@@ -0,0 +1,186 @@
...
+SDL_Cursor *SDL2Mouse::s_loadedCursors[CURSOR_COUNT][8];
+
+SDL2Mouse::SDL2Mouse() : m_nextFreeIndex(0), m_nextGetIndex(0) {}
+
+void SDL2Mouse::Init()
+{
+    Mouse::Init();
+
+    std::memset(m_eventBuffer, 0, sizeof(m_eventBuffer));
+    std::memset(s_loadedCursors, 0, sizeof(s_loadedCursors));
```
xezon: In `Win32Mouse` class the memset is in constructor. Any reason this is put in `Init` function here instead?

src/platform/input/sdl2mouse.cpp
```diff
@@ -0,0 +1,186 @@
...
+    io->middle_state = 0;
+    io->left_state = 0;
+    io->right_frame = 0;
+    io->middle_frame = 0;
+    io->left_frame = 0;
+    io->wheel_delta = 0;
+    io->pos.y = 0;
+    io->pos.x = 0;
+
+    // TODO: mousewheel handling is a bit difficult - how to calculate a delta similar to win32mouse?
```
xezon: Also double clicks and middle mouse button handling is missing here.

src/platform/input/sdl2mouse.cpp
```diff
@@ -0,0 +1,186 @@
...
+                    io->pos.y = event.button.y;
+                    break;
+                case SDL_BUTTON_RIGHT:
+                    io->right_state = MouseIO::MouseState::MOUSE_STATE_UP;
+                    io->right_frame = frame_num;
+                    io->pos.x = event.button.x;
+                    io->pos.y = event.button.y;
+                    break;
+                default:
+                    captainslog_debug("Translate_Event: Unknown SDL mouse up event [%d,%d,%d]\n", event.type);
```
xezon: format expects 3 arguments

src/platform/input/sdl2mouse.cpp
```diff
@@ -0,0 +1,186 @@
...
+                    io->pos.y = event.button.y;
+                    break;
+                case SDL_BUTTON_RIGHT:
+                    io->right_state = MouseIO::MouseState::MOUSE_STATE_DOWN;
+                    io->right_frame = frame_num;
+                    io->pos.x = event.button.x;
+                    io->pos.y = event.button.y;
+                    break;
+                default:
+                    captainslog_debug("Translate_Event: Unknown SDL mouse down event [%d,%d,%d]\n", event.type);
```
xezon: format expects 3 arguments

src/platform/input/sdl2mouse.cpp
```diff
@@ -0,0 +1,186 @@
...
+    // TODO: this should parse the ANI file and load the cursor correctly
+    SDL_Surface *surface = SDL_LoadBMP(file_location);
+    if (!surface) {
+        // Failed to load surface
+        return;
+    }
+    SDL_Cursor *cursor = SDL_CreateColorCursor(surface, 0, 0);
+    if (!cursor) {
+        // Failed to create cursor
+        return;
```
xezon: Since there is no error handling for this function, I suggest to at least log some errors.

src/platform/input/sdl2mouse.cpp
feliwir: It does not work. ANI files need to be parsed and loaded into multiple bitmaps, which are switched dynamically. Animated cursors only exist on windows, so these cursor needs to be set for each frame individually.

src/platform/input/sdl2mouse.cpp
```diff
@@ -40,11 +41,13 @@ void SDL2Mouse::Load_Cursor(const char *file_location, uint32_t load_index, uint
     SDL_Surface *surface = SDL_LoadBMP(file_location);
     if (!surface) {
         // Failed to load surface
+        captainslog_error("Load_Cursor: Failed to load ANI cursor [%s]\n", file_location);
```
xezon: `\n` can be omitted in all logs, because captainslog always appends it already.

src/platform/input/sdl2mouse.cpp
xezon: Ok. So this is something that will be done in a follow up change?

feliwir: Yes. It's a bit hard to program things like these completly blind

src/platform/input/sdl2mouse.cpp
feliwir: Handled now

src/platform/input/sdl2mouse.cpp
feliwir: Copied this from different places. I'll remove it though

src/platform/input/sdl2mouse.cpp
xezon: Hmm. This should be testable on Windows with the game dat + thyme dll. Do you have a Windows install?

feliwir: No. I tried once but didn't really workout all that great

xezon: If you can install Windows on your machine, then we should be able to get Thyme running in DLL mode. This way you can test all changes.

src/CMakeLists.txt
```diff
@@ -468,6 +468,7 @@ if(USE_SDL2)
     list(APPEND GAME_COMPILE_OPTIONS -DBUILD_WITH_SDL2)
 
     list(APPEND GAMEENGINE_SRC
+        platform/sdl2/sdl2keybd.cpp
```
xezon, line 560: `sdl2keyboard` is ok name. There is no need for short acronyms.

src/platform/sdl2/sdl2mouse.cpp
```diff
@@ -162,34 +159,32 @@ void SDL2Mouse::Translate_Event(uint32_t message_num, MouseIO *io)
             break;
         case SDL_MOUSEBUTTONDOWN:
         case SDL_MOUSEBUTTONUP:
-            state = event.type == SDL_MOUSEBUTTONDOWN ? MouseIO::MouseState::MOUSE_STATE_DOWN :
-                                                        MouseIO::MouseState::MOUSE_STATE_UP;
+            doubleClick = event.button.button == m_lastClick.button.button
+                && (event.button.timestamp - m_lastClick.button.timestamp) < Get_Double_Click_Time();
```
xezon, line 169: What happens if you click a button 3 times in quick succession? Will it trigger 2x double click? And if yes, is that desirable?

src/CMakeLists.txt
feliwir: i used that acronym due to `dinputkeybd.cpp`

CMakeLists.txt
```diff
@@ -234,6 +237,25 @@ if(USE_GAMEMATH)
     endif()
 endif()
 
+if(USE_SDL2)
```
OmniBlade, line 265: I'd like to see an attempt to find SDL2 first and fall back to building SDL2 ourselves if we don't find it with a warning message informing that SDL2 wasn't found and that we are going to build it ourselves.

src/platform/sdl2/sdl2keybd.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+    // Translate_Event(m_nextGetIndex, io);
+
+    // Clear the current event - can be overwritten
+    m_eventBuffer[m_nextGetIndex].type = 0;
+    m_nextGetIndex++;
+    if (m_nextGetIndex >= MAX_EVENTS) {
+        m_nextGetIndex = 0;
+    }
+}
```
OmniBlade: Nitpick, files should end on a new line.

OmniBlade (changes requested):

src/CMakeLists.txt
xezon: ok

src/game/main.cpp
```diff
@@ -269,6 +283,49 @@ void Create_Window()
...
+
+    if (SDL_Init(SDL_INIT_VIDEO) < 0) {
+        captainslog_error("SDL could not initialize!");
+        return false;
+    }
+
+    g_creatingWindow = true;
+
+    bool is_windowed = g_gameIsWindowed;
+    g_applicationWindow = SDL_CreateWindow(WIN_TITLE,
```
xezon, line 511: According to documentation this call can return null. This seems to be not handled here.
https://wiki.libsdl.org/SDL_CreateWindow

src/platform/sdl2/sdl2keybd.cpp
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Gets key state.
+ */
+void SDL2InputKeyboard::Get_Key(KeyboardIO *io)
+{
+    if (m_eventBuffer[m_nextGetIndex].type == 0) {
+        return;
+    }
+
+    // Translate_Event(m_nextGetIndex, io);
```
xezon, line 72: What is this comment here for?

src/platform/sdl2/sdl2keybd.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+    // Clear the current event - can be overwritten
+    m_eventBuffer[m_nextGetIndex].type = 0;
```
xezon: `SDL_NOEVENT`

src/platform/sdl2/sdl2keybd.cpp
```diff
@@ -0,0 +1,78 @@
...
+        m_nextFreeIndex = 0;
+    }
+}
+
+/**
+ * Gets key state.
+ */
+void SDL2InputKeyboard::Get_Key(KeyboardIO *io)
+{
+    if (m_eventBuffer[m_nextGetIndex].type == 0) {
```
xezon, line 68: `SDL_NOEVENT`

src/platform/sdl2/sdl2keybd.cpp
```diff
@@ -0,0 +1,78 @@
...
+    SDL_Keymod mod = SDL_GetModState();
+    return mod & KMOD_CAPS;
+}
+
+void SDL2InputKeyboard::Add_SDL2_Event(SDL_Event *ev)
+{
+    auto &event = m_eventBuffer[m_nextFreeIndex];
+
+    // If no space (0 type == empty) for more events throw away event
+    if (event.type != 0) {
```
xezon: `SDL_NOEVENT`

src/platform/sdl2/sdl2mouse.cpp
```diff
@@ -0,0 +1,206 @@
...
+    m_visible = visibility;
+    SDL_ShowCursor(visibility);
+}
+
+void SDL2Mouse::Add_SDL2_Event(SDL_Event *ev)
+{
+    auto &event = m_eventBuffer[m_nextFreeIndex];
+
+    // If no space (0 type == empty) for more events throw away event
+    if (event.type != 0) {
```
xezon, line 104: `SDL_NOEVENT`

src/platform/sdl2/sdl2mouse.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+    m_nextFreeIndex++;
+    if (m_nextFreeIndex >= MAX_EVENTS) {
+        m_nextFreeIndex = 0;
+    }
+}
+
+uint8_t SDL2Mouse::Get_Mouse_Event(MouseIO *io, int8_t unk)
+{
+    if (m_eventBuffer[m_nextGetIndex].type == 0) {
```
xezon: `SDL_NOEVENT`

src/platform/sdl2/sdl2mouse.cpp
```diff
@@ -0,0 +1,206 @@
...
+        return false;
+    }
+
+    Translate_Event(m_nextGetIndex, io);
+
+    // Clear the current event - can be overwritten
+    m_eventBuffer[m_nextGetIndex].type = 0;
```
xezon: `SDL_NOEVENT`

src/platform/sdl2/sdl2mouse.cpp
xezon: On second look this double click does not look proper. Note that this code runs on both `SDL_MOUSEBUTTONDOWN` and `SDL_MOUSEBUTTONUP`, so either `event.button.button == m_lastClick.button.button` condition is never true, or will will trigger as once you press and release mouse button once.

xezon (changes requested):

src/platform/sdl2/sdl2keybd.cpp
feliwir: This still needs to be done. SDL keycodes need to be translated to internal/engine specific keycodes

src/platform/sdl2/sdl2mouse.cpp
feliwir: Thanks, good catch!
I changed this to only check for double click on mouse down. However this still replaces the 2nd click by a doubleclick event, which is probably wrong. I want to get more of the surrounding code implemented, so we can properly test this!

OmniBlade (dismissed):
Seems to be missing a find module for SDL2, https://github.com/TheAssemblyArmada/Vanilla-Conquer/blob/vanilla/cmake/FindSDL2.cmake is the one we use in vanilla-conquer.

src/game/main.cpp
```diff
@@ -479,6 +494,93 @@ void Create_Window()
...
+        SDL_DestroyRenderer(renderer);
+        captainslog_error("SDL could not load splashscreen texture!");
+        return false;
+    }
+
+    // Draw the splashscreen
+    SDL_SetRenderDrawColor(renderer, 0, 0, 0, 255);
+    SDL_RenderClear(renderer);
+    SDL_RenderCopy(renderer, texture, NULL, NULL);
+    SDL_RenderPresent(renderer);
```
feliwir: This gets rendered only once - so it stays until we render something else after loading

src/platform/sdl2/sdl2gameengine.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+SDL_Window *g_applicationWindow;
+
+namespace Thyme
+{
+SDL2GameEngine::SDL2GameEngine() {}
+
+SDL2GameEngine::~SDL2GameEngine() {}
+
+void SDL2GameEngine::Handle_Window_Event(SDL_Event *ev)
```
feliwir, line 38: The stuff here and in `Service_Windows_OS` was inspired by `Wnd_Proc` in main.cpp

feliwir:
What was missing here in the end?

xezon:
Was all the feedback addressed?

feliwir:
> Was all the feedback addressed?

i don't see any comments left unreplied

feliwir:
> Seems to be missing a find module for SDL2, https://github.com/TheAssemblyArmada/Vanilla-Conquer/blob/vanilla/cmake/FindSDL2.cmake is the one we use in vanilla-conquer.

SDL comes with a SDL2config.cmake, which takes care of the `find_package` procedure

src/game/common/gameengine.cpp
```diff
@@ -140,11 +140,13 @@ void GameEngine::Init(int argc, char *argv[])
         "Data/INI/Default/Roads.ini",
         "Data/INI/Roads.ini");
     Init_Subsystem(g_theGlobalLanguage, "TheGlobalLanguageData", new GlobalLanguage);
+#ifdef GAME_DLL
```
xezon: This means it is compiled out in Standalone. I assume this is not what is wanted?

src/game/common/globaldata.cpp
```diff
@@ -1039,7 +1039,8 @@ GlobalData::GlobalData()
 #ifdef PLATFORM_WINDOWS
     m_doubleClickTime = GetDoubleClickTime();
 #else
-// TODO, probably based on whatever crossplatform event framework we use.
+    // TODO, probably based on whatever crossplatform event framework we use. For now use a constant (1000ms)
+    m_doubleClickTime = 1000;
```
xezon, line 1044: Is this a reasonable duration? Seems quite long for a double click. Seems Grandpa friendly tho.

src/game/client/gametext.cpp
```diff
@@ -767,7 +771,11 @@ void GameTextManager::Init()
 
     ntitle.Translate(wtitle);
 
-#ifdef PLATFORM_WINDOWS
+#ifdef BUILD_WITH_SDL2
+    if (g_applicationWindow != NULL) {
```
xezon: nullptr

src/platform/sdl2/sdl2gameengine.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include "w3dfunctionlexicon.h"
+#include "w3dmodulefactory.h"
+#include "win32bigfilesystem.h"
+#include "win32localfilesystem.h"
+
+#ifdef BUILD_WITH_STDFS
+#include "stdlocalfilesystem.h"
+#endif
+
+SDL_Window *g_applicationWindow;
```
xezon: = nullptr

src/game/main.cpp
```diff
@@ -81,14 +85,17 @@ extern HINSTANCE &g_applicationHInstance;
 #else
 #ifdef PLATFORM_WINDOWS
 HWND g_applicationHWnd;
+HGDIOBJ g_splashImage;
+HINSTANCE g_applicationHInstance;
+#endif
+#ifdef BUILD_WITH_SDL2
+// SDL_Window *g_applicationWindow;
```
xezon: What is this for?

src/game/main.cpp
```diff
@@ -479,6 +494,93 @@ void Create_Window()
 #endif
 }
 
+#ifdef BUILD_WITH_SDL2
+namespace Thyme
+{
+bool Create_Window_SDL2()
+{
+    g_applicationWindow = NULL;
```
xezon: nullptr

src/game/main.cpp
```diff
@@ -479,6 +494,93 @@ void Create_Window()
...
+    SDL_GetWindowWMInfo(g_applicationWindow, &wmInfo);
+    g_applicationHWnd = wmInfo.info.win.window;
+#endif
+    return true;
+}
+
+void Destroy_Window_SDL2()
+{
+    if (g_applicationWindow)
+        SDL_DestroyWindow(g_applicationWindow);
```
xezon, line 568: `g_applicationWindow = nullptr`

src/platform/sdl2/sdl2gameengine.h
```diff
@@ -0,0 +1,57 @@
...
+    virtual Radar *Create_Radar() override;
+    virtual WebBrowser *Create_Web_Browser() override;
+    virtual ParticleSystemManager *Create_Particle_System_Manager() override;
+    virtual AudioManager *Create_Audio_Manager() override;
+    virtual Network *Create_Network() override;
+
+private:
+    void Handle_Window_Event(SDL_Event *ev);
+};
+} // namespace Thyme
```
xezon: Add eol

src/game/common/gameengine.cpp
feliwir: Well, until #618 is ready the standalone version will always crash at that line. I can remove it, but then the standalone won't get past here

src/game/main.cpp
feliwir: no idea, i removed it. Probably a leftover from earlier

src/game/common/gameengine.cpp
xezon: Ok put a comment then explaining situation.

src/game/main.cpp
```diff
@@ -576,6 +573,7 @@ void Destroy_Window_SDL2()
     if (g_applicationWindow)
         SDL_DestroyWindow(g_applicationWindow);
 
+    g_applicationWindow = nullptr;
```
xezon, line 570: This can go into scope above.

xezon (approved):
Looks ok to me. Check with OmniBlade for final call on this.

feliwir:
Updated to work with the recent changes

feliwir:
> Seems to be missing a find module for SDL2, https://github.com/TheAssemblyArmada/Vanilla-Conquer/blob/vanilla/cmake/FindSDL2.cmake is the one we use in vanilla-conquer.

Fixed

xezon:
Merge with Rebase

PR #575, opened 2022-02-16
Implements ModalWindow, more of GameWindow.

tomsons26:


src/game/client/gui/gamewindow.cpp
```diff
@@ -397,6 +562,108 @@ int GameWindow::Win_Set_Enabled_Color(int index, int color)
     return 0;
 }
 
+int GameWindow::Win_Set_Enabled_Border_Color(int index, int color)
+{
+    if (index < 0 || index >= WinInstanceData::MAX_DRAW_DATA) {
+        captainslog_debug("Set enabled border color, index out of range '%d'", index);
+        return -3;
```
xezon, line 569: This appears to be error code. 0 is NO_ERROR. -3 is INVALID_INDEX or similar. I suggest returning enum value instead of int.

src/game/client/gui/gamewindow.h
```diff
@@ -30,6 +31,75 @@ enum WindowMsgHandledType
...
+    GWS_USER = 1 << 9,
+
+    GWS_MOUSE_TRACK = 1 << 10,
+
+    GWS_ANIMATED = 1 << 11,
+
+    GWS_TAB_STOP = 1 << 12,
+    GWS_TAB_CONTROL = 1 << 13,
+    GWS_TAB_PLANE = 1 << 14,
+
```
xezon, line 99: Can remove the spacing.

src/game/client/gui/gamewindow.cpp
tomsons26: I need all the code done to make a decision on this, and thats far off

PR #577, opened 2022-02-17
W3DTankDraw & W3DOverlordTankDraw

jonwil:


src/game/client/draw/w3dtankdraw.cpp
```diff
@@ -0,0 +1,379 @@
...
+    for (int i = 0; i < m_treadCount; i++) {
+        float offset = 0;
+
+        if (m_treads[i].m_type == TREAD_LEFT) {
+            offset = uv_delta + m_treads[i].m_materialSettings.m_customUVOffset.X;
+        } else if (m_treads[i].m_type == TREAD_RIGHT) {
+            offset = m_treads[i].m_materialSettings.m_customUVOffset.X - uv_delta;
+        }
+
+        offset = GameMath::Floor(offset);
```
tomsons26: offset = offset - GameMath::Floor(offset);

src/game/client/draw/w3dtankdraw.cpp
```diff
@@ -0,0 +1,379 @@
...
+                    Start_Move_Debris();
+                } else {
+                    Stop_Move_Debris();
+                }
+
+                len = GameMath::Sqrt(len);
+                Coord3D mul;
+                mul.x = 0.5f * len + 0.1f;
+
+                if (mul.x > 1.0) {
```
tomsons26: f

src/game/client/draw/w3dtankdraw.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+NameKeyType W3DTankDraw::Get_Module_Name_Key() const
+{
+    static const NameKeyType nk = g_theNameKeyGenerator->Name_To_Key("W3DTankDraw");
+    return nk;
+}
+
+void W3DTankDraw::Do_Draw_Module(const Matrix3D *transform)
+{
+    bool frozen = g_theTacticalView->Is_Time_Frozen() && !g_theTacticalView->Is_Camera_Movement_Finished();
```
tomsons26, line 92: you sure on these frozen checks? original seems to have two bools

tomsons26:
rename pr and commit since its two classes

src/game/client/draw/w3dtankdraw.cpp
jonwil: fixsed.

src/game/client/draw/w3dtankdraw.cpp
jonwil: fixed.

src/game/client/draw/w3dtankdraw.cpp
jonwil: fixed.

src/game/client/draw/w3dmodeldraw.h
```diff
@@ -400,6 +400,7 @@ class W3DModelDraw : public DrawModule, public ObjectDrawInterface
     void Stop_Client_Particle_Systems();
 
     RenderObjClass *Get_Render_Object() { return m_renderObject; }
+    bool Get_Fully_Obscured_By_Shroud() { return m_fullyObscuredByShroud; }
```
xezon: `const`

src/game/client/draw/w3doverlordtankdraw.h
```diff
@@ -0,0 +1,53 @@
...
+    W3DOverlordTankDrawModuleData() {}
+    virtual ~W3DOverlordTankDrawModuleData() {}
+    static void Build_Field_Parse(MultiIniFieldParse &p);
+
+private:
+    Utf8String m_unk1;
+    Utf8String m_unk2;
+    int m_unk3;
+    int m_unk4;
+    int m_unk5;
```
xezon, line 32: Are these member correct here? I am asking because the same members are already in `W3DTankDrawModuleData`, which this class inherits from. So looks like it should be just that. Parse table for `W3DOverlordTankDrawModuleData` is also empty.

If they belong here and are useless members, I suggest adding a comment to clarify the situation that they are here for binary compatibility. Right now `W3DOverlordTankDrawModuleData` is 2x `W3DTankDrawModuleData`.

src/game/client/draw/w3dtankdraw.h
```diff
@@ -0,0 +1,96 @@
...
+    virtual ~W3DTankDrawModuleData() {}
+    static void Build_Field_Parse(MultiIniFieldParse &p);
+
+private:
+    Utf8String m_treadDebrisNameLeft;
+    Utf8String m_treadDebrisNameRight;
+    float m_treadAnimationRate;
+    float m_treadPivotSpeedFraction;
+    float m_treadDriveSpeedFraction;
+    friend class W3DTankDraw;
```
xezon, line 33: Should `W3DTankDraw` class not be forward declared before it is used here for friending? I am not sure how C++ treats it.

src/game/client/draw/w3dtankdraw.cpp
```diff
@@ -0,0 +1,379 @@
...
+                    float speed = object->Get_AI_Update_Interface()->Get_Cur_Locomotor_Speed();
+
+                    if (turning != TURN_NONE
+                        && physics->Get_Velocity_Magnitude() / speed
+                            < Get_W3D_Tank_Draw_Module_Data()->m_treadPivotSpeedFraction) {
+                        Coord3D dir;
+                        object->Get_Unit_Dir_Vector2D(dir);
+
+                        if (GameMath::Fabs(1.0f - (dir.x * m_treadDirection.x + dir.y * m_treadDirection.y))
+                            > 0.000009999999747378752f) {
```
xezon: I suspect this is equal to `0.00001f`, its just IEEE doing its thing here. Can you check?

src/game/client/draw/w3dtankdraw.cpp
```diff
@@ -0,0 +1,379 @@
...
+        for (int j = 0; j < robj->Get_Num_Sub_Objects() && m_treadCount < 4; j++) {
+            RenderObjClass *subobj = robj->Get_Sub_Object(j);
+
+            if (subobj != nullptr) {
+                if (subobj->Class_ID() == RenderObjClass::CLASSID_MESH) {
+                    if (subobj->Get_Name() != nullptr) {
+                        const char *str = strchr(subobj->Get_Name(), '.');
+                        if (str != nullptr) {
+                            if ((*str++) != 0) {
+                                if (!strncasecmp(str, "TREADS", 6)) {
```
xezon: Could you possibly reduce the indentation here by bundling some conditions?

src/game/client/draw/w3dtankdraw.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+    for (int i = 0; i < m_treadCount; i++) {
+        Ref_Ptr_Release(m_treads[i].m_robj);
+    }
+
+    m_treadCount = 0;
+
+    if (Get_W3D_Tank_Draw_Module_Data() != nullptr && Get_W3D_Tank_Draw_Module_Data()->m_treadAnimationRate != 0.0f
+        && robj != nullptr) {
+        for (int j = 0; j < robj->Get_Num_Sub_Objects() && m_treadCount < 4; j++) {
```
xezon: `m_treadCount < MAX_TREADS_PER_TANK` ?

src/game/client/draw/w3dtankdraw.h
```diff
@@ -0,0 +1,96 @@
...
+    virtual void Load_Post_Process() override;
+
+    virtual void Set_Hidden(bool hidden) override;
+
+    void Toss_Emitters();
+    void Create_Emitters();
+    void Start_Move_Debris();
+    void Stop_Move_Debris();
+    void Update_Tread_Positions(float uv_delta);
+    void Update_Tread_Objects();
```
xezon, line 79: Do any of these non virtual functions need hooking because they are called from class owners? If not, I suggest moving them private.

src/game/client/draw/w3dtankdraw.cpp
```diff
@@ -0,0 +1,379 @@
...
+            RenderObjClass *subobj = robj->Get_Sub_Object(j);
+
+            if (subobj != nullptr) {
+                if (subobj->Class_ID() == RenderObjClass::CLASSID_MESH) {
+                    if (subobj->Get_Name() != nullptr) {
+                        const char *str = strchr(subobj->Get_Name(), '.');
+                        if (str != nullptr) {
+                            if ((*str++) != 0) {
```
xezon: I am confused by this `(*str++) != 0`

Is this meant to be `*str++ != 0` (1) or `++*str != 0` (2). Right now it resembles the preincrement (2) if I am not mistaken. If it does, I suggest writing it as `++*str != 0` for clarity.

src/game/client/draw/w3dtankdraw.cpp
```diff
@@ -0,0 +1,379 @@
...
+    m_treadDirection.y = 0.0f;
+    m_treadDirection.z = 0.0f;
+    Create_Emitters();
+}
+
+W3DTankDraw::~W3DTankDraw()
+{
+    for (int i = 0; i < MAX_TREADS_PER_TANK; i++) {
+        Ref_Ptr_Release(m_treads[i].m_robj);
+    }
```
xezon, line 81: Looks like the emitters are leaked here. Should this call `Toss_Emitters` ?

src/game/client/draw/w3dmodeldraw.h
jonwil: fixed.

src/game/client/draw/w3dtankdraw.h
jonwil: No this works just fine.

src/game/client/draw/w3dtankdraw.cpp
jonwil: Looks that way, fixed.

src/game/client/draw/w3dtankdraw.cpp
jonwil: fixed.

src/game/client/draw/w3dtankdraw.cpp
jonwil: I checked and this is doing the != check before doing the ++. Fixed.

src/game/client/draw/w3doverlordtankdraw.h
jonwil: fixed.

src/game/client/draw/w3dtankdraw.cpp
jonwil: I can't see anything else that would free these emitters and W3DTruckDraw's destructor calls W3DTruckDraw::Toss_Emitters so it seems logical to add a call here. fixed.

src/game/client/draw/w3dtankdraw.cpp
jonwil: fixed.

src/game/client/draw/w3dtankdraw.cpp
```diff
@@ -0,0 +1,376 @@
...
+}
+
+void W3DTankDraw::Update_Tread_Positions(float uv_delta)
+{
+    for (int i = 0; i < m_treadCount; i++) {
+        float offset = 0;
+
+        if (m_treads[i].m_type == TREAD_LEFT) {
+            offset = uv_delta + m_treads[i].m_materialSettings.m_customUVOffset.X;
```
xezon: `m_treads[i].m_materialSettings.m_customUVOffset.X + uv_delta` will make it look better in function.

src/game/client/draw/w3dtankdraw.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+                m_treadDebrisLeft->Set_Velocity_Multiplier(&mul);
+                m_treadDebrisRight->Set_Velocity_Multiplier(&mul);
+                m_treadDebrisLeft->Set_Burst_Count_Multiplier(mul.z);
+                m_treadDebrisRight->Set_Burst_Count_Multiplier(mul.z);
+
+                if (m_treadCount != 0) {
+                    PhysicsTurningType turning = physics->Get_Turning();
+                    float rate = Get_W3D_Tank_Draw_Module_Data()->m_treadAnimationRate;
+                    float speed = object->Get_AI_Update_Interface()->Get_Cur_Locomotor_Speed();
```
xezon, line 138: Can speed be 0 ? If so we would divide by 0 further down.

xezon:
Tank treads appear normal looking in game.

src/game/client/draw/w3dtankdraw.cpp
jonwil: No it isn't a problem.

src/game/client/draw/w3dtankdraw.cpp
jonwil: Fixed.

xezon (approved):
Good from my side.

PR #580, opened 2022-02-18
Remove several C casts in renderer code

xezon:


src/w3d/renderer/dx8renderer.cpp
```diff
@@ -582,29 +583,27 @@ void DX8TextureCategoryClass::Render()
 #endif
             }
 
+            RenderObjClass::Material_Override *moverride =
```
duncanspumpkin: this name is a bit odd. maybe `mat_override` would be better. I first read this as `move rride`

src/w3d/renderer/mapper.cpp
```diff
@@ -882,7 +882,7 @@ BumpEnvTextureMapperClass::BumpEnvTextureMapperClass(const BumpEnvTextureMapperC
 
 inline unsigned long F2DW(float f)
```
duncanspumpkin, line 883: There are two of these in the codebase?? Can you create an issue to track merging them (and changing it to a dword instead of usigned long)

duncanspumpkin (approved):

src/w3d/renderer/mapper.cpp
xezon: There are actually 3 of these.

src/w3d/renderer/dx8renderer.cpp
xezon: Ok

PR #581, opened 2022-02-18
Replace C casts with union in Vector2i and Vector3i

xezon:


duncanspumpkin:
Is A used elsewhere in the code base? Perhaps Data would be a better name

xezon:
Ok changed `Data`.

duncanspumpkin (approved):

PR #584, opened 2022-02-18
Remove some C casts in w3d lib code

xezon:


src/w3d/lib/multilist.h
```diff
@@ -19,6 +19,8 @@
 #include "autopool.h"
 #include <captainslog.h>
 
+// #TODO investigate C casts in this file.
```
duncanspumpkin, line 22: is this note still valid?

src/w3d/lib/random.cpp
```diff
@@ -158,7 +158,7 @@ int Random4Class::operator()()
     y ^= TEMPERING_SHIFT_T(y) & TEMPERING_MASK_C;
     y ^= TEMPERING_SHIFT_L(y);
 
-    int *x = (int *)&y;
+    int *x = reinterpret_cast<int *>(&y);
```
duncanspumpkin: ummm what is this function doing?? It can just `return static_cast<int32_t>(y);` and skip the x completely.

src/w3d/lib/random.cpp
```diff
@@ -171,7 +171,7 @@ int Random4Class::operator()(int minval, int maxval)
 float Random4Class::Get_Float()
 {
     int x = (*this)();
-    unsigned int *y = (unsigned int *)&x;
+    unsigned int *y = reinterpret_cast<unsigned int *>(&x);
```
duncanspumpkin: again this is just plain weird. There is no need to use a pointer here `auto y = static_cast<uint32_t>(x);`

src/w3d/math/tri.h
```diff
@@ -37,8 +37,10 @@ class TriClass
         captainslog_assert(V[1] != nullptr);
         captainslog_assert(V[2] != nullptr);
 
-        Vector3::Cross_Product(*(V[1]) - *(V[0]), *(V[2]) - *(V[0]), (Vector3 *)N);
-        ((Vector3 *)N)->Normalize();
+        // #TODO This is a bad const_cast and should be fixed.
```
duncanspumpkin, line 40: It's a very odd class design. Almost never is a const member what you want.

src/w3d/lib/multilist.h
xezon: Yes. There is a lot going on in this file and it's too much to unpack easily.

src/w3d/lib/random.cpp
xezon: Actually that would not be the same. `static_cast` would perform conversion. Currently it does not and simply takes unsigned value and treats it as is for int. Changing this would change logic, which I would like to not do here without understanding the consequences.

But yes, it's certainly odd on first sight.

src/w3d/lib/random.cpp
xezon: Same as above.

src/w3d/lib/random.cpp
duncanspumpkin: That still doesn't make sense if that was the case you would reinterpret cast straight to the unsigned value, no need for the pointers.

https://stackoverflow.com/questions/1751346/interpret-signed-as-unsigned

It looks like this is changing defined behaviour for implementation defined behaviour

xezon: I can tell you right now that
`int *x = reinterpret_cast<int *>(&y)`
does not yield same result as
`int x = static_cast(y)`

The first will create type without conversion and the second with conversion. I will not change this for the sake of not altering this algorithm.

`int *x = reinterpret_cast<int *>(&y)` is perfectly legal c++

duncanspumpkin: https://godbolt.org/z/cjEzMoGxa its legal but its pointless. `static_cast` will always produce the same result. You can see it in the attached all the mov's are exactly the same except the reinterpret that has to add an extra load.

xezon: Alright then. I changed to static casts. I am still not convinced that this is 100% identical behavior, but if you vouch for that, then you can Approve it like so.

duncanspumpkin (approved):

PR #585, opened 2022-02-18
Remove C casts in RankInfo class

xezon:


src/game/common/ini/ini.cpp
```diff
@@ -112,7 +112,7 @@ BlockParse TheTypeTable[] = {
     {"Road", &TerrainRoadCollection::Parse_Terrain_Road_Definition},
     {"Science", (iniblockparse_t)(0x00488EA0) /*&INI::parseScienceDefinition*/},
     //{ "Rank", (iniblockparse_t)(0x00489800)/*&INI::parseRankDefinition*/ },
-    {"Rank", &RankInfoStore::Parse_Rank_Definition},
+    {"Rank", &RankInfoStore::Friend_Parse_Rank_Definition},
```
duncanspumpkin: Why the name change? The old name was consistent with the other functions

xezon: This is the name I saw in original. Can you double check?

duncanspumpkin: I don't really care about original i care about consistency. The original had this function as part of ini but we just make it public so it's not a friend anymore.

xezon: There is also this:
`friend void RankInfoStore::Friend_Parse_Rank_Definition(INI *)`

I suspect that's the reason for the Friend name?

duncanspumpkin: Yes but it's unused

xezon: Removing that `friend` will break compilation. So it appears to be used.

xezon: Restored new `Parse_Rank_Definition` name.

duncanspumpkin (approved):

PR #586, opened 2022-02-18
Add static asserts to Xfer to invoke extra caution if certain type si…

xezon:
…zes suddenly deviate from the norm


src/game/client/draw/w3dmodeldraw.cpp
```diff
@@ -2970,7 +2970,6 @@ void W3DModelDraw::Xfer_Snapshot(Xfer *xfer)
                         if (anim) {
                             float frame;
                             int frames;
-                            int mode;
```
xezon, line 2975: This looks like an error. `int mode` was shadowing the serialized mode from above.

duncanspumpkin:
Instead of having only some using the correct size types I propose we just switch the codebase in full one type at a time it will lead to more consistent code #606

PR #588, opened 2022-02-18
Match reference counter type in String classes with Windows interlock…

xezon:
…ed method

### DIFF
--- a/src/game/common/system/asciistring.h
+++ b/src/game/common/system/asciistring.h
@@ -55,9 +55,9 @@ public:
 #endif // GAME_DEBUG_STRUCTS
 
 #ifndef GAME_DLL
-        std::atomic<uint16_t> ref_count;
+        std::atomic<int16_t> ref_count;
 #else
-        uint16_t ref_count;
+        volatile int16_t ref_count;
 #endif
         uint16_t num_chars_allocated;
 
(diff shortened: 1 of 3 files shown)


src/game/common/system/asciistring.h
```diff
@@ -55,9 +55,9 @@ class Utf8String
 #endif // GAME_DEBUG_STRUCTS
 
 #ifndef GAME_DLL
-        std::atomic<uint16_t> ref_count;
+        std::atomic<int16_t> ref_count;
 #else
-        uint16_t ref_count;
+        volatile int16_t ref_count;
```
xezon, line 60: I really think this should be a volatile signed short. unsigned makes no sense. And volatile because that's how interlocked variables are used (avoids certain thread optimization).

duncanspumpkin: `volatile` is not correct for this its sole purpose is for memory mapped I/O which this is not. By adding it here it is not improving anything or implying anything.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconc-volatile

xezon: Windows API disagrees. Interlocked function expect volatile variable.

Here is a better explanation on this matter.
https://stackoverflow.com/questions/4437527/why-do-we-use-volatile-keyword

duncanspumpkin: I did not notice it was actually doing the addition through the win api intrinsic. Okay carry on.

OmniBlade (approved):

PR #590, opened 2022-02-19
Remove C casts in client draw classes

xezon:


tomsons26 (approved):

src/game/client/draw/drawable.cpp
```diff
@@ -142,7 +142,7 @@ const Vector3 *Drawable::Get_Selection_Color() const
 
 DrawModule **Drawable::Get_Draw_Modules()
 {
-    DrawModule **modules = (DrawModule **)Get_Modules(MODULE_DRAW);
+    DrawModule **modules = reinterpret_cast<DrawModule **>(Get_Modules(MODULE_DRAW));
```
duncanspumpkin, line 145: Couldn't this have used `Get_Draw_Modules_Non_Dirty`. That would limit the number of reinterpret_casts to just the header. same goes for L169

xezon: I do not know that.

PR #591, opened 2022-02-19
Fix malicious invocations of SetPixelShaderConstant

xezon:
I could be wrong here, but to me it looks like to invocations of `SetPixelShaderConstant` with `D3DXVECTOR4` are wrong everywhere. It expects a pointer to one or multiple `D3DXVECTOR4`, but in Thyme we cast a stack object to `float*`. That does not look right.

### DIFF
--- a/src/game/client/shader/bwfilter.cpp
+++ b/src/game/client/shader/bwfilter.cpp
@@ -177,7 +177,7 @@ int ScreenBWFilter::Set(FilterModes mode)
     DX8Wrapper::Apply_Render_State_Changes();
     DX8Wrapper::Get_D3D_Device8()->SetPixelShader(m_dwBWPixelShader);
     D3DXVECTOR4 v(0.30000001f, 0.58999997f, 0.11f, 1.0f);
-    DX8Wrapper::Get_D3D_Device8()->SetPixelShaderConstant(0, (float *)v, 1);
+    DX8Wrapper::Get_D3D_Device8()->SetPixelShaderConstant(0, static_cast<CONST FLOAT *>(v), 1);
     D3DXVECTOR4 v2(1.0f, 1.0f, 1.0f, 1.0f);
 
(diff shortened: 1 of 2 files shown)


jonwil:
The casts here are incorrect but they should actually be changed to FLOAT casts so they will invoke D3DXVECTOR4::operator FLOAT* like the original binary does.

Actually technically they don't need to be (and are working fine) since FLOAT is a typedef for float.

Checked the compiled obj files and the calls here are definitely invoking the correct D3DXVECTOR4 operator.

xezon:
Ah ok it has cast operator overload. Yes in that case `static_cast<const FLOAT*>(vector)` will work. `&vector` is equally valid however. What do you like better?

tomsons26:
SetPixelShaderConstant basically takes 4 floats as a constant, nothing really has to be D3DXVECTOR4, just what they chose to use in this case
DX8 sdk example
```
        // Set the pixel shader constants
        FLOAT fPixelShaderConstants[8][4] =
        { //  Red  Green  Blue  Alpha
            { 1.0f, 0.0f, 0.0f, 1.0f, },  // red
            { 0.0f, 1.0f, 0.0f, 1.0f, },  // green
            { 0.0f, 0.0f, 1.0f, 1.0f, },  // blue
            { 1.0f, 1.0f, 0.0f, 1.0f, },  // yellow
            { 0.0f, 1.0f, 1.0f, 1.0f, },  // cyan
            { 1.0f, 0.0f, 1.0f, 1.0f, },  // purple
            { 1.0f, 1.0f, 1.0f, 1.0f, },  // white
            { 0.0f, 0.0f, 0.0f, 1.0f, },  // black
        };
        m_pd3dDevice->SetPixelShaderConstant( 0, (VOID*)fPixelShaderConstants, 8 );
```

xezon:
Personally i like simply `&vector` best. Is least text and easily understood.

OmniBlade (approved):

PR #592, opened 2022-02-19
Remove C casts in logic object code

xezon:


src/game/logic/object/update/aiupdate.h
```diff
@@ -50,7 +50,7 @@ class AICommandInterface
     virtual void AI_Do_Command(AICommandParms const *params) = 0;
 };
 
-class AIUpdateModuleData
+class AIUpdateModuleData : public ModuleData // #TODO or UpdateModuleData ??
```
tomsons26: mmmmmm @jonwil

xezon: UpdateModuleData it is.

tomsons26 (approved):

PR #594, opened 2022-02-19
Remove C casts in w3dengine code

xezon:


duncanspumpkin:
Can you make it so its purely removing c style casts and not modifying other items then there is no need to go work out if there are issues introduced.

xezon:
I don't understand. What is the problem you are having with this change?

duncanspumpkin (approved):

PR #596, opened 2022-02-19
Remove C casts in w3dwater code

xezon:


src/game/client/water/w3dwater.cpp
```diff
@@ -639,11 +641,11 @@ int WaterRenderObjClass::Init(float water_level, float dx, float dy, SceneClass
     m_shaderClass = g_zFillAlphaShader;
     m_shaderClass.Set_Cull_Mode(ShaderClass::CULL_MODE_DISABLE);
     m_alphaClippingTexture = W3DAssetManager::Get_Instance()->Get_Texture("TSMoonLarg.tga");
-    m_skyBox = ((GameAssetManager *)W3DAssetManager::Get_Instance())
+    m_skyBox = static_cast<GameAssetManager *>(W3DAssetManager::Get_Instance())
```
duncanspumpkin, line 644: Is there not a method that gets the specific type?

src/game/client/water/w3dwater.cpp
```diff
@@ -1069,7 +1072,7 @@ bool WaterRenderObjClass::Get_Clipped_Water_Plane(CameraClass *cam, AABoxClass *
 
 inline unsigned long F2DW(float f)
 {
-    return *((unsigned *)&f);
+    return *(reinterpret_cast<unsigned long *>(&f));
```
duncanspumpkin, line 1075: uint32_t suggest also changing the function signature to uint32_t

duncanspumpkin (approved):
Looks okay. Is there not a method that gets a GameAssetManager directly from the global. Perhaps I'm misremembering

src/game/client/water/w3dwater.cpp
xezon: I don't understand.

src/game/client/water/w3dwater.cpp
xezon: Yeah good point. There is another review with a 2 more of these functions. I think that should be done in a new review entirely.

src/game/client/water/w3dwater.cpp
xezon: Hmm no. `static W3DAssetManager *&s_theInstance;` is protected class member.

PR #597, opened 2022-02-19
Simplify code in Create_Bmp_File function

xezon:


src/platform/w3dengine/client/w3ddisplay.cpp
```diff
@@ -1397,42 +1397,47 @@ void W3DDisplay::Preload_Texture_Assets(Utf8String texture)
...
-    header->biWidth = width;
-    header->biHeight = height;
-    header->biPlanes = 1;
-    header->biBitCount = 24;
-    header->biCompression = 0;
-    header->biSizeImage = 24 * header->biHeight * ((header->biWidth + 7) / 8);
-    header->biClrImportant = 0;
-    HANDLE h = CreateFile(filename, GENERIC_READ | GENERIC_WRITE, 0, nullptr, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr);
+    // #BUGFIX Simplify code and avoid LocalAlloc. Use the stack.
+    BITMAPINFOHEADER iheader = { 0 };
```
xezon, line 1407: I removed `LocalAlloc` alltogether because it is useless to use. Just allocate on stack and it is fine.

src/platform/w3dengine/client/w3ddisplay.cpp
```diff
@@ -1397,42 +1397,47 @@ void W3DDisplay::Preload_Texture_Assets(Utf8String texture)
...
+    iheader.biSize = sizeof(BITMAPINFOHEADER);
+    iheader.biWidth = width;
+    iheader.biHeight = height;
+    iheader.biPlanes = 1;
+    iheader.biBitCount = 24;
+    iheader.biCompression = 0;
+    iheader.biSizeImage = 24 * iheader.biHeight * ((iheader.biWidth + 7) / 8);
+    iheader.biClrImportant = 0;
+    HANDLE h =
```
duncanspumpkin, line 1416: If we are refactoring you might as well use the normal c++ operations for this and avoid the win32 api functions.

duncanspumpkin: And then you could take this out of the `BUILD_WITH_D3D8` ifdef.

src/platform/w3dengine/client/w3ddisplay.cpp
```diff
@@ -1397,42 +1397,47 @@ void W3DDisplay::Preload_Texture_Assets(Utf8String texture)
...
-
-    // BUGFIX prevent memory leak;
-    LocalFree(header);
 #endif
 }
 
+#undef BFT_BITMAP
+#undef DibNumColors
+#undef DibSize
+#undef DibPaletteSize
```
duncanspumpkin, line 1439: why bother undefining. This is just the cpp file so they aren't going to leak to any other file.

src/platform/w3dengine/client/w3ddisplay.cpp
xezon: This is not possible with STL Port in place. I am not going to change logic here.

src/platform/w3dengine/client/w3ddisplay.cpp
xezon: The macros are undefined because they are not used beyond this place. I can move them into the D3D8 macro to further confine them.

src/platform/w3dengine/client/w3ddisplay.cpp
duncanspumpkin: Hmm you're write we don't have iostreams...
But we do have c api file i/o which would avoid this as well and is used throughout the codebase.

xezon: Yes it could be rewritten maybe, but that is beyond the scope of this change.

src/platform/w3dengine/client/w3ddisplay.cpp
```diff
@@ -1389,47 +1389,49 @@ void W3DDisplay::Preload_Texture_Assets(Utf8String texture)
...
+
+    // #BUGFIX Simplify code and avoid LocalAlloc. Use the stack.
+    BITMAPINFOHEADER iheader = { 0 };
+    iheader.biSize = sizeof(BITMAPINFOHEADER);
+    iheader.biWidth = width;
+    iheader.biHeight = height;
+    iheader.biPlanes = 1;
+    iheader.biBitCount = 24;
+    iheader.biCompression = 0;
+    iheader.biSizeImage = 24 * iheader.biHeight * ((iheader.biWidth + 7) / 8);
```
feliwir: why can this not just be 3 * height * width or for uncompressed RGB this also can just be set to 0 (according to docs: `Specifies the size, in bytes, of the image. This can be set to 0 for uncompressed RGB bitmaps.`)

src/platform/w3dengine/client/w3ddisplay.cpp
```diff
@@ -1389,47 +1389,49 @@ void W3DDisplay::Preload_Texture_Assets(Utf8String texture)
...
-    HANDLE h = CreateFile(filename, GENERIC_READ | GENERIC_WRITE, 0, nullptr, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr);
+
+    // #BUGFIX Simplify code and avoid LocalAlloc. Use the stack.
+    BITMAPINFOHEADER iheader = { 0 };
+    iheader.biSize = sizeof(BITMAPINFOHEADER);
+    iheader.biWidth = width;
+    iheader.biHeight = height;
+    iheader.biPlanes = 1;
+    iheader.biBitCount = 24;
+    iheader.biCompression = 0;
```
feliwir: I think this should be `BI_RGB` - see https://docs.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-bitmapinfoheader

feliwir: Or to be more precise: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-wmf/4e588f70-bd92-4a6f-b77f-35d0feaf7a57

src/platform/w3dengine/client/w3ddisplay.cpp
xezon: Excellent point. I didn't think twice about the weird math. `3 * height * width` yields same result. Tested and worked.

src/platform/w3dengine/client/w3ddisplay.cpp
xezon: Fixed.

src/platform/w3dengine/client/w3ddisplay.cpp
tomsons26: Could just use RawFileClass

jonwil (approved):
Approved.

PR #599, opened 2022-02-19
Fix unsafe string operation in DDSFileClass and ThumbnailClass

xezon:
As I was looking through code I saw this unsafe memory operation. The way it wants to write file extension is unsafe and will corrupt memory if file name is shorter than 3 characters. Or destroy file name if file extension is not set.

I placed proper function that can react to file names and then perform appropriate action. I put breakpoint at `// No file extension found.` and checked if hit, but it was not in my tests, which is good.

### DIFF
--- a/src/w3d/renderer/ddsfile.cpp
+++ b/src/w3d/renderer/ddsfile.cpp
@@ -44,11 +44,11 @@ DDSFileClass::DDSFileClass(const char *filename, unsigned reduction_factor) :
     m_totalSizeMaybe(0)
 {
     // Copy the name across and replace the extension with dds.
-    strlcpy_tpl(m_name, filename);
+    const size_t name_len = strlcpy_tpl(m_name, filename);
 
     // #BUGFIX Change texture extension safely.
     // Originally code would just write over name length -3 bytes without any checking.
-    Thyme::Change_Texture_File_Extension(m_name, "dds");
+    Thyme::Change_Texture_File_Extension(m_name, name_len, "dds");
 
     auto_file_ptr fp(g_theFileFactory, m_name);
 
@@ -639,12 +639,13 @@ unsigned DDSFileClass::Calculate_S3TC_Surface_Size(unsigned width, unsigned heig
 
 namespace Thyme
 {
-void Change_Texture_File_Extension(char *str, size_t str_capacity, const char *new_ext)
+void Change_Texture_File_Extension(char *str, size_t str_len, size_t str_capacity, const char *new_ext)
 {
-    const size_t len = strlen(str);
+    captainslog_dbgassert(str_len < str_capacity, "String length is expected to be smaller than string capacity");
+
     char *begin = str;
     char *end = str + str_capacity - 1u;
-    char *ext = str + len;
+    char *ext = str + str_len;
 
     while (--ext != begin) {
         char ch = *ext;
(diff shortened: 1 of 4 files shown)


src/w3d/renderer/ddsfile.cpp
```diff
@@ -45,10 +45,16 @@ DDSFileClass::DDSFileClass(const char *filename, unsigned reduction_factor) :
...
     strlcpy_tpl(m_name, filename);
+
+#if 0
     size_t name_len = strlen(m_name);
     m_name[name_len - 3] = 'd';
     m_name[name_len - 2] = 'd';
     m_name[name_len - 1] = 's';
+#else
+    // #BUGFIX Change texture extension safely.
+    Thyme::Change_Texture_File_Extension(m_name, "dds");
```
duncanspumpkin: This is pointless. There is absolutely no point for keeping the old code we have git for source control

src/w3d/renderer/ddsfile.cpp
```diff
@@ -636,3 +642,47 @@ unsigned DDSFileClass::Calculate_S3TC_Surface_Size(unsigned width, unsigned heig
 
     return ret;
 }
+
+namespace Thyme
+{
+void Change_Texture_File_Extension(char *str, size_t str_capacity, const char *new_ext)
+{
+    const size_t len = strlen(str);
```
duncanspumpkin: This is just as unsafe as the previous code. If str is missing its null terminator then function writes to invalid memory.

duncanspumpkin:
What was wrong with the built-in functions for handling this strchr, strcat? If you are replacing code to make a safe version then it should be bounds checking all of this not just one bit.

src/w3d/renderer/ddsfile.cpp
xezon: Do we have a strlen function with max len somewhere? I think we don't. Sounds like a separate fixing pass, as everything else uses strlen as well right now. We expect zero terminated string here. Your "unsafe" remark for the scope of this review is void.

xezon:
Feel free to reimplement function if you are not happy with it. I tested functions and it works properly.

src/w3d/renderer/ddsfile.cpp
xezon: Ok. I removed the ifdef 0.

xezon:
I made an improvement to get rid of the `strlen` call within that function.

src/w3d/renderer/ddsfile.cpp
```diff
@@ -636,3 +636,48 @@ unsigned DDSFileClass::Calculate_S3TC_Surface_Size(unsigned width, unsigned heig
...
+
+    if (ext == begin) {
+        // No file extension found.
+        ext = begin + str_len;
+        if (ext != end) {
+            *ext++ = '.';
+        }
+    }
+
+    while (ext != end && *new_ext != '\0') {
```
duncanspumpkin, line 675: This can be Replaced with strcat if you null terminate after the '.'

xezon: That would be no improvement, because then string needs to be walked from the beginning again, which is unnecessary. Algorithm already knows where to start here.

duncanspumpkin: You don't need to give it the pointer to the start just give it the one you already have. Also as this is just a generic file extension switcher you can drop the Texture from the name. Then it will be easier to find and replace with `fs::path::replace_extension`.

xezon: `strlcpy` could be used here, by calculating and passing it the remaining storage for the given buffer. I would refrain from doing that however. I already tested this code here and it works. Rewriting it again would require me testing it again for no real benefit in functionality. It would have to do the same thing that it already does here correctly.

PR #600, opened 2022-02-19
Remove Vector3 C casts

xezon:
I put static asserts and new methods to provide casting functionality for Vector4 to Vector3. This way there is visibility and confidence that this is ok to do.

### DIFF
--- a/src/w3d/math/vector4.cpp
+++ b/src/w3d/math/vector4.cpp
@@ -1,0 +1,23 @@
+/**
+ * @file
+ *
+ * @author Tiberian Technologies
+ * @author OmniBlade
+ *
+ * @brief 4D Vector class.
+ *
+ * @copyright Thyme is free software: you can redistribute it and/or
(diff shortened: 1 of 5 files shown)


src/w3d/math/vector4.cpp
```diff
@@ -0,0 +1,23 @@
...
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#include "vector4.h"
+
+#include "vector3.h"
+
```
duncanspumpkin, line 19: You could add a note that these must hold true for `To_Vector3_Ptr` to work

duncanspumpkin (approved):

src/w3d/math/vector4.cpp
xezon: Good point. I added hint.

PR #601, opened 2022-02-19
Remove Matrix4 C casts

xezon:
The goal here is to not have `Matrix4` mistaken for a `D3DMATRIX` and vice versa. **They are not the same thing**. And therefore should not be used interchangeably.

If a matrix is in `D3DMATRIX` form, then it will be stored that way. Otherwise it is a `Matrix4`. New conversion functions make the conversion explicit and readable. All C casts have been removed. ~~Only one `reinterpret_cast` was left.~~

### DIFF
--- a/src/game/client/water/w3dwater.cpp
+++ b/src/game/client/water/w3dwater.cpp
@@ -1090,17 +1090,16 @@ void WaterRenderObjClass::Draw_Sea(RenderInfoClass &rinfo)
         ww3d.m[2][1] = 1.0f;
         ww3d.m[1][2] = 1.0f;
         ww3d.m[3][3] = 1.0f;
-        Matrix3D m4(m_transform);
 
-        DX8Wrapper::Set_Transform(D3DTS_WORLD, m4);
+        DX8Wrapper::Set_Transform(D3DTS_WORLD, m_transform);
         DX8Wrapper::Set_Texture(0, nullptr);
         DX8Wrapper::Set_Texture(1, nullptr);
         DX8Wrapper::Apply_Render_State_Changes();
 
         Vector3 v;
         rinfo.m_camera.Get_Transform().Get_Translation(&v);
-        DX8Wrapper::Get_DX8_Transform(D3DTS_VIEW, (Matrix4 &)view);
-        DX8Wrapper::Get_DX8_Transform(D3DTS_PROJECTION, (Matrix4 &)proj);
+        DX8Wrapper::Get_DX8_Transform(D3DTS_VIEW, view);
+        DX8Wrapper::Get_DX8_Transform(D3DTS_PROJECTION, proj);
 
         m_pDev->SetTextureStageState(0, D3DTSS_COLORARG1, D3DTA_TEXTURE);
         m_pDev->SetTextureStageState(0, D3DTSS_COLORARG2, D3DTA_DIFFUSE);
@@ -1217,8 +1216,8 @@ void WaterRenderObjClass::Draw_Sea(RenderInfoClass &rinfo)
         m_pDev->SetTextureStageState(2, D3DTSS_COLOROP, D3DTOP_DISABLE);
         m_pDev->SetTextureStageState(2, D3DTSS_ALPHAOP, D3DTOP_DISABLE);
 
-        DX8Wrapper::Set_DX8_Transform(D3DTS_VIEW, (Matrix4 &)view);
-        DX8Wrapper::Set_DX8_Transform(D3DTS_PROJECTION, (Matrix4 &)proj);
+        DX8Wrapper::Set_DX8_Transform(D3DTS_VIEW, view);
+        DX8Wrapper::Set_DX8_Transform(D3DTS_PROJECTION, proj);
         m_pDev->SetPixelShader(0);
         m_pDev->SetVertexShader(DX8_FVF_XYZDUV1);
         DX8Wrapper::Invalidate_Cached_Render_States();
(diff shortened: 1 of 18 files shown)


src/w3d/renderer/sortingrenderer.cpp
```diff
@@ -271,14 +271,17 @@ void SortingRendererClass::Insert_Triangles(const SphereClass &bounding_sphere,
         captainslog_assert(state->vertex_count <= vertex_buffer->Get_Vertex_Count());
 
         D3DXMATRIX mtx;
-        D3DXMatrixMultiply(
-            &mtx, (const D3DXMATRIX *)&state->sorting_state.world, (const D3DXMATRIX *)&state->sorting_state.view);
-        D3DXVECTOR3 vec;
```
xezon, line 276: Note that `vec`, now `bounding_sphere_center`, did not have values assigned here. Thus math was wrong. Fixed.

src/game/client/water/w3dwater.cpp
```diff
@@ -1087,17 +1087,16 @@ void WaterRenderObjClass::Draw_Sea(RenderInfoClass &rinfo)
         ww3d.m[2][1] = 1.0f;
         ww3d.m[1][2] = 1.0f;
         ww3d.m[3][3] = 1.0f;
-        Matrix3D m4(m_transform);
```
xezon, line 1093: I think this needs checking.

src/game/client/water/w3dwater.cpp
```diff
@@ -1483,8 +1482,7 @@ void WaterRenderObjClass::Render_Water_Mesh()
         }
 
         m_vertexBufferD3D->Unlock();
-        Matrix3D m(m_transform);
```
xezon, line 1490: I think this needs checking.

src/game/client/water/w3dwater.cpp
xezon: Looks safe to do. `m_transform` already is `Matrix3D`, so this just creates a useless copy here and `DX8Wrapper::Set_Transform` just reads it.

src/game/client/water/w3dwater.cpp
xezon: Looks safe to do. `m_transform` already is `Matrix3D`, so this just creates a useless copy here and `DX8Wrapper::Set_Transform` just reads it.

src/w3d/math/matrix4.h
```diff
@@ -354,7 +354,7 @@ class Matrix4
         out->X = (A[0][0] * in.X + A[0][1] * in.Y + A[0][2] * in.Z + A[0][3]);
         out->Y = (A[1][0] * in.X + A[1][1] * in.Y + A[1][2] * in.Z + A[1][3]);
         out->Z = (A[2][0] * in.X + A[2][1] * in.Y + A[2][2] * in.Z + A[2][3]);
-        out->W = 1.0f;
+        out->W = (A[3][0] * in.X + A[3][1] * in.Y + A[3][2] * in.Z + A[3][3]);
```
xezon: I cannot recall why this was changed or why this would be correct, so I am going to revert this in this change.

src/w3d/renderer/dx8wrapper.h
```diff
@@ -197,15 +206,15 @@ class DX8Wrapper
     static void Clamp_Color(Vector4 &color);
     static unsigned int Convert_Color_Clamp(const Vector4 &color);
     static void Set_DX8_Light(int index, D3DLIGHT8 *light);
-    static void Get_DX8_Transform(D3DTRANSFORMSTATETYPE transform, Matrix4 &m);
-    static void Set_DX8_Transform(D3DTRANSFORMSTATETYPE transform, const Matrix4 &m);
+    static void Get_DX8_Transform(D3DTRANSFORMSTATETYPE transform, D3DMATRIX &m);
+    static void Set_DX8_Transform(D3DTRANSFORMSTATETYPE transform, const D3DMATRIX &m);
```
xezon, line 210: Matrix4 has been changed to D3DMATRIX for these 2 function because they clearly expect matrix in Direct X coordinate space, not a transposed Matrix4.

xezon:
I tested test level with many units and buildings. Killed bunch of units with Superweapons. Produced units. Played match vs Hard AI. Could not spot any new issues. From my side this change looks proper.

src/game/client/water/w3dwater.cpp
```diff
@@ -2388,15 +2386,15 @@ void WaterRenderObjClass::Setup_Jba_Water_Shader()
...
+        DX8Wrapper::Get_DX8_Transform(D3DTS_VIEW, m2);
         float f;
         D3DXMatrixInverse(&m, &f, &m2);
         D3DXMATRIX m3;
         D3DXMatrixScaling(&m3, 0.0625f, 0.0625f, 1.0f);
         D3DXMATRIX m4 = m * m3;
         D3DXMatrixTranslation(&m3, m_riverVOrigin, m_riverVOrigin, 0.0f);
         m4 = m4 * m3;
-        DX8Wrapper::Set_Transform(D3DTS_TEXTURE2, (Matrix4 &)m4);
+        DX8Wrapper::Set_Transform(D3DTS_TEXTURE2, reinterpret_cast<Matrix4 &>(m4));
```
xezon: This one looks suspicious to me. Should this not call `Set_DX8_Transform` ?

jonwil (approved):
Looks OK to me.

src/game/client/water/w3dwater.cpp
xezon: According to WB it calls `DX8Wrapper::_Set_DX8_Transform`. So this seems wrong.

xezon: DX8Wrapper::Set_Transform removed:

shot_20221002_142609_1

DX8Wrapper::Set_Transform(D3DTS_TEXTURE2, reinterpret_cast<Matrix4 &>(m4)):

shot_20221002_142924_1

DX8Wrapper::Set_DX8_Transform(D3DTS_TEXTURE2, m4);

shot_20221002_142734_1

It looks like this code is responsible for the white sparkle on the river water. Both calls Set_Transform and Set_DX8_Transform show the effect. It is difficult to tell the difference. Maybe the effect is flipped horizontally. In any case, looks like `Set_DX8_Transform` is working.

xezon: Ok. Fixed and pushed.

xezon:
Ok with this one fixed, this also means all matrix C / reinterpret_casts are gone now and thus type safe. Unless I missed something that is.

PR #602, opened 2022-02-19
Remove misc C casts

xezon:


src/game/client/shadow/w3dprojectedshadow.cpp
```diff
@@ -994,8 +1002,14 @@ int W3DProjectedShadowManager::Render_Shadows(RenderInfoClass &rinfo)
                         }
 
                         projector->Peek_Material_Pass()->UnInstall_Materials();
+
+                        Coord3D sphere_pos;
+                        sphere_pos.x = sphere.Center.X;
+                        sphere_pos.y = sphere.Center.Y;
+                        sphere_pos.z = sphere.Center.Z;
+
                         SimpleObjectIterator *iter = g_thePartitionManager->Iterate_Objects_In_Range(
```
duncanspumpkin, line 1011: Have you checked to confirm it doesn't modify sphere? I'm sure it doesn't but best to check

duncanspumpkin (approved):
All looks good.

src/game/client/shadow/w3dprojectedshadow.cpp
xezon: `PartitionManager::Iterate_Objects_In_Range` is not yet implemented. But the `Coord3D` is given as const, so it should not modify it, unless someone uses `const_cast` in there.

It seems unlikely to change coord, because it is meant to return an iterator for nearby objects.

duncanspumpkin: Good enough for me

PR #603, opened 2022-02-19
Fix issues in FontCharsClass::Create_GDI_Font

xezon:


src/w3d/renderer/render2dsentence.cpp
```diff
@@ -246,20 +246,22 @@ void FontCharsClass::Create_GDI_Font(const char *font_name)
...
-    bitmap_info.biClrUsed = 0;
-    bitmap_info.biClrImportant = 0;
-    m_gdiBitmap = CreateDIBSection(
-        screen_dc, (const BITMAPINFO *)&bitmap_info, DIB_RGB_COLORS, (void **)&m_gdiBitmapBits, nullptr, 0L);
+    // #BUGFIX Do not pass a BITMAPINFOHEADER* for BITMAPINFO* into CreateDIBSection function.
+
+    BITMAPINFO bitmap_info = { 0 };
+    bitmap_info.bmiHeader.biSize = sizeof(BITMAPINFOHEADER);
+    bitmap_info.bmiHeader.biWidth = m_pointSize * 2;
+    bitmap_info.bmiHeader.biHeight = m_pointSize * -2;
```
xezon: Was `~2` before. That would be -3. But WorldBuilder shows -2 here.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -226,10 +226,10 @@ void FontCharsClass::Create_GDI_Font(const char *font_name)
     }
 
     if (is_generals) {
-        font_width = ((double)-font_height / 2.5);
+        font_width = ((float)-font_height * 0.4f);
     }
 
-    m_widthReduction = std::clamp(font_height / ~7, 0, 4);
+    m_widthReduction = std::clamp(font_height / -8, 0, 4);
```
xezon: ~7 is -8, and WorldBuilder shows -8 here as well.

src/w3d/renderer/render2dsentence.cpp
tomsons26: -(m_pointSize * 2); actually

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -226,10 +226,10 @@ void FontCharsClass::Create_GDI_Font(const char *font_name)
     }
 
     if (is_generals) {
-        font_width = ((double)-font_height / 2.5);
+        font_width = ((float)-font_height * 0.4f);
```
tomsons26: / 2.5 is the same thing

src/w3d/renderer/render2dsentence.cpp
xezon: It's same result but yes I agree that reads better. I will change.

src/w3d/renderer/render2dsentence.cpp
xezon: I know. Do you want to keep the `2.5` ? World Builder shows * 0.4f

tomsons26: yes, i can just divide with/by 1 to find out what it should be so its not really a issue for comparing to pseudo
figuring out the real value is preferred for floats/doubles
2.5 is what the actual code would have had, it has a meaning, the multiplications compilers optimize divisions to can end up being cryptic

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -226,10 +226,10 @@ void FontCharsClass::Create_GDI_Font(const char *font_name)
...
     }
 
-    m_widthReduction = std::clamp(font_height / ~7, 0, 4);
+    m_widthReduction = std::clamp(-(font_height / 8), 0, 4);
```
tomsons26, line 232: where did - come from?

tomsons26 (approved):

PR #604, opened 2022-02-19
Double shadow buffers in W3DBufferManager

xezon:
This change doubles the shadow slots in W3DBufferManager. This allows twice as many meshes in a scene to cast shadows before the buffer space runs out.

### DIFF
--- a/src/platform/w3dengine/client/shadow/w3dbuffermanager.cpp
+++ b/src/platform/w3dengine/client/shadow/w3dbuffermanager.cpp
@@ -329,10 +329,9 @@ W3DBufferManager::W3DVertexBufferSlot *W3DBufferManager::Allocate_Slot_Storage(V
 
         W3DVertexBuffer *buf = m_W3DVertexBuffers[fvf_type];
 
-        // #TODO investigate
+        // TODO investigate
         // We allocate count for vertex buffer but size and start index is set by the original size
-        // As it stands this doesn't seem to make any sense.
-        // A 4 byte alignment on the buffer would be understandable but this...
+        // As it stands this doesn't seem to make any sense, i'd understand a 4 byte alignement on the buffer but this...
         int vb_size = std::max(8192, size);
 
         buf->m_DX8VertexBuffer =
@@ -340,7 +339,7 @@ W3DBufferManager::W3DVertexBufferSlot *W3DBufferManager::Allocate_Slot_Storage(V
         buf->m_format = fvf_type;
         buf->m_startFreeIndex = size;
         buf->m_size = vb_size;
-        // #BUGFIX Original didn't clear this
+        // BUGFIX Original didn't clear this
         buf->m_renderTaskList = nullptr;
 
         W3DVertexBufferSlot *slot = &m_W3DVertexBufferEmptySlots[m_numEmptyVertexSlotsAllocated++];
--- a/src/platform/w3dengine/client/shadow/w3dbuffermanager.cpp
+++ b/src/platform/w3dengine/client/shadow/w3dbuffermanager.cpp
@@ -329,9 +329,10 @@ W3DBufferManager::W3DVertexBufferSlot *W3DBufferManager::Allocate_Slot_Storage(V
 
         W3DVertexBuffer *buf = m_W3DVertexBuffers[fvf_type];
 
-        // TODO investigate
+        // #TODO investigate
         // We allocate count for vertex buffer but size and start index is set by the original size
-        // As it stands this doesn't seem to make any sense, i'd understand a 4 byte alignement on the buffer but this...
+        // As it stands this doesn't seem to make any sense.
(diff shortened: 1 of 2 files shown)


tomsons26:
When i messed with this when i wrote the buf manager initially eventually shadows started degrading and disappearing, its possible i changed something else too that caused that i guess.. So this needs careful verification of if there are regressions

Should be noted none of these changed in BFME1/2 :
MAX_IB_SIZES changed to 1024
MAX_VB_SIZES to 512
8192 in Allocate Slot Storage for VB changed to 16384

thus i suspect the issue of the shadows overflowing the buffers likely stems from the Volume Shadow classes themselves as i did also test with double values in WB when i wrote the code and it eventually throws the asserts anyway

xezon:
I would say let's leave this change open for now and re-investigate as soon as more surrounding code is available. Ideally we fix it so it never crashes even if the pools depleted. I tried adding nullptr checks where the asserts hit, but that just caused the game to crash somewhere else.

Edit: Crashes are fixed with #1101

xezon:
I have added better comments to the shadow buffer limits.

Doubling `MAX_NUMBER_SLOTS` is certainly legal and achieves the desired effect. It allows for more shadows to be drawn in the scene.

It is not yet entirely clear to me if `MAX_VERTEX_BUFFERS_CREATED` and `MAX_INDEX_BUFFERS_CREATED` need doubling too, but investigation indicated so. Decreasing these buffers will have similar effect as decreasing `MAX_NUMBER_SLOTS`, so likely they should be scaled uniformly.

We do have a shadow test map which could be useful for testing higher limits, but this test map currently does not load correctly in Thyme because of this bug: #1053 Needs fixing first.


> Should be noted none of these changed in BFME1/2 :
MAX_IB_SIZES changed to 1024
MAX_VB_SIZES to 512
8192 in Allocate Slot Storage for VB changed to 16384

@tomsons26
~~Can you please also check what the other constant in the other `W3DBufferManager::Allocate_Slot_Storage` has changed to in BFME?~~ This may be interesting to know.

Thyme:
```c++
int vb_size = std::max(8192, size);
int ib_size = std::max(32768, size);
```

BFME:
```c++
int vb_size = std::max(16384, size);
int ib_size = std::max(32768, size);
```

xezon:
BFME vs ZH
```
MAX_FVF          = 18                   ; same in ZH
MAX_INDEX_BUFFERS_CREATED  = 32         ; same in ZH
MAX_VERTEX_BUFFERS_CREATED  = 32        ; same in ZH
MAX_VB_SIZES     = 512                  ; 128 in ZH
MAX_IB_SIZES     = 1024                 ; 128 in ZH
MAX_NUMBER_SLOTS  = 4096                ; same in ZH
allocateSlotStorage_VB_val  = 16384     ; 8192 in ZH
allocateSlotStorage_IB_val  = 32768     ; same in ZH
```

xezon:
We know that 512 + 1 Scorpion Tanks in the scene will deplete the shadow memory pools.

I have now tested against 1024 Scorpion Tanks in the scene.

# Double MAX_NUMBER_SLOTS only with 1024 Scorpion Tanks

```
        MAX_NUMBER_SLOTS = 4096 * 2
        MAX_VERTEX_BUFFERS_CREATED = 32
        MAX_INDEX_BUFFERS_CREATED = 32
        MAX_VB_SIZES = 128
        MAX_IB_SIZES = 128
```
results in
```
-		g_theW3DBufferManager
		m_numEmptyVertexSlotsAllocated	4318	int
		m_numEmptyVertexBuffersAllocated	32	int
		m_numEmptyIndexSlotsAllocated	4323	int
		m_numEmptyIndexBuffersAllocated	22	int
```
`m_numEmptyVertexBuffersAllocated` now caps out at 32 and it cannot draw shadows for all meshes. `m_numEmptyVertexSlotsAllocated` should cap out at 8192.

# Double MAX_NUMBER_SLOTS and BUFFERS_CREATED with 1024 Scorpion Tanks

```
        MAX_NUMBER_SLOTS = 4096 * 2
        MAX_VERTEX_BUFFERS_CREATED = 32 * 2
        MAX_INDEX_BUFFERS_CREATED = 32 * 2
        MAX_VB_SIZES = 128
        MAX_IB_SIZES = 128
```
results in
```
-		g_theW3DBufferManager
		m_numEmptyVertexSlotsAllocated	8192	int
		m_numEmptyVertexBuffersAllocated	62	int
		m_numEmptyIndexSlotsAllocated	8192	int
		m_numEmptyIndexBuffersAllocated	41	int
```
`m_numEmptyVertexSlotsAllocated` now caps out at 8192 and all meshes cast shadows in the scene

shot_20240211_095540_1

# Conclusion

Scaling the following 3 limits uniformly is the correct strategy.
```
MAX_NUMBER_SLOTS
MAX_VERTEX_BUFFERS_CREATED
MAX_INDEX_BUFFERS_CREATED
```

PR #606, opened 2022-02-19
Switch all int types to int32_t

duncanspumpkin:
Little refactor to simplify things


jonwil:
I don't see that this is necessary.
On all platforms that we are ever likely to bump into (given that Thyme is never going to run on a Cray supercomputer) int and int32_t are identical in size and totally interchangeable.

int vs int32_t, uint vs uint32_t, short vs int16_t, ushort vs uint16_t, char vs int8_t, unsigned char vs uint8_t will make no difference to the code on the platforms we are ever going to run on.

duncanspumpkin:
We have already had situations where the wrong size is used due to long. It seems silly to have this issue when it's easily fixed by using the sized types

xezon:
Can we do this in chunks? Reviewing 560 files is quite much.

duncanspumpkin:
> Can we do this in chunks? Reviewing 560 files is quite much.

Certainly can do. I wasn't sure if it was easier for a review doing all of one type at a time or all types in one folder of files at a time

xezon:
Must not be submitted, if ever, before
* #586
* #620

duncanspumpkin:
> Must not be submitted, if ever, before
> 
> * [Add static asserts to Xfer to invoke extra caution if certain type si… #586](https://github.com/TheAssemblyArmada/Thyme/pull/586)
> * [Add type safe INI field parse macros #620](https://github.com/TheAssemblyArmada/Thyme/pull/620)

Why?

This just resolves all of it in a oner using a find and replace it can be done at any time and takes all of a couple of seconds to do + a little bit of massaging clang format. Unless you are just saying your prs will be a bit annoying to rebase?

PR #610, opened 2022-02-20
Add natvis for stl port vector

xezon:
stlvector


duncanspumpkin (approved):
for my own knowledge is there a guide on how to do these?

xezon:
https://docs.microsoft.com/en-us/visualstudio/debugger/create-custom-views-of-native-objects?view=vs-2022

PR #612, opened 2022-02-21
Partially implements IMEManager.

tomsons26:


src/game/client/gui/imemanagerinterface.h
```diff
@@ -0,0 +1,64 @@
...
+class Utf16String;
+
+class IMEManagerInterface : public SubsystemInterface
+{
+public:
+    virtual ~IMEManagerInterface() {}
+
+    virtual void Init() = 0;
+    virtual void Reset() = 0;
+    virtual void Update() = 0;
```
duncanspumpkin: these are just inherited and dont need to be here.

src/game/client/gui/imemanager.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+Utf16String *IMEManager::Get_Candidate(int index)
+{
+    if (m_candidateString && index >= 0 && index < m_candidateCount) {
+        return &m_candidateString[index];
+    }
+
+    // TODO this may be dangerous, need to find out if candidate is modified...
+    return (Utf16String *)&Utf16String::s_emptyString;
```
xezon, line 132: return `const Utf16String *`, remove the cast, then it is safe.

Also I suspect it returns reference, not pointer. Is unusual to return pointer this way if it is not expected to be null.

src/game/client/gui/imemanagerinterface.h
```diff
@@ -0,0 +1,60 @@
...
+    virtual int Get_Composition_Cursor_Position() = 0;
+
+    virtual int Get_Index_Base() = 0;
+
+    virtual int Get_Candidate_Count() = 0;
+    virtual Utf16String *Get_Candidate(int index) = 0;
+    virtual int Get_Selected_Candidate_Index() = 0;
+
+    virtual int Get_Candidate_Page_Size() = 0;
+    virtual int Get_Candidate_Page_Start() = 0;
```
xezon: Can we make these virtual get functions `const` ?

src/game/client/gui/imemanager.cpp
tomsons26: i suspect candidate is modified

PR #613, opened 2022-02-21
WndProc

duncanspumpkin:
Requires #612. WiP

### DIFF
--- a/src/game/main.cpp
+++ b/src/game/main.cpp
@@ -331,14 +331,18 @@ static LRESULT __stdcall Wnd_Proc(HWND window_handle, UINT message, WPARAM w_par
         }
         case WM_MOUSEWHEEL: {
             RECT window_rect{};
+            // Note that this is different to MOUSEMOVE in that its the WindowRect
             GetWindowRect(g_applicationHWnd, &window_rect);
             POINT mouse_position{ GET_X_LPARAM(l_param), GET_Y_LPARAM(l_param) };
             if (mouse_position.x < window_rect.left || mouse_position.x > window_rect.right
                 || mouse_position.y < window_rect.top || mouse_position.y > window_rect.bottom) {
                 return 0;
             }
+            if (g_theWin32Mouse != nullptr) {
+                g_theWin32Mouse->Add_Win32_Event(message, w_param, l_param, g_theMessageTime);
+            }
+            return 0;
         }
-            //[[fallthrough]]
         case WM_LBUTTONDOWN:
         case WM_LBUTTONDBLCLK:
         case WM_LBUTTONUP:
--- a/src/game/main.cpp
+++ b/src/game/main.cpp
@@ -13,18 +13,26 @@
  *            LICENSE
  */
 #include "main.h"
+#include "audiomanager.h"
+#include "copyprotect.h"
 #include "cpudetect.h"
 #include "crashhandler.h"
 #include "critsection.h"
+#include "gameengine.h"
 #include "gamemain.h"
 #include "gamememory.h"
 #include "gitverinfo.h"
+#include "imemanager.h"
+#include "keyboard.h"
 #include "memdynalloc.h"
 #include "mempool.h"
+#include "messagestream.h"
 #include "stackdump.h"
 #include "unicodestring.h"
 #include "version.h"
+#include "w3ddisplay.h"
 #include "win32compat.h"
+#include "win32mouse.h"
 #include <algorithm>
 #include <captainslog.h>
 #include <cstdio>
@@ -51,6 +59,7 @@ using std::rename;
 #ifdef PLATFORM_WINDOWS
 #include <direct.h>
 #include <shellapi.h>
+#include <windowsx.h>
 #include <wingdi.h>
 #include <winuser.h>
 
@@ -65,6 +74,7 @@ using std::rename;
 // Some globals, replace exe versions with own once all code that uses them is implemented.
 extern bool &g_gameIsWindowed;
 extern bool &g_gameNotFullscreen;
+extern bool &g_gameActive;
 extern bool &g_creatingWindow;
 extern HGDIOBJ &g_splashImage;
 extern HINSTANCE &g_applicationHInstance;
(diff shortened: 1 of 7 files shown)


src/game/main.cpp
```diff
@@ -162,6 +171,203 @@ void Check_Windowed(int argc, char *argv[])
 #endif
 }
 
+// zh: 0x0073C740
+// TODO: Move to Display
+void sub_73C740(bool activateState)
+{
+    Call_Function<void, bool>(0x0073C740, activateState);
```
tomsons26: this is Reset_D3D_Device, its in w3ddisplay

src/game/main.cpp
```diff
@@ -162,6 +171,203 @@ void Check_Windowed(int argc, char *argv[])
...
+                auto hDC = BeginPaint(hWnd, &ps);
+                if (g_splashImage != nullptr) {
+                    const auto old_dc = SaveDC(hDC);
+                    auto compat_dc = CreateCompatibleDC(hDC);
+                    auto ha = SelectObject(compat_dc, g_splashImage);
+                    BitBlt(hDC, 0, 0, 800, 600, compat_dc, 0, 0, SRCCOPY);
+                    SelectObject(compat_dc, ha);
+                    DeleteDC(compat_dc);
+                    RestoreDC(hDC, old_dc);
+                }
```
tomsons26, line 250: EndPaint missing

src/game/main.cpp
```diff
@@ -162,6 +171,203 @@ void Check_Windowed(int argc, char *argv[])
...
+            if (g_theGameEngine->Get_Quitting()) {
+                if (g_theKeyboard != nullptr) {
+                    g_theKeyboard->Reset_Keys();
+                }
+                if (g_theWin32Mouse != nullptr) {
+                    g_theWin32Mouse->Set_Ignore_Events(false);
+                }
+                break;
+            }
+            //[[fallthrough]]
```
tomsons26: probably just copy it into above instead

src/game/main.cpp
```diff
@@ -162,6 +171,203 @@ void Check_Windowed(int argc, char *argv[])
...
+                case SC_MOVE:
+                case SC_KEYMENU:
+                case SC_MONITORPOWER:
+                    if (!g_gameIsWindowed) {
+                        return 0; // vanilla returned 1
+                    }
+            }
+            break;
+
+        case WM_MOUSEMOVE:
```
tomsons26: seems like if you pass Add_Win32_Event msg instead of WM_MOUSEMOVE its handled the same as mouse wheel?

src/game/main.cpp
```diff
@@ -162,6 +171,203 @@ void Check_Windowed(int argc, char *argv[])
...
+                if (g_splashImage != nullptr) {
+                    const auto old_dc = SaveDC(hDC);
+                    auto compat_dc = CreateCompatibleDC(hDC);
+                    auto ha = SelectObject(compat_dc, g_splashImage);
+                    BitBlt(hDC, 0, 0, 800, 600, compat_dc, 0, 0, SRCCOPY);
+                    SelectObject(compat_dc, ha);
+                    DeleteDC(compat_dc);
+                    RestoreDC(hDC, old_dc);
+                }
+                return 0; // vanilla returned 1
```
tomsons26: BUGFIX

src/game/main.cpp
duncanspumpkin: yes do you want me to combine them?

tomsons26: yeah

duncanspumpkin: Ah forgot one is using clientrect the other is using windowrect so they are subtly different.

src/game/main.cpp
```diff
@@ -162,6 +173,201 @@ void Check_Windowed(int argc, char *argv[])
...
+                if (g_theKeyboard != nullptr) {
+                    g_theKeyboard->Reset_Keys();
+                }
+                if (g_theWin32Mouse != nullptr) {
+                    g_theWin32Mouse->Set_Ignore_Events(false);
+                }
+                break;
+            }
+            g_theMessageStream->Append_Message(GameMessage::MessageType::MSG_META_DEMO_INSTANT_QUIT);
+            return 0;
```
xezon, line 267: It can return `FALSE` for 0 since it already returns `TRUE` for 1.

src/game/main.cpp
```diff
@@ -162,6 +173,201 @@ void Check_Windowed(int argc, char *argv[])
...
 
+#if defined PLATFORM_WINDOWS && defined GAME_DLL
+// zh: 0x00401000
+static LRESULT __stdcall Wnd_Proc(HWND window_handle, UINT message, WPARAM w_param, LPARAM l_param)
+{
+    if (g_theIMEManager != nullptr && g_theIMEManager->Service_IME_Message(window_handle, message, w_param, l_param)) {
+        return g_theIMEManager->Result();
+    }
+
+    CopyProtect::checkForMessage(message, l_param);
```
xezon, line 184: Does CopyProtect need calling in Thyme? Does it do anything for us?

src/game/main.cpp
duncanspumpkin: Winapi states that this returns 0 for handling with message.

src/game/main.cpp
duncanspumpkin: Its just an empty function so it does nothing. I think it makes sense to just leave it calling it like this for posterity.

src/game/main.cpp
```diff
@@ -162,6 +173,201 @@ void Check_Windowed(int argc, char *argv[])
...
+        case WM_MOUSEWHEEL: {
+            RECT window_rect{};
+            GetWindowRect(g_applicationHWnd, &window_rect);
+            POINT mouse_position{ GET_X_LPARAM(l_param), GET_Y_LPARAM(l_param) };
+            if (mouse_position.x < window_rect.left || mouse_position.x > window_rect.right
+                || mouse_position.y < window_rect.top || mouse_position.y > window_rect.bottom) {
+                return 0;
+            }
+        }
+            //[[fallthrough]]
```
xezon: Instead of the fallthrough, how about we just copy up what it does below? This way we cut this entanglement.

xezon (approved):
Logically looks ok, although the function is not that easy to look at.

PR #618, opened 2022-02-23
OpenAL audio backend

feliwir:
- [x] Enumerate OpenAL devices list
- [x] Work on alternate file loading (see #619 )

PR #619, opened 2022-02-23
Work on FFmpeg audio cache

feliwir:
This is a requirement for #618, since OpenAL has no file handling by it's own
- [x] Replace MilesAudioCacheFile when compiling with FFmpeg: Decode audio data into a "WaveFile" structure and pass that to Miles


src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,356 @@
...
+#include "audiomanager.h"
+#include "ffmpegaudiofilecache.h"
+#include "filesystem.h"
+#include <captainslog.h>
+#include <list>
+
+struct WavHeader
+{
+    /* RIFF Chunk Descriptor */
+    uint8_t RIFF[4] = { 'R', 'I', 'F', 'F' }; // RIFF Header Magic header
```
OmniBlade: Member name style should conform to project standards, normally `member_name` for public access structs. At the very least I'd like it consistent within a struct.

OmniBlade (changes requested):

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: fixed

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,364 @@
...
+bool FFmpegAudioFileCache::Open_FFmpeg_Contexts(FFmpegOpenAudioFile *file, unsigned char *file_data, uint32_t file_size)
+{
+    // FFmpeg setup
+    int ret = 0;
+    file->fmt_ctx = avformat_alloc_context();
+    if (!file->fmt_ctx) {
+        captainslog_error("Failed to alloc AVFormatContext");
+        return false;
+    }
+    file->avio_ctx = avio_alloc_context((unsigned char *)file_data, file_size, 0, nullptr, nullptr, nullptr, nullptr);
```
xezon: Remove C cast.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,449 @@
...
+        Release_Open_Audio(&it->second);
+    }
+}
+
+/**
+ * Read an FFmpeg packet from file
+ */
+int FFmpegAudioFileCache::Read_FFmpeg_Packet(void *opaque, uint8_t *buf, int buf_size)
+{
+    File *file = (File *)opaque;
```
xezon: `static_cast`

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,449 @@
...
+    int read = file->Read(buf, buf_size);
+    return read;
+}
+
+/**
+ * Open all the required FFmpeg handles for a required file.
+ */
+bool FFmpegAudioFileCache::Open_FFmpeg_Contexts(FFmpegOpenAudioFile *file, File *f)
+{
+#ifndef NDEBUG
```
xezon: This define is a double negative. Hard to read. Can we fix that?

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,449 @@
...
+{
+    File *file = (File *)opaque;
+    int read = file->Read(buf, buf_size);
+    return read;
+}
+
+/**
+ * Open all the required FFmpeg handles for a required file.
+ */
+bool FFmpegAudioFileCache::Open_FFmpeg_Contexts(FFmpegOpenAudioFile *file, File *f)
```
xezon: `file` and `f` do not look like descriptive names on first sight.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,449 @@
...
+    // FFmpeg setup
+    int ret = 0;
+    file->fmt_ctx = avformat_alloc_context();
+    if (!file->fmt_ctx) {
+        captainslog_error("Failed to alloc AVFormatContext");
+        return false;
+    }
+
+    size_t avio_ctx_buffer_size = 0x2000;
+    file->avio_ctx_buffer = (uint8_t *)av_malloc(avio_ctx_buffer_size);
```
xezon: `static_cast`

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,449 @@
...
+    if (!file->avio_ctx_buffer) {
+        captainslog_error("Failed to alloc AVIOContextBuffer");
+        Close_FFmpeg_Contexts(file);
```
xezon: `avio_ctx_buffer` seems to be not deleted on this code path

Edit: Inside `Close_FFmpeg_Contexts` I mean.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,449 @@
...
+    if (file->avio_ctx) {
+        av_freep(&file->avio_ctx);
+    }
+}
+
+void FFmpegAudioFileCache::Fill_Wave_Data(FFmpegOpenAudioFile *file)
+{
+    WavHeader wav;
+    wav.chunk_size = file->data_size - 8;
+    wav.subchunk2_size = file->data_size - 44;
```
xezon: What do these magic number here mean?

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,449 @@
...
+
+/**
+ * Opens an audio file. Reads from the cache if available or loads from file if not.
+ */
+uint8_t *FFmpegAudioFileCache::Open_File(const Utf8String &filename)
+{
+    ScopedMutexClass lock(&m_mutex);
+
+    // Load the file from disk
+    File *file = g_theFileSystem->Open(filename, File::READ | File::BINARY);
```
xezon: I suggest using `File::BUFFERED` flag too.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,449 @@
...
+    wav.block_align = file->codec_ctx->channels * (wav.bits_per_sample / 8);
+    memcpy(file->wave_data, &wav, sizeof(WavHeader));
+}
+
+/**
+ * Opens an audio file. Reads from the cache if available or loads from file if not.
+ */
+uint8_t *FFmpegAudioFileCache::Open_File(const Utf8String &filename)
+{
+    ScopedMutexClass lock(&m_mutex);
```
xezon, line 260: This looks scary, because the entire file read is locked. This means huge performance penalty in threaded context. What exactly needs locking here?

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,449 @@
...
+    if (file == nullptr) {
+        if (filename.Is_Not_Empty()) {
+            captainslog_warn("Missing audio file '%s', could not cache.", filename.Str());
+        }
+
+        return nullptr;
+    }
+
+    FFmpegOpenAudioFile open_audio;
+    open_audio.wave_data = (uint8_t *)av_malloc(sizeof(WavHeader));
```
xezon: `static_cast`

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,449 @@
...
+    open_audio.data_size = sizeof(WavHeader);
+
+    if (!Open_FFmpeg_Contexts(&open_audio, file)) {
+        captainslog_warn("Failed to load audio file '%s', could not cache.", filename.Str());
+        return nullptr;
```
xezon, line 292: Is `open_audio.wave_data` freed here?

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+    if (!Decode_FFmpeg(&open_audio)) {
+        captainslog_warn("Failed to decode audio file '%s', could not cache.", filename.Str());
+        Close_FFmpeg_Contexts(&open_audio);
+        return nullptr;
```
xezon, line 300: Is `open_audio.wave_data` freed here?

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,449 @@
...
+    m_cacheMap[filename] = open_audio;
+
+    return open_audio.wave_data;
+}
+/**
+ * Opens an audio file for an event. Reads from the cache if available or loads from file if not.
+ */
+uint8_t *FFmpegAudioFileCache::Open_File(AudioEventRTS *audio_event)
+{
+    ScopedMutexClass lock(&m_mutex);
```
xezon, line 328: What is this lock for?

xezon (changes requested):
Please reduce lock time or explain the locks and please investigate potential memory leaks in various places.

I reviewed only half way. Shall continue when below things are addressed.

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Those are the the offsets of the fields inside the WAV struct. Will use `offsetof`

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: I copied that from MilesAudioFileCache

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: I copied that from MilesAudioFileCache

feliwir:
> Please reduce lock time or explain the locks and please investigate potential memory leaks in various places.
> 
> I reviewed only half way. Shall continue when below things are addressed.

Thanks for the review! Those locks are taken 1:1 from MilesAudioFileCache. Not sure if we need them to be honest

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Good catch, i added that later. Fix it

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: I can do that, however FFmpeg already does buffering by itself, so not sure if there's a large benefit

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: This happens in `Release_Open_Audio`

feliwir: Good catch

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: No, it wans't. Good catch

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Used `open_audio` and `file` now

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: I think `NDEBUG` is the most portable flag to detect debug/release builds

xezon: In Thyme we can set our own. The Problem with NDEBUG is that you end up with "is not not debug" which equates to "is debug". I think we have `GAME_DEBUG` macro no?

src/platform/audio/ffmpegaudiofilecache.cpp
xezon: You can run performance profiler and check. If you look at `FFmpegAudioFileCache::Read_FFmpeg_Packet` you can see that it reads in chunks. Each of these reads incurs File IO without buffering, which can cause huge performance penalty depending on chunk size. How large is chunk size? Ideally it should be 512 bytes or larger.

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Yep, used that now

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -129,14 +128,27 @@ bool FFmpegAudioFileCache::Open_FFmpeg_Contexts(FFmpegOpenAudioFile *open_audio,
...
 
-    ret = avcodec_open2(open_audio->fmt_ctx->streams[0]->codec, input_codec, NULL);
+    open_audio->codec_ctx = avcodec_alloc_context3(input_codec);
+    if (!open_audio->codec_ctx) {
+        captainslog_error("Could not allocate codec context");
+        Close_FFmpeg_Contexts(open_audio);
+        return false;
+    }
+
+    ret = avcodec_parameters_to_context(open_audio->codec_ctx, open_audio->fmt_ctx->streams[0]->codecpar);
```
xezon: Does `open_audio->fmt_ctx->streams[0]` need checking before use?

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,468 @@
...
+        if (result < 0) {
+            captainslog_error("Failed to send audio packet to decoder.");
+            return false;
+        }
+        result = avcodec_receive_frame(file->codec_ctx, frame);
+        // Check if this was a real error or we just need more data
+        if (result < 0 && result != AVERROR(EAGAIN) && result != AVERROR_EOF) {
+            captainslog_error("Failed to receive audio frame from decoder.");
+            return false;
+        } else if (result >= 0) {
```
xezon: Is `if (result >= 0)` necessary here?

src/platform/audio/ffmpegaudiofilecache.cpp
xezon: Ok. Fair enough. Might be subject for optimization later.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,468 @@
...
+    }
+
+    m_cacheMap[filename] = open_audio;
+
+    return open_audio.wave_data;
+}
+/**
+ * Opens an audio file for an event. Reads from the cache if available or loads from file if not.
+ */
+uint8_t *FFmpegAudioFileCache::Open_File(AudioEventRTS *audio_event)
```
xezon: Code of this function looks very similar to the `Open_File` function above. Can you refactor these to avoid redundant code? That said, why do we need 2 `Open_File` functions? `MilesAudioFileCache` only has one.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,468 @@
...
+
+    m_cacheMap[filename] = open_audio;
+
+    return open_audio.wave_data;
+}
+
+/**
+ * Closes a file, reducing the references to it. Does not actually free the cache.
+ */
+void FFmpegAudioFileCache::Close_File(uint8_t *file)
```
xezon: According to `AudioManager` interface, file handle is expected to be `void *`. See `virtual void Close_Any_Sample_Using_File(const void *handle) = 0;` Maybe adhere to same user facing type in `FFmpegAudioFileCache` ?

src/platform/audio/ffmpegaudiofilecache.h
```diff
@@ -0,0 +1,87 @@
...
+struct FFmpegOpenAudioFile
+{
+    // FFmpeg handles
+    AVFormatContext *fmt_ctx = nullptr;
+    AVIOContext *avio_ctx = nullptr;
+    AVCodecContext *codec_ctx = nullptr;
+    uint8_t *wave_data = nullptr;
+    int ref_count = 0;
+    int data_size = 0;
+    const AudioEventInfo *audio_event_info;
```
xezon: All members are initialized except for this one.

src/platform/audio/ffmpegaudiofilecache.cpp
xezon: `AudioEventRTS *audio_event` function argument should be const (and `Is_Positional_Audio()` function too + argument of `MilesAudioFileCache::Open_File`).

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,468 @@
...
+    auto it = m_cacheMap.find(filename);
+
+    if (it != m_cacheMap.end()) {
+        ++it->second.ref_count;
+
+        return it->second.wave_data;
+    }
+
+    // Load the file from disk
+    File *file = g_theFileSystem->Open(filename, File::READ | File::BINARY);
```
xezon: Maybe add `File::BUFFERED` ?

src/platform/audio/ffmpegaudiofilecache.h
```diff
@@ -0,0 +1,87 @@
...
+    inline unsigned Get_Max_Size() const { return m_maxSize; }
+    inline unsigned Get_Current_Size() const { return m_currentSize; }
+
+private:
+    bool Free_Space_For_Sample(const FFmpegOpenAudioFile &file);
+    void Release_Open_Audio(FFmpegOpenAudioFile *file);
+
+    bool Open_FFmpeg_Contexts(FFmpegOpenAudioFile *file, File *f);
+    bool Decode_FFmpeg(FFmpegOpenAudioFile *file);
+    void Close_FFmpeg_Contexts(FFmpegOpenAudioFile *file);
```
xezon: Declared function argument names do not appear to match defined function argument names.

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Before there is a check:
```c
    if (open_audio->fmt_ctx->nb_streams != 1) {
        captainslog_error("Expected exactly one audio stream per file");
        Close_FFmpeg_Contexts(open_audio);
        return false;
    }
```

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: I needed one function that works completly without AudioEvents to do proper testing - i reducing duplicate code as much as possible by creating seperate functions (`Open_FFmpeg...` etc.)

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: I think "void*" should be avoided if possible. 
`Open_File` returns a byte array that contains the WAV data. The same byte array is used to close the file again. Using `void*` is confusing in that regard that it doesn't show what that pointer really is - WAVE data

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Done

src/platform/audio/ffmpegaudiofilecache.h
feliwir: Done

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Reverted

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Yes

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,485 @@
...
+}
+
+/**
+ * Closes any playing instances of an audio file and then frees the memory for it.
+ */
+void FFmpegAudioFileCache::Release_Open_Audio(FFmpegOpenAudioFile *open_audio)
+{
+    // Close any playing samples that use this data.
+    if (open_audio->ref_count && g_theAudio) {
+        g_theAudio->Close_Any_Sample_Using_File(open_audio);
```
xezon: `open_audio->wave_data`

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,485 @@
...
+        if (result < 0 && result != AVERROR(EAGAIN) && result != AVERROR_EOF) {
+            av_strerror(result, error_buffer, sizeof(error_buffer));
+            captainslog_error("Failed 'avcodec_receive_frame': %s", error_buffer);
+            return false;
+        } else if (result >= 0) {
+            int frame_data_size = av_samples_get_buffer_size(
+                NULL, file->codec_ctx->channels, frame->nb_samples, file->codec_ctx->sample_fmt, 1);
+            file->wave_data = static_cast<uint8_t *>(av_realloc(file->wave_data, file->data_size + frame_data_size));
+            memcpy(file->wave_data + file->data_size, frame->data[0], frame_data_size);
+            file->data_size += frame_data_size;
```
xezon, line 208: Why is this adding to `file->data_size` ? Does this mean it could already have a size and potentially also a `file->wave_data` that needs cleaning before? In this loop `file->wave_data` just keeps getting assigned, but there is no free. Does it maybe need to break out here? I don't quite understand the intent here, but it looks broken if `av_read_frame` reads twice successfully.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,485 @@
...
+    return true;
+}
+
+/**
+ * Closes any playing instances of an audio file and then frees the memory for it.
+ */
+void FFmpegAudioFileCache::Release_Open_Audio(FFmpegOpenAudioFile *open_audio)
+{
+    // Close any playing samples that use this data.
+    if (open_audio->ref_count && g_theAudio) {
```
xezon: Style: `open_audio->ref_count != 0`

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,485 @@
...
+        g_theAudio->Close_Any_Sample_Using_File(open_audio);
+    }
+
+    // Deallocate the data buffer depending on how it was allocated.
+    if (open_audio->wave_data) {
```
xezon: Style: `open_audio->wave_data != nullptr`

tests/test_audiofilecache.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    auto filepath = Utf8String(TESTDATA_PATH) + "/audio/pcm1644m.wav";
+    void *data = cache.Open_File(filepath);
+    // We expect this to fail since our cache size is 0
+    EXPECT_EQ(data, static_cast<uint8_t *>(nullptr));
+
+    // Use something that can hold our file
+    cache.Set_Max_Size(0xFFFFF);
+    data = cache.Open_File(filepath);
+    EXPECT_NE(data, static_cast<uint8_t *>(nullptr));
```
xezon: Can remove the static_cast's here.

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: I use `av_realloc` here, which will invalidate/free the previous pointer. It works similar to C `realloc`. And yes, this is / can be called multiple times. We might want to switch this to a growing vector if this ever becomes a performance concern

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: no, i think it was right like that. See milesaudiofilecache

tests/test_audiofilecache.cpp
feliwir: Done

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Done

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Done

OmniBlade (commented):
After a bit of testing, it seems that while initially audio does work, after a while it stops, not sure where the issue is at a glance. Also, a few of the files show as not ending on a new line which is a minor nitpick, but if you need to make more changes anyway it would be nice to have them fixed too.

src/platform/audio/ffmpegaudiofilecache.cpp
xezon: Yes but it is wrong.

feliwir: Fixed

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -324,9 +334,11 @@ void *FFmpegAudioFileCache::Open_File(AudioEventRTS *audio_event)
         case 3:
             return nullptr;
         default:
-            break;
+            return nullptr;
```
xezon: Why does it return here now?

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -77,6 +79,7 @@ bool FFmpegAudioFileCache::Open_FFmpeg_Contexts(FFmpegOpenAudioFile *open_audio,
     av_log_set_level(AV_LOG_INFO);
 #endif
 
+    // REMOVE: this is required for FFmpeg older than 4.0 -> deprecated afterwards though
```
xezon: Is there a macro for the FFMPEG version? If so, then this call can be guarded according to the version.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -337,7 +349,7 @@ void *FFmpegAudioFileCache::Open_File(AudioEventRTS *audio_event)
     }
 
     // Load the file from disk
-    File *file = g_theFileSystem->Open(filename, File::READ | File::BINARY | File::BUFFERED);
+    File *file = g_theFileSystem->Open(filename, File::READ | File::BINARY);
```
xezon: `File::BUFFERED` was removed again. Was there something wrong with it?

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Well, the break here makes no sense, but i can revert it

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Unfortunately, there's only a runtime check - it's not possible to check the version during compiletime to my knowledge

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: No, it worked fine. I was trying to figure out an error and `MilesAudioFileCache` doesn't have this

src/platform/audio/ffmpegaudiofilecache.cpp
OmniBlade: Looks to me like there are API version macros in version.h for each library?

feliwir: Added a version check

feliwir:
> After a bit of testing, it seems that while initially audio does work, after a while it stops, not sure where the issue is at a glance. Also, a few of the files show as not ending on a new line which is a minor nitpick, but if you need to make more changes anyway it would be nice to have them fixed too.

Fixed - the cache was not being emptied correctly

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,537 @@
...
+unsigned FFmpegAudioFileCache::Free_Space()
+{
+    std::list<Utf8String> to_free;
+    unsigned freed = 0;
+
+    // First check for samples that don't have any references.
+    for (const auto &cached : m_cacheMap) {
+        if (cached.second.ref_count == 0) {
+            to_free.push_back(cached.first);
+            freed += cached.second.data_size;
```
duncanspumpkin, line 452: this is minor but it doens't quite make sense to add to the freed count at this point it should be done in the second loop where its actually freeing it.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,537 @@
...
+    ScopedMutexClass lock(&m_mutex);
+    m_maxSize = size;
+}
+
+/**
+ * Attempts to free space by releasing files with no references
+ */
+unsigned FFmpegAudioFileCache::Free_Space()
+{
+    std::list<Utf8String> to_free;
```
duncanspumpkin, line 445: Using a list for this is odd we add to it at the end always and destruct it all at once so none of the benefits of std::list with a number of negatives. Use a std::vector instead.

xezon: I think he uses std::list because that is what the original does in the MilesAudioCache. But yes, can change container.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,537 @@
...
+    for (const auto &file : to_free) {
+        auto to_remove = m_cacheMap.find(file);
+
+        if (to_remove != m_cacheMap.end()) {
+            Release_Open_Audio(&to_remove->second);
+            m_currentSize -= to_remove->second.data_size;
+            m_cacheMap.erase(to_remove);
+        }
+    }
+
```
duncanspumpkin, line 470: I question if this is the most efficient way of doing this. You are creating a list of items to remove doing allocations to create that list then remove them. It would be easier if you just did:
```cpp
// Free any unreferenced audio
for (const auto &cached : m_cacheMap) {
        if (cached.second.ref_count == 0) {
            freed += cached.second.data_size;
            Release_Open_Audio(&to_remove->second);
            m_currentSize -= to_remove->second.data_size;
        }
}
// Erase the unreferenced entries
for(auto it =m_cacheMap.begin(); it != m_cacheMap.end(); ) {
    if(it->second.ref_count == 0){
        it = m_cacheMap.erase(it);
    }
    else{
        ++it;
    }
}
```

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,537 @@
...
+    return freed;
+}
+
+/**
+ * Attempts to free space for a file by releasing files with no references and lower priority sounds.
+ */
+bool FFmpegAudioFileCache::Free_Space_For_Sample(const FFmpegOpenAudioFile &file)
+{
+    captainslog_assert(m_currentSize >= m_maxSize); // Assumed to be called only when we need more than allowed.
+    std::list<Utf8String> to_free;
```
duncanspumpkin, line 480: again a std::vector would be better we dont need a list

src/platform/audio/ffmpegaudiofilecache.cpp
xezon: I think it is ok if FFmpegAudioCache is similar to MilesAudioCache for now. No urgent need to deviate just yet. Maybe both can be refactored at the same time.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,537 @@
...
+    for (const auto &cached : m_cacheMap) {
+        if (cached.second.ref_count == 0) {
+            to_free.push_back(cached.first);
+            freed += cached.second.data_size;
+
+            if (freed >= required) {
+                break;
+            }
+        }
+    }
```
duncanspumpkin: You could without much modification change Free_Space to take a limit of how much space you want freed then combine the two functions mostly.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,537 @@
...
+void FFmpegAudioFileCache::Set_Max_Size(unsigned size)
+{
+    ScopedMutexClass lock(&m_mutex);
+    m_maxSize = size;
+}
+
+/**
+ * Attempts to free space by releasing files with no references
+ */
+unsigned FFmpegAudioFileCache::Free_Space()
```
duncanspumpkin: All the other functions have a mutex should there be one for this (and the other free function)?

feliwir: The free functions are called from functions that already lock themselves - This free function is currently only used in tests

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: This is taken from `MilesAudioCache - will refactor the design in a seperate PR

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,537 @@
...
+ *
+ * @copyright Thyme is free software: you can redistribute it and/or
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+
+extern "C" {
```
xezon, line 16: Why is this `extern "C"` ?

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,537 @@
...
+#endif
+
+// This is required for FFmpeg older than 4.0 -> deprecated afterwards though
+#if LIBAVFORMAT_VERSION_MAJOR < 58
+    av_register_all();
+#endif
+
+    // FFmpeg setup
+    int result = 0;
+    char error_buffer[1024];
```
xezon: I suggest to move the stack allocations just before each `av_strerror` call.
Same in next function.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,537 @@
...
+    av_log_set_level(AV_LOG_INFO);
+#endif
+
+// This is required for FFmpeg older than 4.0 -> deprecated afterwards though
+#if LIBAVFORMAT_VERSION_MAJOR < 58
+    av_register_all();
+#endif
+
+    // FFmpeg setup
+    int result = 0;
```
xezon: This declaration can be moved down to first time use.

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,537 @@
...
+            break;
+    }
+
+    captainslog_trace("FFmpegAudioFileCache: opening file %s", filename.Str());
+
+    // Try to find existing data for this file to avoid loading it if unneeded.
+    auto it = m_cacheMap.find(filename);
+
+    if (it != m_cacheMap.end()) {
+        ++it->second.ref_count;
```
xezon: Minor: `++(it->second.ref_count);` will make it easier to read, as one could think on first sight that it is `it` being incremented here.

src/platform/audio/ffmpegaudiofilecache.cpp
xezon: Where is this function called? I can't find it here with Browser Text Search. Also, this function is `public`. Shouldn't it be private if it is meant to be called internally only?

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Because the headers can't be included from C++. See https://trac.ffmpeg.org/wiki/Including%20FFmpeg%20headers%20in%20a%20C%2B%2B%20application

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: It's called by the audiofilecache tests. See `test_audiofilecache.cpp`

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,535 @@
...
+    wav.samples_per_sec = open_audio->codec_ctx->sample_rate;
+    wav.bytes_per_sec = open_audio->codec_ctx->sample_rate * open_audio->codec_ctx->channels * wav.bits_per_sample / 8;
+    wav.block_align = open_audio->codec_ctx->channels * wav.bits_per_sample / 8;
+    memcpy(open_audio->wave_data, &wav, sizeof(WavHeader));
+}
+
+/**
+ * Opens an audio file. Reads from the cache if available or loads from file if not.
+ */
+void *FFmpegAudioFileCache::Open_File(const Utf8String &filename)
```
xezon: `AudioDataHandle`

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,535 @@
...
+
+    captainslog_trace("FFmpegAudioFileCache: opening file %s", filename.Str());
+
+    // Try to find existing data for this file to avoid loading it if unneeded.
+    auto it = m_cacheMap.find(filename);
+
+    if (it != m_cacheMap.end()) {
+        ++(it->second.ref_count);
+
+        return it->second.wave_data;
```
xezon: `static_cast<AudioDataHandle>`

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,535 @@
...
+        captainslog_warn("Cannot play audio file since cache is full: %s", filename.Str());
+        m_currentSize -= open_audio.data_size;
+        Release_Open_Audio(&open_audio);
+
+        return nullptr;
+    }
+
+    m_cacheMap[filename] = open_audio;
+
+    return open_audio.wave_data;
```
xezon: `static_cast<AudioDataHandle>`

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,535 @@
...
+
+    captainslog_trace("FFmpegAudioFileCache: opening file %s", filename.Str());
+
+    // Try to find existing data for this file to avoid loading it if unneeded.
+    auto it = m_cacheMap.find(filename);
+
+    if (it != m_cacheMap.end()) {
+        ++(it->second.ref_count);
+
+        return it->second.wave_data;
```
xezon: `static_cast<AudioDataHandle>`

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,535 @@
...
+        captainslog_warn("Cannot play audio file since cache is full: %s", filename.Str());
+        m_currentSize -= open_audio.data_size;
+        Release_Open_Audio(&open_audio);
+
+        return nullptr;
+    }
+
+    m_cacheMap[filename] = open_audio;
+
+    return open_audio.wave_data;
```
xezon: `static_cast<AudioDataHandle>`

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,535 @@
...
+}
+
+/**
+ * Closes a file, reducing the references to it. Does not actually free the cache.
+ */
+void FFmpegAudioFileCache::Close_File(void *file)
```
xezon: `AudioDataHandle`

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,535 @@
...
+{
+    if (file == nullptr) {
+        return;
+    }
+
+    ScopedMutexClass lock(&m_mutex);
+
+    for (auto it = m_cacheMap.begin(); it != m_cacheMap.end(); ++it) {
+        if (it->second.wave_data == file) {
```
xezon: `static_cast<uint8_t *>`

src/platform/audio/ffmpegaudiofilecache.h
```diff
@@ -0,0 +1,91 @@
...
+typedef std::unordered_map<const Utf8String, FFmpegOpenAudioFile, rts::hash<Utf8String>, std::equal_to<Utf8String>>
+    ffmpegaudiocachemap_t;
+#endif
+
+class FFmpegAudioFileCache
+{
+public:
+    FFmpegAudioFileCache() : m_maxSize(0), m_currentSize(0), m_mutex("AudioFileCacheMutex") {}
+    virtual ~FFmpegAudioFileCache();
+    void *Open_File(AudioEventRTS *file);
```
xezon: `AudioDataHandle`

src/platform/audio/ffmpegaudiofilecache.h
```diff
@@ -0,0 +1,91 @@
...
+    void *Open_File(const Utf8String &filename);
+
+    void Close_File(void *file);
```
xezon: `AudioDataHandle`

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -0,0 +1,535 @@
...
+/**
+ * Decode the input data and append it to our wave data stream
+ */
+bool FFmpegAudioFileCache::Decode_FFmpeg(FFmpegOpenAudioFile *file)
+{
+    AVPacket *packet = av_packet_alloc();
+    AVFrame *frame = av_frame_alloc();
+
+    int result = 0;
+    char error_buffer[1024];
```
xezon: You can move `char error_buffer[1024];` stack allocation closer to the fail scopes where it is used.

xezon (commented):
If you get latest code, you can now use `AudioDataHandle`.

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Done

src/platform/audio/ffmpegaudiofilecache.h
feliwir: Done

src/platform/audio/ffmpegaudiofilecache.h
feliwir: Done

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Done

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Taken from MilesAudioFileCache

duncanspumpkin (approved):
Can you add an issue stating: refactor both miles and this to a base class and remove lists in favour of vectors.

feliwir:
> Can you add an issue stating: refactor both miles and this to a base class and remove lists in favour of vectors.

Done

OmniBlade (approved):

xezon (approved):

PR #614, opened 2022-02-22
Fix some INI field offsets & parsing for standalone

feliwir:
- [x] GlobalData
- [x] WaterTransparency 
- [x] Weather 
- [x] Science (WIP #359 )

PR #620, opened 2022-02-23
Add type safe INI field parse macros

xezon:
I suggest to review this through looking at the individual commits, which makes it easier than looking at all of it at once.

Must not be started before submission of 
* #614


src/game/client/anim2d.cpp
```diff
@@ -30,7 +30,8 @@ static const char *s_Anim2DModeNames[] = {
 const FieldParse Anim2DTemplate::s_anim2dFieldParseTable[] = { { "NumberImages", &Parse_Num_Images, nullptr, 0 },
     { "Image", &Parse_Image, nullptr, 0 },
     { "Image_Sequence", &Parse_Image_Sequence, nullptr, 0 },
-    { "AnimationMode", &INI::Parse_Index_List, s_Anim2DModeNames, offsetof(Anim2DTemplate, m_animMode) },
+    // { "AnimationMode", &INI::Parse_Index_List, s_Anim2DModeNames, offsetof(Anim2DTemplate, m_animMode) },
+    FIELD_PARSE_INDEXLIST("AnimationMode", s_Anim2DModeNames, Anim2DTemplate, m_animMode),
```
xezon: game lib will now assert on compilation if `m_animMode` was not an enum and `s_Anim2DModeNames` was not a C string array.

src/game/common/ini/inimacros.h
```diff
@@ -0,0 +1,715 @@
...
+    size_t Size,
+    typename Pred>
+constexpr ExpectedType ReturnWithListCheck(ElemType (&list)[Size], Pred pred)
+{
+#ifndef THYME_USE_STLPORT
+    static_assert(std::is_convertible<decltype(list), ExpectedType>::value,
+        "Wrong type passed to parse function!");
+    static_assert(Size > 0,
+        "List with invalid size passed to parse function!");
+    static_assert(pred()[Size - 1] == nullptr,
```
xezon, line 67: I have not yet found another way to pass constexpr array without this constexpr lambda. Using `list` issues compiler error for some reason.

duncanspumpkin:
Why don't you skip all this macro stuff and just make a templated constructor that checks the types. Would mean you don't need to change every file.

xezon:
Hmm, what checks would that be able to achieve? It then would not know what type is to be expected, and will just take whatever type is passed, even if incorrect.

duncanspumpkin:
> Hmm, what checks would that be able to achieve? It then would not know what type is to be expected, and will just take whatever type is passed, even if incorrect.

You would change all the ini pass functions to actually specify type. Then you use the templated constructor to check the types. The issue with this macro pr is you need to go out your way to add the type checking and if you are doing that you probably already know what you are doing so there isn't any benefit.

duncanspumpkin:
I made a short example of what i mean. It would need further work to fully align with thyme but i think it would work. http://godbolt.org/z/Y8o53xhsb 
It will compiler error if you give it an offset for different type to what the function wants.

xezon:
As far as I am aware there is no syntactic way to call a templated constructor with template type, such that `FieldParse<int32>(arg1, arg2, ...)` would call the int32 templated one. Templated type on constructor needs to be deduced from passed argument. I could be wrong.

An alternative could be to somehow make a
```c++
template <ExpectedType>
constexpr FieldParse Create_FieldParse(arg1, arg2, ...)
```
helper function to create FieldParse with. However, passing this class member offset conveniently could be a problem. With the macros that is easily done as illustrated in this change. But if we can find good solution for it I would also prefer pure constexpr functions over macros.

PR #622, opened 2022-02-24
Add googletest for testing & add filesystem test suite

feliwir:
- [x] Add option for enabling testing (via GoogleTest and CTest)
- [x] Integrate into GitHub workflows
- [x] Add different kind of tests to show usage (for filesystem & bigarchives)


src/w3d/lib/cpudetect.cpp
```diff
@@ -273,6 +273,11 @@ void Get_OS_Info(OSInfoStruct &os_info,
             os_info.VersionString = "UNKNOWN";
             break;
     }
+#else
+    memset(&os_info, 0, sizeof(os_info));
```
duncanspumpkin: This is c++ can't you just do `os_info = OSInfoStruct {};`

feliwir: I copied this from 5 lines above. However i can change it in both places

duncanspumpkin: Oh yes so you did. just leave as is.

feliwir: I still fixed this in my last commit

duncanspumpkin:
Shouldn't these tests be purely for standalone and therefore it can be expected that std::filesystem will be available.

feliwir:
> Shouldn't these tests be purely for standalone and therefore it can be expected that std::filesystem will be available.

Right now these tests are a seperate option and work for both standalone & not standalone - that's why i check for the option

duncanspumpkin:
> > Shouldn't these tests be purely for standalone and therefore it can be expected that std::filesystem will be available.
> 
> Right now these tests are a seperate option and work for both standalone & not standalone - that's why i check for the option

I just fear you are going to need lots of #ifdef.

feliwir:
Well the ifdef for BUILD_WITH_STDFS is required in all cases - it can always be disabled, no matter if standalone or not

duncanspumpkin (approved):
All looks sensible to me but since its new infrastructure @OmniBlade should have final say.

OmniBlade (changes requested):
Just a few files need newline endings, other than that I'm happy to merge it.

feliwir:
> Just a few files need newline endings, other than that I'm happy to merge it.

Done

OmniBlade (approved):

PR #623, opened 2022-02-24
Simplify minor code around ThingTemplate class

xezon:
My review was too late on the last submission from @jonwil, so I put changes here.

### DIFF
--- a/src/game/common/rts/productionprerequisite.cpp
+++ b/src/game/common/rts/productionprerequisite.cpp
@@ -26,24 +26,27 @@ void ProductionPrerequisite::Add_Unit_Prereq(Utf8String name, bool or_with_previ
 
 void ProductionPrerequisite::Resolve_Names()
 {
-    for (unsigned int i = 0; i < m_prereqUnit.size(); i++) {
-        if (m_prereqUnit[i].name.Is_Not_Empty()) {
-            m_prereqUnit[i].unit = g_theThingFactory->Find_Template(m_prereqUnit[i].name, true);
-            captainslog_dbgassert(m_prereqUnit[i].unit != nullptr, "could not find prereq %s", m_prereqUnit[i].name.Str());
+    for (PrereqUnitRec &prereqUnit : m_prereqUnit) {
+        if (prereqUnit.name.Is_Not_Empty()) {
+            prereqUnit.unit = g_theThingFactory->Find_Template(prereqUnit.name, true);
+            captainslog_dbgassert(prereqUnit.unit != nullptr, "could not find prereq %s", prereqUnit.name.Str());
         }
 
-        m_prereqUnit[i].name.Clear();
+        prereqUnit.name.Clear();
     }
 }
 
 int ProductionPrerequisite::Get_All_Possible_Build_Facility_Templates(ThingTemplate **tmpls, int max_tmpls)
 {
-    int count = 0;
-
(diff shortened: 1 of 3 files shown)


src/game/common/rts/productionprerequisite.cpp
```diff
@@ -26,24 +26,25 @@ void ProductionPrerequisite::Add_Unit_Prereq(Utf8String name, bool or_with_previ
 
 void ProductionPrerequisite::Resolve_Names()
 {
-    for (unsigned int i = 0; i < m_prereqUnit.size(); i++) {
```
xezon, line 29: No need to call `m_prereqUnit.size()` on each iteration.

src/game/common/rts/productionprerequisite.cpp
```diff
@@ -26,24 +26,25 @@ void ProductionPrerequisite::Add_Unit_Prereq(Utf8String name, bool or_with_previ
...
     }
 }
 
 int ProductionPrerequisite::Get_All_Possible_Build_Facility_Templates(ThingTemplate **tmpls, int max_tmpls)
 {
-    int count = 0;
+    int i = 0;
+    const int size = static_cast<int>(m_prereqUnit.size());
 
-    for (unsigned int i = 0; i < m_prereqUnit.size() && (i <= 0 || (m_prereqUnit[i].flags & 1) != 0) && count < max_tmpls;
```
xezon, line 43: This function had 2 counters doing the same thing. Reduced to one counter and caching the size to simplify algorithm.

duncanspumpkin: This is still not great. Is there no way to simplify this into something else.

src/game/common/thing/thingtemplate.cpp
```diff
@@ -493,13 +493,13 @@ void ThingTemplate::Set_Copied_From_Default()
...
 void ThingTemplate::Resolve_Names()
 {
-    for (unsigned int i = 0; i < m_prerequisites.size(); i++) {
-        m_prerequisites[i].Resolve_Names();
+    for (ProductionPrerequisite &rerequisite : m_prerequisites) {
+        rerequisite.Resolve_Names();
     }
 
-    for (unsigned int i = 0; i < m_prerequisites.size(); i++) {
+    for (ProductionPrerequisite &rerequisite : m_prerequisites) {
```
duncanspumpkin: this goes for the other loop in this function as well `**p**rerequisite`

src/game/common/rts/productionprerequisite.cpp
duncanspumpkin: I don't think you can change it like this count and i wont align at all times.

duncanspumpkin: What about something like. I think this follows the original code:
```cpp
auto i = 0;
for (auto& prereq_unit : m_prereqUnit){
    if (i > 0 && ((prereq_unit.flags & 1) == 0)){
        break;
    }
    if (i > max_tmpls){
        break;
    }
    tmpls[i++] = prereq_unit.unit;
}
return i;
```

xezon: It also has 2 iterators but ok. I like the splitting of the long condition. `i > max_tmpls` should be `i >= max_tmpls` I think.

src/game/common/thing/thingtemplate.cpp
xezon: Ok.

duncanspumpkin (approved):

PR #633, opened 2022-02-26
Fix potential nullptr access in SurfaceClass::Is_Transparent_Column

xezon:


src/w3d/renderer/surfaceclass.cpp
```diff
@@ -489,26 +489,25 @@ bool SurfaceClass::Is_Transparent_Column(unsigned column)
...
-    while (!(mask & (*pos >> (8 - alpha)))) {
-        ++row;
-        pos += lock_rect.Pitch;
-
-        if (row >= desc.height) {
-            Unlock();
-
-            return true;
+    // #BUGFIX Test bits
+    if (lock_rect.pBits) {
```
duncanspumpkin: Is this actually necessary? 
The d3d api doesn't say that you have to check the validity of the pointer only that you should ensure that the function was successful. If anything should be changed its L486 making that not a warning but a full-blown assert. Also if we are going to check pBits shouldn't you also check Pitch isn't a 0 pitch.

I don't know the full makeup of the d3d api but since we pass unitialised memory to the function would it actually clear pBits to a nullptr on failure?

Perhaps it does make sense but I just don't see it ever getting hit.
Eitherway use != nullptr for a nullptr check and the rest of the function fails fast  so suggest following that convention.

xezon: Check if you like the new code better. Now the static analyzer will complain about this code again however. Either it is too stupid for this or we must check for nullptr.

duncanspumpkin: Ah that's a point didn't think it would make the analyser unhappy.

duncanspumpkin (approved):

PR #634, opened 2022-02-26
Add missing breaks in several switch cases

xezon:
It is not clear to me if some of these fallthroughs are intentional. There are no markings and most look odd.

### DIFF
--- a/src/platform/audio/milesaudiomanager.cpp
+++ b/src/platform/audio/milesaudiomanager.cpp
@@ -1985,7 +1985,7 @@ void MilesAudioManager::Adjust_Playing_Volume(PlayingAudio *audio)
             }
 
             AIL_set_stream_volume_pan(audio->miles.stream, vol, pan);
-        }
+        } break;
         default:
             break;
     }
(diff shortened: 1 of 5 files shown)


src/platform/audio/milesaudiomanager.cpp
```diff
@@ -1985,7 +1985,7 @@ void MilesAudioManager::Adjust_Playing_Volume(PlayingAudio *audio)
             }
 
             AIL_set_stream_volume_pan(audio->miles.stream, vol, pan);
-        }
+        } break;
         default:
```
duncanspumpkin, line 1989: Is this the convention break after the scope? I've worked on different codebases and everyone seems to have a different idea.

xezon: I am not aware about convention. Either way is fine.

duncanspumpkin: Coding style says this is the convention.

duncanspumpkin (approved):

PR #635, opened 2022-02-26
Compare HRESULT against D3D_OK

xezon:


jonwil (approved):
Looks good to me, approved.

PR #640, opened 2022-02-26
Initialize all pointers in WriteLockClass classes

xezon:


jonwil (approved):
Looks good to me, approved.

PR #641, opened 2022-02-26
Avoid invalid indexing in GameLODManager::Get_Recommended_Texture_Red…

xezon:
…uction

### DIFF
--- a/src/game/common/gamelod.cpp
+++ b/src/game/common/gamelod.cpp
@@ -418,7 +418,10 @@ int GameLODManager::Get_Recommended_Texture_Reduction()
     }
 
     if (m_didMemPass) {
-        return m_staticLOD[m_idealStaticGameDetail].texture_reduction_factor;
+        // #BUGFIX Test to avoid invalid indexing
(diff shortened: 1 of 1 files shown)


tomsons26 (approved):

src/game/common/gamelod.cpp
```diff
@@ -418,7 +418,10 @@ int GameLODManager::Get_Recommended_Texture_Reduction()
     }
 
     if (m_didMemPass) {
-        return m_staticLOD[m_idealStaticGameDetail].texture_reduction_factor;
+        // #BUGFIX Test to avoid invalid indexing
+        if (m_idealStaticGameDetail > STATLOD_INVALID && m_idealStaticGameDetail < STATLOD_COUNT) {
```
duncanspumpkin, line 422: shouldn't this be `>= 0`. `STATLOD_INVALID` is currently -1 but if it ever became -2 or a different value then this check would stop being a bounds check.

xezon: Hmm good point. But its already committed.

PR #642, opened 2022-02-26
Initialize additional class members in audio code

xezon:


jonwil (commented):
m_objectID is part of a union. Not sure what the C++ standards say (or what the compilers actually do) about the default initialization of a union and whether you need to explicitly initialize it or not.

duncanspumpkin:
Its the same as other variables in that its ub if you read it before initialising so this all looks correct.

If a member of a union has a user defined constructor then you have to initialise it to one of the types by providing a constructor. But thats not how these are being used here and even if it was this all looks correct.

jonwil (approved):
Looks good to me, approved.

PR #645, opened 2022-02-26
Fix bug in Drawable::Set_Stealth_Look

xezon:


duncanspumpkin:
Is there a related issue that this fixes?

xezon:
Well. Static analyzer warns about this rightfully. If you look at function:

```c++
StealthUpdate *update = object->Get_Stealth_Update();

if (update == nullptr) {
...
}

if (update->Has_Disguised_Template()) {
...
}
```
Note how in second condition it accesses update pointer without check. This could cause crash.

Original code also appears to look different than what Thyme does.

jonwil (approved):
Looks ok to me.

PR #647, opened 2022-02-26
Initialize additional class members in w3dengine code

xezon:


jonwil (approved):
Looks good to me, approved.

PR #650, opened 2022-02-26
Initialize additional class members in common game code

xezon:


jonwil (approved):
Looks good to me, approved.

PR #659, opened 2022-03-01
Force tests off when building with dll

duncanspumpkin:
This is because thyme.lib is not available when in dll mode.
Also removed incorrect declaration its part of the header


duncanspumpkin:
Not sure if this is the best way to override the option but it works on my machine.

CMakeLists.txt
```diff
@@ -98,6 +98,7 @@ cmake_dependent_option(USE_STDFS "Use C++ 17 filesystem for crossplatform file h
 if(WIN32 OR "${CMAKE_SYSTEM}" MATCHES "Windows")
     if(CMAKE_SIZEOF_VOID_P EQUAL 4)
         option(BUILD_DLL "Build an injectable dll version." ON)
+	set(BUILD_TESTS OFF)
```
xezon, line 101: Use whitespace and not tab for indentation.

xezon: I don't know whether or not this is elegant approach.

OmniBlade:
Closed in favour of #660

PR #661, opened 2022-03-02
Implements a CMakePresets.json file.

OmniBlade:
Now uses preset config for CI builds.
Moves compiler flags out of CMakeLists.txt into presets.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,114 @@
+{
+    "version": 1,
+    "configurePresets": [
+        {
+            "name": "default",
+            "displayName": "Default Config",
+            "generator": "Ninja",
+            "hidden": true,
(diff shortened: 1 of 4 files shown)


CMakePresets.json
```diff
@@ -0,0 +1,114 @@
...
+                "CXX": "g++"
+            }
+        },
+		{
+            "name": "clang",
+			"inherits": "default-unix",
+            "displayName": "Clang build",
+			"environment": {
+                "CC": "clang",
+                "CXX": "clang++"
```
xezon, line 110: This file has lines with leading tabs and leading spaces.

OmniBlade: Copy/paste artifacts, I'll clean it up.

xezon (approved):

PR #667, opened 2022-03-05
Print assert message on assert log

xezon:
Before:
```
FATAL D:\Projects\TheAssemblyArmada\Thyme\src\game\client\draw\w3dtruckdraw.cpp:531: ASSERTION FAILED!
  Function:void __thiscall W3DTruckDraw::Update_Bones(void)
  Expression:m_frontLeftTireBone != 0
```

After
```
FATAL D:\Projects\TheAssemblyArmada\Thyme\src\game\client\draw\w3dtruckdraw.cpp:531: ASSERTION FAILED!
  Function:void __thiscall W3DTruckDraw::Update_Bones(void)
  Expression:m_frontLeftTireBone != 0
  Message:Missing front-left tire bone Tire01 in model UVBATTBUS_D1
```

### DIFF
--- a/deps/captnlog/src/captainslog.h
+++ b/deps/captnlog/src/captainslog.h
@@ -68,8 +68,12 @@ extern "C" {
 
 #if LOGLEVEL_FATAL <= LOGGING_LEVEL
 #define captainslog_fatal(x, ...) captainslog_log(LOGLEVEL_FATAL, __FILE__, __LINE__, x, ##__VA_ARGS__)
+#define captainslog_fatal_alloc_varg_str(x, ...) captainslog_alloc_varg_str(LOGLEVEL_FATAL, x, ##__VA_ARGS__)
+#define captainslog_fatal_free_varg_str(x) captainslog_free_varg_str(x)
 #else
 #define captainslog_fatal(x, ...) ((void)0)
+#define captainslog_fatal_alloc_varg_str(x, ...) ((char *)NULL)
+#define captainslog_fatal_free_varg_str(x, ...) ((void)0)
 #endif
 
 /**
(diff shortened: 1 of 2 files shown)


deps/captnlog/src/captainslog.h
```diff
@@ -100,11 +104,15 @@ typedef struct captains_settings_s
 #define captainslog_log(level, file, line, format, ...) ((void)0)
 #define captainslog_line(fmt, ...) ((void)0)
 #define captainslog_deinit() ((void)0)
+#define captainslog_vargs_str(level, format, ...) NULL
```
xezon: This looks wrong.

deps/captnlog/src/captnlog.c
```diff
@@ -123,6 +120,9 @@ void captainslog_log(int level, const char *file, int line, const char *fmt, ...
         file_str[0] = '\0';
     }
 
+    /* Lock the mutex to prevent multiple threads trying to write at the same time */
+    lock();
```
xezon, line 124: I moved this down because there is no need to lock already when writing to the stack buffers.

deps/captnlog/src/captainslog.h
xezon: Fixed.

deps/captnlog/src/captnlog.c
```diff
@@ -214,3 +214,29 @@ void captainslog_deinit()
...
+        vsnprintf(str, 512, fmt, args);
+        va_end(args);
+    }
+    return str;
+}
+
+void captainslog_free_varg_str(char *str)
+{
+    if (str != NULL) {
+        free(str);
```
OmniBlade: Standard says free should be fine to call on a NULL pointer so the null check is redundant.

OmniBlade (changes requested):
Looks okay to me, just the possible redundant free check.

deps/captnlog/src/captnlog.c
xezon: According to this stackoverflow there are some library implementations where it will incur crash. See
https://stackoverflow.com/questions/1938735/does-freeptr-where-ptr-is-null-corrupt-memory

But I can remove check, no problem.

xezon: Fixed.

OmniBlade (approved):

PR #669, opened 2022-03-05
Remove log spam and reinstate disabled asserts as error log messages

xezon:
This change gets rid of log spam and reinstate disabled asserts as error log messages.


jonwil (approved):
Looks good to me.

PR #673, opened 2022-03-08
Add AudioDataHandle alias and fix issue in MilesAudioFileCache::Release_Open_Audio

xezon:


src/game/common/audio/audiomanager.h
```diff
@@ -56,6 +56,8 @@ enum CachedVarsType
     CACHED_UNK8 = 1 << 8, // Speech related.
 };
 
+using AudioDataHandle = void *;
```
xezon, line 59: `AudioDataHandle` helps understanding what this `void *` is for.

src/platform/audio/milesaudiofilecache.cpp
```diff
@@ -227,12 +227,12 @@ bool MilesAudioFileCache::Free_Space_For_Sample(const OpenAudioFile &file)
 void MilesAudioFileCache::Release_Open_Audio(OpenAudioFile *file)
 {
     // Close any playing samples that use this data.
-    if (file->ref_count) {
-        g_theAudio->Close_Any_Sample_Using_File(file);
+    if (file->ref_count != 0) {
+        g_theAudio->Close_Any_Sample_Using_File(file->wave_data);
```
xezon, line 231: Thyme bugfix here. I put breakpoint to see if it is called in game, but did not hit, so maybe was not major problem on runtime.

feliwir: Shouldn't you add the `// #BUGFIX` thing here?

xezon: No, because this is what the original does. It was just wrong in Thyme.

feliwir: Okay, good!

src/platform/audio/milesaudiomanager.cpp
```diff
@@ -926,7 +926,7 @@ bool MilesAudioManager::Has_3D_Sensitive_Streams_Playing()
  *
  * 0x00780820
  */
-void *MilesAudioManager::Get_Bink_Handle()
+AILLPDIRECTSOUND MilesAudioManager::Get_Bink_Handle()
```
OmniBlade: Should't this be BinkHandle now?

OmniBlade (changes requested):

src/platform/audio/milesaudiomanager.cpp
xezon: Yes. Fixed.

OmniBlade (approved):

PR #676, opened 2022-03-09
Add basic math tests

feliwir:
- [x] Basic float vector tests
- [x] Basic integer vector tests
- [x] Basic matrix tests


tests/test_w3d_math.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include <vector3.h>
+#include <vector4.h>
+
+#include <vector2i.h>
+#include <vector3i.h>
+
+#include <matrix3.h>
+#include <matrix4.h>
+
+template<class V> inline void TestLength(const V &v)
```
xezon: Why did you choose to create template functions shared among different test cases? On first sight this unnecessary as for the flow of the test it would be easier to look at everything in the test function step by step.

tests/test_w3d_math.cpp
```diff
@@ -0,0 +1,189 @@
...
+    EXPECT_FLOAT_EQ(a.Dot_Product(a, a), 4.0f);
+    EXPECT_FLOAT_EQ(a * a, 4.0f);
+
+    EXPECT_FLOAT_EQ(a.Dot_Product(a, b), 0.0f);
+    EXPECT_FLOAT_EQ(a * b, 0.0f);
+}
+
+TEST(w3d_math, vector2)
+{
+    Vector2 a(2.0f, 0.0f);
```
xezon: Maybe you could use a vector which has a non-x value as well, otherwise this test is X biased.

tests/test_w3d_math.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+    Vector2 b(0.0f, 2.0f);
+
+    // Test dot product
+    TestDotProduct(a, b);
+
+    // Test distance
+    TestDistance(a, b);
+
+    // Test equality operator
```
xezon, line 51: Maybe all these comments should be separate tests then? It seems odd on first sight that the test does many things but we have comments indicating sub-tests?

For example you could have a signle test for vector3 dot product, vector3 cross product, vector3 translation, ... This way if a test fails you know exactly which functionality broke. With big function like this testing "whole class", you don't know exactly what broke eventually.

tests/test_w3d_math.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+    // Test normalize
+    auto tmp = a;
+    tmp.Normalize();
+    EXPECT_FLOAT_EQ(tmp.Length(), 1.0f);
+
+    Vector2 b(0.0f, 2.0f);
+
+    // Two notations for dot product
+    EXPECT_FLOAT_EQ(a.Dot_Product(a, a), 5.0f);
```
xezon: `Dot_Product` is a static function no? I think it would be better to call as `Vector2::Dot_Product`.

feliwir: Right - fixed

xezon (approved):

tests/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ FetchContent_Declare(
 
 FetchContent_MakeAvailable(googletest)
 
-add_executable(thyme_tests globals.cpp test_crc.cpp test_filesystem.cpp)
+add_executable(thyme_tests globals.cpp test_crc.cpp test_filesystem.cpp test_w3d_math.cpp)
```
duncanspumpkin: I feel you should break this out into a `set(TESTS_SRC)` maybe not in this PR but soon.

src/w3d/math/matrix4.h
```diff
@@ -476,3 +476,13 @@ __forceinline Vector4 operator*(const Matrix4 &a, const Vector4 &v)
         a[2][0] * v[0] + a[2][1] * v[1] + a[2][2] * v[2] + a[2][3] * v[3],
         a[3][0] * v[0] + a[3][1] * v[1] + a[3][2] * v[2] + a[3][3] * v[3]);
 }
+
+__forceinline int operator==(const Matrix4 &a, const Matrix4 &b)
```
duncanspumpkin, line 480: I know this is following the convention of the above but `a` and `b` are rubbish names `lhs` and `rhs` would be better.

feliwir: I'd like to keep this consistent with the other function signatures

tests/CMakeLists.txt
feliwir: Yes, i can do this in another PR

feliwir:
After rebasing this again, I'd appreciate this getting reviewed / merged. Last time i already addressed all review comments

src/w3d/math/matrix4.h
```diff
@@ -493,4 +503,4 @@ void To_Matrix4(Matrix4 &m, const _D3DMATRIX &dxm);
 [[nodiscard]] Matrix4 To_Matrix4(const _D3DMATRIX &dxm);
 
 } // namespace Thyme
-#endif
+#endif
```
xezon: Can add EOL here.

tests/test_w3d_math.cpp
```diff
@@ -0,0 +1,183 @@
...
+    EXPECT_FLOAT_EQ(a.Length2(), 7.0f);
+
+    // Test normalize
+    auto tmp = a;
+    tmp.Normalize();
+    EXPECT_FLOAT_EQ(tmp.Length(), 1.0f);
+
+    Vector4 b(0.0f, 2.0f, 0.0f, 0.0f);
+
+    // Two notations for dot product
```
xezon, line 113: You can make smaller test for all these individual comments. Then it will be easier to see from the test report what operation exactly failed.

As of right now, if dot product fails but all others pass, then one would not immediately know that dot product was causing the fail.

xezon: I see I proposed the same in March. Hmm. Why was this closed?

feliwir: That would work, but that kind of destroys the purpose of `EXPECT_` macros. I didn't see the comment in march

PR #681, opened 2022-03-10
Use range based for-loop where applicable

feliwir:
- [x] add `.clang-tidy` file
- [x] Use range-based loops when we can:
  - Remove a magic number
  - Improve readability

### DIFF
--- a/src/game/common/modules/modulefactory.cpp
+++ b/src/game/common/modules/modulefactory.cpp
@@ -48,10 +48,10 @@ void ModuleFactory::Init()
  */
 void ModuleFactory::CRC_Snapshot(Xfer *xfer)
 {
-    for (auto it = m_moduleDataList.begin(); it != m_moduleDataList.end(); ++it) {
+    for (auto &it : m_moduleDataList) {
         // TODO Snapshot methods are not const, so need to cast here.
         // Really the ModuleData vector should be none const I guess, but symbols suggest it was const.
-        const_cast<ModuleData *>(*it)->CRC_Snapshot(xfer);
+        const_cast<ModuleData *>(it)->CRC_Snapshot(xfer);
     }
 }
 
@@ -62,9 +62,9 @@ void ModuleFactory::CRC_Snapshot(Xfer *xfer)
  */
 void ModuleFactory::Xfer_Snapshot(Xfer *xfer)
 {
-    for (auto it = m_moduleDataList.begin(); it != m_moduleDataList.end(); ++it) {
+    for (auto &it : m_moduleDataList) {
         // TODO Snapshot methods are not const, so need to cast here.
-        const_cast<ModuleData *>(*it)->Xfer_Snapshot(xfer);
+        const_cast<ModuleData *>(it)->Xfer_Snapshot(xfer);
     }
 }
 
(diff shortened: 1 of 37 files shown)


feliwir:
Can't do this with STLPort - wait for standalone

duncanspumpkin:
Should work after a rebase now.

src/game/common/modules/modulefactory.cpp
```diff
@@ -48,10 +48,10 @@ void ModuleFactory::Init()
  */
 void ModuleFactory::CRC_Snapshot(Xfer *xfer)
 {
-    for (auto it = m_moduleDataList.begin(); it != m_moduleDataList.end(); ++it) {
+    for (auto &it : m_moduleDataList) {
```
xezon, line 51: Should this maybe have a better name than `it` ? For example `moduleData` ?

xezon, line 51: Should this maybe have a better name than `it` ? For example `moduleData` ?

src/game/common/modules/modulefactory.cpp
```diff
@@ -62,9 +62,9 @@ void ModuleFactory::CRC_Snapshot(Xfer *xfer)
  */
 void ModuleFactory::Xfer_Snapshot(Xfer *xfer)
 {
-    for (auto it = m_moduleDataList.begin(); it != m_moduleDataList.end(); ++it) {
+    for (auto &it : m_moduleDataList) {
```
xezon, line 65: Should this maybe have a better name than `it` ? For example `moduleData` ?

xezon, line 65: Should this maybe have a better name than `it` ? For example `moduleData` ?

src/game/common/rts/productionprerequisite.cpp
```diff
@@ -149,8 +149,8 @@ Utf16String ProductionPrerequisite::Get_Requires_List(const Player *player) cons
 
         bool has_sciences = true;
 
-        for (size_t i = 0; i < m_prereqScience.size(); i++) {
-            if (!player->Has_Science(m_prereqScience[i])) {
+        for (auto reqScience : m_prereqScience) {
```
xezon, line 152: Maybe just call it `science`

xezon, line 152: Maybe just call it `science`

src/game/common/userpreferences.cpp
```diff
@@ -61,8 +61,8 @@ bool UserPreferences::Write()
     FILE *fp = fopen(m_filename.Str(), "w");
 
     if (fp != nullptr) {
-        for (auto it = begin(); it != end(); ++it) {
-            fprintf(fp, "%s = %s\n", it->first.Str(), it->second.Str());
+        for (auto &it : *this) {
```
xezon, line 64: What is `it` ? Can this have a better name? If not please don't use `auto` type. It is very difficult to read from review what is going on here.

xezon, line 64: What is `it` ? Can this have a better name? If not please don't use `auto` type. It is very difficult to read from review what is going on here.

feliwir:
I'll revisit this at a later stage

PR #682, opened 2022-03-10
Simplify Array Hooking

duncanspumpkin:
This should have no noticeable changes and just means that the `ArrayHelper` can be removed. I also fixed g_waterSettings to make it match all the other arrays.

### DIFF
--- a/src/hooker/setupglobals_zh.cpp
+++ b/src/hooker/setupglobals_zh.cpp
@@ -55,7 +55,7 @@ uint32_t &View::s_idNext = Make_Global<uint32_t>(PICK_ADDRESS(0x009D4178, 0));
 
 // water.cpp
 #include "water.h"
-WaterSetting *g_waterSettings = Make_Pointer<WaterSetting>(PICK_ADDRESS(0x00A2F0B8, 0x00A2F0B8));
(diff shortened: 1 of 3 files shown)


src/hooker/setupglobals_zh.cpp
```diff
@@ -55,7 +55,7 @@ uint32_t &View::s_idNext = Make_Global<uint32_t>(PICK_ADDRESS(0x009D4178, 0));
 
 // water.cpp
 #include "water.h"
-WaterSetting *g_waterSettings = Make_Pointer<WaterSetting>(PICK_ADDRESS(0x00A2F0B8, 0x00A2F0B8));
+ARRAY_DEF(PICK_ADDRESS(0x00A2F0B8, 0x00A2F0B8), WaterSetting, g_waterSettings, 5);
```
xezon: Can use `TIME_OF_DAY_COUNT` here too?

Btw, why are these Macros necessary? Can this not be typed out simply without the macros or would that look too complicated?

duncanspumpkin: The macros are not necessary and i can remove them I initially didn't do that as I wanted to ensure it would work with minimal changes.

xezon (approved):

PR #683, opened 2022-03-11
Remove ArrayDef macro and cleanup of Make_Global

duncanspumpkin:
I used a regex to find/replace the ArrayDefs with the equivalent Make_Global code.
I also noticed that two users of Make_Global were from a very old version where a macro was used for the identifier name. I've brought those into normal alignment.
One item of note whilst doing this is that we have a handful of Make_Globals outside of the setupglobals file.
Unsure if they should all be moved into setupglobals.

### DIFF
--- a/src/w3d/lib/systimer.h
+++ b/src/w3d/lib/systimer.h
@@ -54,10 +54,7 @@ inline SysTimeClass::~SysTimeClass()
 }
 
 #ifdef GAME_DLL
-#include "hooker.h"
-
-// This needs to stay here like this to avoid static init issues in cpudetect.
-#define g_theSysTimer (Make_Global<SysTimeClass>(0x00A66B30))
+extern SysTimeClass &g_theSysTimer;
 #else
 extern SysTimeClass g_theSysTimer;
 #endif
--- a/src/game/client/shadermanager.h
+++ b/src/game/client/shadermanager.h
@@ -111,7 +111,7 @@ private:
     static w3dsurface_t s_newRenderSurface;
     static w3dsurface_t s_oldDepthSurface;
 #ifdef GAME_DLL
-    static ARRAY_DEC(TextureClass *, s_textures, MAX_TEXTURE_STAGES);
+    static TextureClass *(&s_textures)[MAX_TEXTURE_STAGES];
     static bool &s_renderingToTexture;
     static ShaderTypes &s_currentShader;
 #else
--- a/src/game/client/water.h
+++ b/src/game/client/water.h
@@ -79,7 +79,7 @@ private:
 };
 
 #ifdef GAME_DLL
-extern ARRAY_DEC(WaterSetting, g_waterSettings, TIME_OF_DAY_COUNT);
+extern WaterSetting (&g_waterSettings)[TIME_OF_DAY_COUNT];
 extern Override<WaterTransparencySetting> &g_theWaterTransparency;
 #else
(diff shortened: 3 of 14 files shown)


src/w3d/lib/systimer.h
```diff
@@ -54,10 +54,7 @@ inline SysTimeClass::~SysTimeClass()
 }
 
 #ifdef GAME_DLL
-#include "hooker.h"
-
-// This needs to stay here like this to avoid static init issues in cpudetect.
```
xezon, line 59: Not sure what the comment tries to tell us. Did you check that this works correctly?

xezon (commented):
Hmm. To be fair it actually looks nicer to look at with the macros. The duplication of types and array sizes is quite a strain when typing it out by hand. Maybe we can just leave it as is.

src/w3d/lib/systimer.h
duncanspumpkin: It's a generic comment from years back as to why it is using a make global. At least that's my understanding of it.

duncanspumpkin:
> Hmm. To be fair it actually looks nicer to look at with the macros. The duplication of types and array sizes is quite a strain when typing it out by hand. Maybe we can just leave it as is.

For the make globals you can get rid of the left hand side and replace with `auto&` to reduce the duplication. I'm not overly fussed either way it is done this doesn't exactly represent much work

xezon:
Yes `auto` could be ok here, because the type is already written on the right side.

```c++
auto &g_waterSettings = Make_Global<WaterSetting[TIME_OF_DAY_COUNT]>(PICK_ADDRESS(0x00A2F0B8, 0x00A2F0B8));
```

Will this compile?

duncanspumpkin:
> Yes `auto` could be ok here, because the type is already written on the right side.
> 
> ```c++
> auto &g_waterSettings = Make_Global<WaterSetting[TIME_OF_DAY_COUNT]>(PICK_ADDRESS(0x00A2F0B8, 0x00A2F0B8));
> ```
> 
> Will this compile?

Yes that should compile.

duncanspumpkin:
After further testing I can confirm it would compile but only if we didn't forward declare so basically we can't do it that way.

src/w3d/lib/systimer.h
OmniBlade: For reference, this is before I added `#pragma init_seg(lib)` to the setupglobals_zh.cpp file to force the globals to be initialised before other code. Before I did this, the CPUDetect code that runs in the pre-main init code would fail on a call to the timer because the static wasn't initialised in time which isn't an issue when using a standalone build as it wouldn't be a pointer.

xezon (approved):

OmniBlade:
@duncanspumpkin Looks like CI checks are broken, once its passing the CI I'll merge this.

duncanspumpkin:
Even with how much of a mouthful the definition ends up being?

OmniBlade:
We could keep definitions macros just to keep some of the duplicate type and size declarations to a minimum? At least we don't need the helper structs anymore.

duncanspumpkin:
> We could keep definitions macros just to keep some of the duplicate type and size declarations to a minimum? At least we don't need the helper structs anymore.

The helper struct was removed in the other PR this was just removing the macros. Tbh I'm not sure if there is much merrit in this PR it ends up being so verbose its annoying that auto can't come to our rescue for this.

PR #699, opened 2022-03-22
Refactors dll injection system to proxy via avifil32.dll.

OmniBlade:
Builds a small stub dll that proxies avifil32.dll and then loads
thyme.dll and runs its Setup_Hooks function. Can in theory support
multiple binaries with different dlls to inject for each.


src/hookproxy/proxydll.cpp
```diff
@@ -0,0 +1,251 @@
...
+ * Get the payload dll name we expect for the binary based on its hash.
+ */
+const char* Get_Payload_Name()
+{
+    //Exe hashes we support and the payload dll name we expect for them.
+    static const struct ExePayloads
+    {
+        const char *exe_hash;
+        const char *payload_name;
+    } Payloads[] = {
```
xezon: Declaration and instantiation of struct can be split up for better readability.

src/hookproxy/proxydll.cpp
```diff
@@ -0,0 +1,251 @@
...
+    return nullptr;
+}
+
+/**
+ * Loads the dll we are proxying for and the functions the target application imports.
+ */
+void Load_Forwarded_Functions()
+{
+    char path[MAX_PATH];
+    CopyMemory(path + GetSystemDirectoryA(path, MAX_PATH - 14), "\\avifil32.dll", 14);
```
xezon: Any particular reason to use `CopyMemory` instead of `memcpy` ? Also, can unhardcode value 14 here? Can use `const char dll_name[] = "\\avifil32.dll";` and `sizeof(dll_name)`

src/hookproxy/proxydll.cpp
```diff
@@ -0,0 +1,251 @@
...
+    HMODULE dll = LoadLibraryA(path);
```
xezon, line 143: There is a `FreeLibraryA` pairing missing for this?

src/hookproxy/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+set_target_properties(proxydll PROPERTIES OUTPUT_NAME avifil32 PDB_NAME proxydll)
+target_exports(proxydll SYMBOLS 
+    AVIStreamWrite=FakeAVIStreamWrite
+    AVIFileInit=FakeAVIFileInit
+    AVIFileOpenA=FakeAVIFileOpenA
+    AVIFileRelease=FakeAVIFileRelease
+    AVIFileExit=FakeAVIFileExit
+    AVIFileCreateStreamA=FakeAVIFileCreateStreamA
+    AVIStreamSetFormat=FakeAVIStreamSetFormat
+    AVIStreamRelease=FakeAVIStreamRelease
```
xezon, line 15: Who will call these exports?

src/hookproxy/proxydll.cpp
```diff
@@ -0,0 +1,251 @@
...
+        DWORD size;
+
+        size = GetModuleFileNameW(nullptr, buffer, sizeof(buffer) / sizeof(buffer[0]));
+
+        if (size == 0) {
+            break;
+        } else if (size == (DWORD)(sizeof(buffer) / sizeof(buffer[0]))) {
+            DWORD size_ = size;
+            do {
+                wchar_t *path_ = (wchar_t *)std::realloc(path, sizeof(wchar_t) * size_ * 2);
```
xezon: Instead of reallocating here I suggest to pass sufficient writable buffer as function argument. This way buffer responsibility is handled more elegantly purely by caller. Plus caller can use stack buffer of MAX_PATH. This function overall appears to be quite overengineered. I suggest to remove all unnecessary code and then take a fresh look at it.

src/hookproxy/proxydll.cpp
```diff
@@ -0,0 +1,251 @@
...
+wchar_t *Get_Path()
+{
+    wchar_t buffer[MAX_PATH];
+    wchar_t *path = nullptr;
+    wchar_t *full_path = nullptr;
+
+    while (true) {
+        DWORD size;
+
+        size = GetModuleFileNameW(nullptr, buffer, sizeof(buffer) / sizeof(buffer[0]));
```
xezon: `ARRAY_SIZE(buffer)`

src/hookproxy/proxydll.cpp
```diff
@@ -0,0 +1,251 @@
...
+bool Load_Setup_Hooks()
+{
+    const char *payload = Get_Payload_Name();
+
+    if (payload == nullptr) {
+        MessageBoxA(0, "Cannot find a payload for this binary", "Proxy", MB_ICONERROR);
+        return false;
+    }
+
+    HMODULE dll = LoadLibraryA(payload);
```
xezon, line 173: `FreeLibraryA` pair is missing?

src/hookproxy/proxydll.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+    return full_path;
+}
+
+/**
+ * Get the hash for the exe we were loaded by.
+ */
+const char* Get_Hash()
+{
+    static char hash[41];
```
xezon: `SHA_DIGEST_LENGTH * 2 + 1`

src/hookproxy/proxydll.cpp
```diff
@@ -0,0 +1,251 @@
...
+        const char *exe_hash;
+        const char *payload_name;
+    } Payloads[] = {
+        { "5b805b4ecba79581fe8ec454b5acf7b7e13d4dd5", "thyme.dll" },
+        { nullptr, nullptr },
+    };
+
+    const char *target_hash = Get_Hash();
+
+    for (unsigned i = 0; Payloads[i].exe_hash != nullptr; ++i) {
```
xezon: Loop can just condition with `i < ARRAY_SIZE(Payloads)`. There is no need for a null terminated array here.

src/hookproxy/proxydll.cpp
```diff
@@ -0,0 +1,251 @@
...
+/**
+ * Loads the dll we are proxying for and the functions the target application imports.
+ */
+void Load_Forwarded_Functions()
+{
+    char path[MAX_PATH];
+    CopyMemory(path + GetSystemDirectoryA(path, MAX_PATH - 14), "\\avifil32.dll", 14);
+    HMODULE dll = LoadLibraryA(path);
+    if (dll == nullptr) {
+        MessageBoxA(0, "Cannot load original avifil32.dll library", "Proxy", MB_ICONERROR);
```
xezon: Minor: `NULL`

src/hookproxy/proxydll.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+/**
+ * Loads the dll payload and retrieves the Setup_Hooks function from it.
+ */
+bool Load_Setup_Hooks()
+{
+    const char *payload = Get_Payload_Name();
+
+    if (payload == nullptr) {
+        MessageBoxA(0, "Cannot find a payload for this binary", "Proxy", MB_ICONERROR);
```
xezon: Minor: `NULL`

src/hookproxy/sha1.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+        for (int i = 0; i < 5; ++i) {
+            finalp[i] = htobe32(finalp[i]);
+        }
+
+        m_computed = true;
+        std::memcpy(output, &m_finalDigest, SHA_DIGEST_LENGTH);
+    }
+
+    return 20;
```
xezon: `SHA_DIGEST_LENGTH` ?

src/hookproxy/sha1.cpp
```diff
@@ -0,0 +1,252 @@
...
+    digest.H3 += d;
+    digest.H4 += e;
+}
+
+/**
+ * @brief Formats the hash into a string for printing.
+ */
+void SHAEngine::Print(const void *buffer, char *output)
+{
+    for (int i = 0; i < 20; i++) {
```
xezon: `SHA_DIGEST_LENGTH` ?

src/hookproxy/sha1.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * @brief Prints the SHA1 hash to the provided buffer as a C string.
+ *
+ * @param output Pointer to the destination buffer.
+ * @return Length of the string in bytes.
+ */
+int SHAEngine::Print_Result(char *output)
+{
+    uint8_t buffer[20];
```
xezon: `SHA_DIGEST_LENGTH` ?

src/hookproxy/sha1.cpp
```diff
@@ -0,0 +1,252 @@
...
+    digest.H4 += e;
+}
+
+/**
+ * @brief Formats the hash into a string for printing.
+ */
+void SHAEngine::Print(const void *buffer, char *output)
+{
+    for (int i = 0; i < 20; i++) {
+        std::sprintf(&output[2 * i], "%02x", *(static_cast<const uint8_t *>(buffer) + i));
```
xezon: This looks dangerous. What if `output` buffer is not size 2*20? Then we corrupt memory here, no? Any chance to rewrite this to be safe to call always?

xezon:
How does game load this? I assume game calls the AVI functions from random dll? But this would mean that you will inject later than some other code and eventually lose the ability to debug very early?

src/hookproxy/proxydll.cpp
OmniBlade: MAX_PATH doesn't really represent the actual maximum path size on windows anymore so its possible for a path to exceed that length up to 32,767 characters according to microsofts docs. I guess we could just allocate that on the stack or as a static buffer from the get go?

src/hookproxy/proxydll.cpp
OmniBlade: The library needs to remain loaded as long as the proxy library is as its used to fulfill the function jumps and this library will only be unloaded when the program closes so its probably safe to let the OS clean up the loaded libraries.

src/hookproxy/CMakeLists.txt
OmniBlade: game.dat calls them, or at the very least imports them. They are just jumps to the real dll when called.

src/hookproxy/proxydll.cpp
xezon: It's not a deal breaker to not clean up on unload, but it's not very nice either. Next person looking at it may wonder again.

src/hookproxy/proxydll.cpp
```diff
@@ -0,0 +1,251 @@
...
+            path = buffer;
+        }
+
+        full_path = _wfullpath(nullptr, path, MAX_PATH);
+
+        break;
+    }
+
+    if (path != buffer) {
+        free(path);
```
xezon: shouldn't use `std:free` when using `std::realloc` ?

src/hookproxy/proxydll.cpp
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    return full_path;
```
xezon: Is there a chance function can return non-null but `path` was freed above? If so, then user may double free?

src/hookproxy/proxydll.cpp
xezon: Hmm must be a new thing. I think originally Windows API just handles with MAX_PATH. If you want function to allocate and user to free, you maybe make it obvious in file name and also offer a matching free function for it? It's always odd to return something where user needs to take extra care to free something correctly, whether it be free, delete or delete[] or something else.

OmniBlade: On reflection it probably is overkill as who is likely to be running this and also have the game located in a directory that resolves to greater than MAX_PATH?

src/hookproxy/proxydll.cpp
OmniBlade: You aren't supposed to call LoadLibrary from within DLLMain either, this is breaking all kinds of best practice. I will add a note to the detach case in DLLMain that we are deliberately not cleaning up.

src/hookproxy/proxydll.cpp
```diff
@@ -75,96 +76,54 @@ __declspec(naked) void FakeAVIStreamRelease()
  */
 wchar_t *Get_Path()
```
xezon: Should return `const wchar_t*` then.

src/hookproxy/proxydll.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+/**
+ * Loads the dll payload and retrieves the Setup_Hooks function from it.
+ */
+bool Load_Setup_Hooks()
+{
+    const char *payload = Get_Payload_Name();
+
+    if (payload == nullptr) {
+        MessageBoxA(nullptr, "Cannot find a payload for this binary", "Proxy", MB_ICONERROR);
```
xezon: I suspect most users will not understand this message. Maybe can tell something about game.dat?

src/hookproxy/sha1.cpp
```diff
@@ -0,0 +1,246 @@
...
+}
+
+/**
+ * @brief Returns the SHA1 hash as a C string.
+ *
+ * @warning Multiple calls share a single buffer.
+ */
+const char *SHAEngine::Print_Result()
+{
+    static char output[SHA_DIGEST_LENGTH * 2 + 1];
```
xezon: You can make this a class member instead of static.

xezon (approved):

PR #703, opened 2022-03-27
Implements Volume Shadow Geometry class.

tomsons26:
Split off my shadow PR cause its simpler this way.

### DIFF
--- a/src/game/client/shadow/w3dvolumetricshadow.cpp
+++ b/src/game/client/shadow/w3dvolumetricshadow.cpp
@@ -525,3 +525,78 @@ W3DShadowGeometry *W3DShadowGeometryManagerIterator::Get_Current_Geom()
 {
     return static_cast<W3DShadowGeometry *>(Get_Current());
 }
+
+Geometry::Geometry() :
+    m_verts(nullptr),
+    m_indices(nullptr),
+    m_numPolygon(0),
+    m_numVertex(0),
+    m_flags(0),
+    // BUGFIX original didn't init these
+    m_numActivePolygon(0),
+    m_numActiveVertex(0),
+    m_visibleState(STATE_UNKNOWN)
+{
+}
+
+Geometry::~Geometry()
+{
+    Release();
+}
+
+bool Geometry::Create(int num_vert, int num_poly)
+{
+    if (num_vert) {
+        // BUGFIX Geos can be reused so clean up to avoid memory leak
+        if (m_verts != nullptr) {
+            delete[] m_verts;
+        }
+
+        m_verts = new Vector3[num_vert];
+        if (m_verts == nullptr) {
+            return false;
+        }
+    }
+
+    if (num_poly) {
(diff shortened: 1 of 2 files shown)


src/game/client/shadow/w3dvolumetricshadow.cpp
```diff
@@ -525,3 +525,73 @@ W3DShadowGeometry *W3DShadowGeometryManagerIterator::Get_Current_Geom()
...
+    if (num_poly) {
+        // BUGFIX Geos can be reused so clean up to avoid memory leak
+        if (m_indices != nullptr) {
+            delete[] m_indices;
+        }
+
+        m_indices = new short[3 * num_poly];
+
+        if (m_indices == nullptr) {
+            return false;
```
xezon, line 575: Should this not cleanup `m_verts` if it fails here? Despite, this checking looks obsolete, because it will only fail if out of memory and then it will crash somewhere else anyway.

src/game/client/shadow/w3dvolumetricshadow.h
```diff
@@ -153,3 +156,92 @@ class W3DShadowGeometryManagerIterator : public HashTableIteratorClass
...
+    void Set_Vertex(int index, Vector3 *vert) { m_verts[index] = *vert; }
+
+    // not yet found
+    int Find_Vertex_In_Range(int, int, Vector3 *);
+
+    AABoxClass &Get_Bounding_Box() { return m_boundingBox; }
+    void Set_Bounding_Box(AABoxClass &box) { m_boundingBox = box; }
+
+    void Set_Bounding_Sphere(SphereClass &sphere) { m_boundingSphere = sphere; }
+    SphereClass &Get_Bounding_Sphere() { return m_boundingSphere; }
```
xezon, line 225: Move get function above set function for consistency of order.

src/game/client/shadow/w3dvolumetricshadow.h
```diff
@@ -153,3 +156,92 @@ class W3DShadowGeometryManagerIterator : public HashTableIteratorClass
...
+        return 3;
+    }
+
+    Vector3 *Get_Vertex(int index) const { return &m_verts[index]; }
+    void Set_Vertex(int index, Vector3 *vert) { m_verts[index] = *vert; }
+
+    // not yet found
+    int Find_Vertex_In_Range(int, int, Vector3 *);
+
+    AABoxClass &Get_Bounding_Box() { return m_boundingBox; }
```
xezon, line 222: This returns mutable reference for sure?

src/game/client/shadow/w3dvolumetricshadow.h
xezon: This returns mutable reference for sure?

tomsons26: yes

xezon (approved):

PR #705, opened 2022-03-31
Implements 3 more shadow functions.

tomsons26:


src/game/client/shadow/w3dshadow.h
```diff
@@ -110,6 +110,29 @@ class Shadow
         }
     }
 
+    void Set_Color(int color)
+    {
+        m_opacity = color & 0xFFFFFF;
+
+        if ((m_type & SHADOW_ALPHA_DECAL) != 0) {
+            m_color2 = (color << 24) | m_color1;
```
xezon: What is going on here? Does the blue channel become alpha channel?

src/game/client/shadow/w3dshadow.h
```diff
@@ -110,6 +110,29 @@ class Shadow
...
+    {
+        m_opacity = color & 0xFFFFFF;
+
+        if ((m_type & SHADOW_ALPHA_DECAL) != 0) {
+            m_color2 = (color << 24) | m_color1;
+        } else if ((m_type & SHADOW_ADDITIVE_DECAL) != 0) {
+            float o = m_opacity / 255.0f;
+            m_color2 = GameMath::Fast_To_Int_Truncate(((m_color1 >> 16) & 0xFF) * o)
+                | GameMath::Fast_To_Int_Truncate(((m_color1 >> 8) & 0xFF) * o)
+                | GameMath::Fast_To_Int_Truncate((m_color1 & 0xFF) * o);
```
xezon, line 123: I have seen this pattern of color values somewhere else in Thyme. Is it possible that there could be a macro made for this?

xezon: On second look, are you sure this logic is correct? It looks as if all color channels are pushed into the blue channel. Why?

tomsons26: dunno if the logics are correct but pseudo matches

xezon: Like so would make more sense if I am not mistaken.

Basically what happens is, color value is shifted to lowest 8 bits, masked and scaled by opacity, and then moved back to the byte position it belongs for R G B channel respectively.

```c++
            m_color2 = (GameMath::Fast_To_Int_Truncate(((m_color1 >> 16) & 0xFF) * o) << 16)
                | (GameMath::Fast_To_Int_Truncate(((m_color1 >> 8) & 0xFF) * o) << 8)
                | GameMath::Fast_To_Int_Truncate((m_color1 & 0xFF) * o);
```

PR #709, opened 2022-04-06
Various fixes to game-side code that are needed for w3dviewer

jonwil:


tomsons26 (approved):

src/w3d/lib/hashtemplate.h
```diff
@@ -213,20 +216,23 @@ template<typename Key, typename Value> class HashTemplateIterator
...
-        int size = m_hashTable->m_size;
 
-        for (m_hashIndex++; m_hashIndex < size; m_hashIndex++) {
-            m_handle = m_hashTable->m_hash[m_hashIndex];
+        if (m_handle == -1) {
+            int size = m_hashTable->m_size;
 
-            if (m_handle != -1) {
-                break;
+            for (m_hashIndex++; m_hashIndex < size; m_hashIndex++) {
```
xezon, line 223: Is `for (m_hashIndex++;` correct? It looks very strange.

src/w3d/lib/threadtrack.cpp
```diff
@@ -15,6 +15,11 @@
 #include "threadtrack.h"
 #include <cstring>
 
+#pragma warning(push)
+#pragma warning(disable : 4073) // warning C4073: initializers put in library initialization area
+#pragma init_seg(lib) // Forces objects and variables in this file to initialize before other stuff.
```
xezon, line 20: Why is this necessary? What happens without this?

PR #710, opened 2022-04-06
W3dviewer

jonwil:


src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -69,17 +69,17 @@ int CDataTreeView::OnCreate(LPCREATESTRUCT lpCreateStruct)
...
+        m_imageListIDs[ICON_ANIMATION] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_ANIMATION)));
+        m_imageListIDs[ICON_ANIMSPEED] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_ANIMSPEED)));
+        m_imageListIDs[ICON_TEXTURE] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_TEXTURE)));
+        m_imageListIDs[ICON_MESH] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_MESH)));
+        m_imageListIDs[ICON_MATERIAL] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_MATERIAL)));
+        m_imageListIDs[ICON_HLOD] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_HEIRARCHY)));
+        m_imageListIDs[ICON_AGGREGATE] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_HEIRARCHY)));
+        m_imageListIDs[ICON_EMITTER] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_HEIRARCHY)));
+        m_imageListIDs[ICON_HIERARCHY] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_HEIRARCHY)));
```
xezon: Shouldn't it be called `IDI_HIERARCHY` ?

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -69,17 +69,17 @@ int CDataTreeView::OnCreate(LPCREATESTRUCT lpCreateStruct)
...
-        m_imageListIDs[7] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_PRIMITIVE)));
-        m_imageListIDs[10] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_SOUND)));
+        m_imageListIDs[ICON_ANIMATION] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_ANIMATION)));
+        m_imageListIDs[ICON_ANIMSPEED] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_ANIMSPEED)));
+        m_imageListIDs[ICON_TEXTURE] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_TEXTURE)));
+        m_imageListIDs[ICON_MESH] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_MESH)));
+        m_imageListIDs[ICON_MATERIAL] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_MATERIAL)));
+        m_imageListIDs[ICON_HLOD] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_HEIRARCHY)));
+        m_imageListIDs[ICON_AGGREGATE] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_HEIRARCHY)));
+        m_imageListIDs[ICON_EMITTER] = il.Replace(-1, LoadIcon(AfxGetResourceHandle(), MAKEINTRESOURCE(IDI_HEIRARCHY)));
```
xezon: Why do HLOD, AGGREGATE, EMITTER not have their own resource?

src/tools/w3dviewer/assetinfo.h
```diff
@@ -0,0 +1,37 @@
...
+
+class RenderObjClass;
+class TextureClass;
+
+class AssetInfoClass
+{
+public:
+    AssetInfoClass(const char *name, AssetType type, RenderObjClass *robj, TextureClass *texture);
+    virtual ~AssetInfoClass();
+    void GetHeirarchyName();
```
xezon: `GetHierarchyName` ?

src/tools/w3dviewer/assetinfo.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+            if (robj == nullptr) {
+                return;
+            }
+        }
+
+        const HTreeClass *tree = robj->Get_HTree();
+
+        if (tree != nullptr) {
+            m_heirarchyName = tree->Get_Name();
```
xezon: `m_hierarchyName` ?

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+// clang-format off
+BEGIN_MESSAGE_MAP(CDataTreeView, CTreeView)
+    ON_WM_CREATE()
+    ON_NOTIFY_REFLECT(TVN_SELCHANGED, &OnSelectTree)
+    ON_NOTIFY_REFLECT(TVN_DELETEITEM, &OnDeleteItem)
+    ON_NOTIFY_REFLECT(NM_DBLCLK, &OnDblClk)
+END_MESSAGE_MAP()
+// clang-format on
+
+CDataTreeView::CDataTreeView() : m_restrictAnims(true), m_categoryTreeItems{}
```
xezon, line 40: Wrong init order.

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+        AddCategories();
+        return 0;
+    }
+
+    return result;
+}
+
+void CDataTreeView::OnSelectTree(NMHDR *pNMHDR, LRESULT *pResult)
+{
+    NMTREEVIEW *nm = (NMTREEVIEW *)pNMHDR;
```
xezon: `static_cast` ?

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+
+void CDataTreeView::AddRenderObjects()
+{
+    SetRedraw(FALSE);
+    RenderObjIterator *iter = W3DAssetManager::Get_Instance()->Create_Render_Obj_Iterator();
+
+    if (iter != nullptr) {
+        iter->First();
+
+        while (!iter->Is_Done()) {
```
xezon: Maybe can make this a compact for loop `for (iter->First(), !iter->Is_Done(), iter->Next)`

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+            if (W3DAssetManager::Get_Instance()->Render_Obj_Exists(name)) {
+                HTREEITEM category;
+                int image;
+                AssetType type;
+
+                switch (iter->Current_Item_Class_ID()) {
+                    case RenderObjClass::CLASSID_MESH:
+                        category = m_categoryTreeItems[CATEGORY_MESH];
+                        image = m_imageListIDs[ICON_MESH];
+                        type = ASSET_TYPE_MESH;
```
xezon, line 140: These 3 lines can be moved into else of condition below.

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+
+                            AssetInfoClass *info = new AssetInfoClass(name, type, nullptr, nullptr);
+                            GetTreeCtrl().SetItem(newitem, TVIF_PARAM, nullptr, 0, 0, 0, 0, (LPARAM)info);
+                        }
+
+                        break;
+                    case RenderObjClass::CLASSID_PARTICLEEMITTER:
+                        category = m_categoryTreeItems[CATEGORY_EMITTER];
+                        image = m_imageListIDs[ICON_EMITTER];
+                        type = ASSET_TYPE_PARTICLEEMITTER;
```
xezon, line 167: These 3 lines can be moved into else of condition below.

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+                        image = m_imageListIDs[ICON_HIERARCHY];
+                        type = ASSET_TYPE_HIERARCHY;
+
+                        if (HasMultipleLODs(name)) {
+                            category = m_categoryTreeItems[CATEGORY_HLOD];
+                            image = m_imageListIDs[ICON_HLOD];
+                            type = ASSET_TYPE_HLOD;
+                        }
+
+                        if (HasBaseModelName(name)) {
```
xezon, line 202: This can be rewritten as
```c++
if (HasBaseModelName(name)) {
}
else if (HasMultipleLODs(name)) {
}
else {
}
```

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+}
+
+void CDataTreeView::Select(HTREEITEM item)
+{
+    if (item == nullptr) {
+        item = GetTreeCtrl().GetSelectedItem();
+    }
+
+    if (item != nullptr) {
+        AssetInfoClass *info = (AssetInfoClass *)GetTreeCtrl().GetItemData(item);
```
xezon: `static_cast` ?

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+
+        if (info != nullptr) {
+            CW3DViewDoc *document = ((CW3DViewDoc *)m_pDocument);
```
xezon: `static_cast` ?

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+                        if (parent != nullptr) {
+                            RenderObjClass *robj = GetRenderObj(parent);
+                            document->SetAnimationByName(robj, info->m_name, true, true);
+                            Ref_Ptr_Release(robj);
+                        }
+
+                    } break;
+                    case ASSET_TYPE_PARTICLEEMITTER: {
+                        RenderObjClass *robj = GetRenderObj(item);
+                        document->SetParticleEmitter((ParticleEmitterClass *)robj, true, true);
```
xezon: `static_cast` ?

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+                        }
+                    }
+                }
+            }
+        }
+
+        return;
+    }
+
+    CW3DViewDoc *document = ((CW3DViewDoc *)m_pDocument);
```
xezon: `static_cast` ?

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+
+    if (document != nullptr) {
+        document->SetRenderObject(nullptr, true, true, false);
+        CMainFrame *frame = ((CMainFrame *)AfxGetMainWnd());
```
xezon: `static_cast` ?

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+            }
+        }
+    }
+
+    return robj;
+}
+
+void CDataTreeView::AddTextures()
+{
+    for (HashTemplateIterator<StringClass, TextureClass *> textureIter(W3DAssetManager::Get_Instance()->Texture_Hash());
```
xezon: This would be a bit easier to look at by moving the `HashTemplateIterator<StringClass, TextureClass *> textureIter(W3DAssetManager::Get_Instance()->Texture_Hash()` out of the for loop.

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+                        i,
+                        TVI_SORT);
+
+                    AssetInfoClass *info =
+                        new AssetInfoClass(iterator->Current_Item_Name(), ASSET_TYPE_ANIMATION, nullptr, nullptr);
+                    GetTreeCtrl().SetItem(newitem, TVIF_PARAM, nullptr, 0, 0, 0, 0, (LPARAM)info);
+                }
+            }
+
+            for (HTREEITEM i = FindFirstItem(m_categoryTreeItems[CATEGORY_HLOD], anim->Get_HName()); i != nullptr;
```
xezon, line 522: What is the difference in code of these 3 for loops other than the different `m_categoryTreeItems` ? Can this be refactored into 1 sub function and then called 3 times here, reducing source code size?

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+            item1 = item2;
+        }
+    }
+
+    return item1;
+}
+
+HTREEITEM CDataTreeView::FindNextItem(HTREEITEM item, const char *name)
+{
+    HTREEITEM found = 0;
```
xezon: `nullptr` ?

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+            found = i;
+        }
+    }
+
+    return found;
+}
+
+HTREEITEM CDataTreeView::FindFirstItem(HTREEITEM item, const char *name)
+{
+    HTREEITEM found = 0;
```
xezon: `nullptr` ?

src/tools/w3dviewer/datatreeview.cpp
```diff
@@ -0,0 +1,680 @@
...
+    if (selected != nullptr) {
+        HTREEITEM prev = GetTreeCtrl().GetPrevSiblingItem(selected);
+
+        if (prev != nullptr) {
+            GetTreeCtrl().SelectItem(prev);
+        }
+    }
+}
+
+void CDataTreeView::RestrictAnims(bool restrict)
```
xezon: Restrict is a c++ keyword. Maybe can use different name to avoid wrong syntax highlighting in Editors.

src/tools/w3dviewer/datatreeview.cpp
jonwil: No idea, ask Westwood...

src/tools/w3dviewer/datatreeview.cpp
jonwil: Fixed.

src/tools/w3dviewer/assetinfo.h
jonwil: fixed.

src/tools/w3dviewer/assetinfo.cpp
jonwil: fixed.

src/tools/w3dviewer/datatreeview.cpp
jonwil: Fixed all the C style pointer casts.

src/tools/w3dviewer/datatreeview.cpp
jonwil: Fixed missing nullptrs.

src/tools/w3dviewer/datatreeview.cpp
jonwil: Fixed missing nullptrs.

src/tools/w3dviewer/datatreeview.cpp
jonwil: fixed.

src/tools/w3dviewer/datatreeview.cpp
jonwil: Fixed.

src/tools/w3dviewer/datatreeview.cpp
jonwil: fixed.

src/tools/w3dviewer/datatreeview.h
```diff
@@ -0,0 +1,93 @@
...
+#pragma once
+#include "w3dafx.h"
+#include "vector.h"
+
+class RenderObjClass;
+
+class CDataTreeView : public CTreeView
+{
+public:
+    virtual ~CDataTreeView() override {}
```
xezon, line 24: Does `m_categoryTreeItems` member need cleanup here? I am asking because HTREEITEM is a pointer and it is not clear to me if this is owning pointer or not.

```c++
typedef struct _TREEITEM *HTREEITEM;
```

src/tools/w3dviewer/graphicview.h
```diff
@@ -0,0 +1,125 @@
...
+    {
+        DIRECTION_FRONT = -1,
+        DIRECTION_BACK = 0,
+        DIRECTION_TOP = 1,
+        DIRECTION_BOTTOM = 2,
+        DIRECTION_LEFT = 3,
+        DIRECTION_RIGHT = 4,
+    };
+
+    virtual ~CGraphicView() override {}
```
xezon, line 66: Maybe add some asserts in the destructor to check if cleanup of `OnDestroy` was successful.

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+#include "utils.h"
+#include "viewerscene.h"
+#include "w3d.h"
+#include "w3dview.h"
+#include "w3dviewdoc.h"
+
+extern int DeviceWidth;
+extern int DeviceHeight;
+extern int BPP;
+extern int Device;
```
xezon, line 35: Hmm. Why are there `extern` variables in cpp file? I would expect them in header only.

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+
+Vector3 Pos;
+float Zoom;
+Quaternion Rotation;
```
xezon, line 39: Maybe initialize all of these globals?

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+    CW3DViewDoc *document = reinterpret_cast<CW3DViewDoc *>(m_pDocument);
+
+    if (document != nullptr) {
+        document->Initialize();
+    }
+}
+
+int CGraphicView::OnCreate(LPCREATESTRUCT lpCreateStruct)
+{
+    int result = CView::OnCreate(lpCreateStruct);
```
xezon, line 133: OnCreate also returns LRESULT type?

https://docs.microsoft.com/en-us/windows/win32/wmp/the-oncreate-method

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+
+        Vector3 v;
+
+        v.X = (((float)point.x - (float)(r.right >> 1)) / (float)(r.right >> 1)
+                  - ((float)m_mousePos.x - (float)(r.right >> 1)) / (float)(r.right >> 1))
+            * m_radius * -1.0f;
+
+        v.Y = (((float)point.y - (float)(r.bottom >> 1)) / (float)(r.bottom >> 1)
+                  - ((float)m_mousePos.y - (float)(r.bottom >> 1)) / (float)(r.bottom >> 1))
+            * m_radius * -1.0f;
```
xezon, line 243: These 2 operations or `Vector3 v` are hard to read. May I suggest to break them up a bit to increase readability? Similar to a bit further down below in line 264.

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+            obj->Set_LOD_Level(0);
+        }
+    }
+}
+
+void CGraphicView::Render(BOOL update, unsigned int time)
+{
+    static bool IsRendering = false;
+
+    if (!IsRendering) {
```
xezon, line 554: Does this condition serve any purpose? It looks like it does nothing. Can it be removed along with `IsRendering`?

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+        IsRendering = true;
+        CW3DViewDoc *document = reinterpret_cast<CW3DViewDoc *>(m_pDocument);
+
+        if (document->m_initialized && document->m_scene != nullptr && !m_renderingDisabled) {
+            unsigned int newtime = timeGetTime();
+            unsigned int currenttime = m_time;
```
xezon: Maybe `previous_time`

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+            m_time = newtime;
+            int tm = newtime - currenttime;
```
xezon: Maybe `delta_time`

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+
+    float hfov;
+    float vfov;
+
+    if (height <= width) {
+        hfov = 0.7853981852531433f;
+        vfov = (float)height / (float)width * 0.7853981852531433f;
+    } else {
+        hfov = (float)height / (float)width * 0.7853981852531433f;
+        vfov = 0.7853981852531433f;
```
xezon: This value is `DEG_TO_RAD(45.f)`. Can make it a `constexpr` to type it just once.

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+    Zoom = m_radius * 0.0052631581f;
+    Rotation = Build_Quaternion(tm);
+    m_camera->Set_Transform(tm);
+    CW3DViewDoc *document = reinterpret_cast<CW3DViewDoc *>(m_pDocument);
+    LightClass *light = document->m_light;
+
+    if (m_light != nullptr && light != nullptr) {
+        tm.Make_Identity();
+        tm.Adjust_Translation(Pos);
+        tm[2].W += m_radius * 0.69999999f;
```
xezon: Likely is `0.7f` ?

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+    if (znear < 0.2f) {
+        znear = 0.2f;
+    }
+
+    if (!document->m_useManualClipPlanes) {
+        m_camera->Set_Clip_Planes(znear, zfar);
+        float start;
+        float end;
+        document->m_scene->Get_Fog_Range(&start, &end);
+        document->m_scene->Set_Fog_Range(end, end);
```
xezon: `end, end` is correct?

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+    float f9 = 0.0f;
+    float f10 = 0.0f;
+    float f11 = f6 + f5;
+    float f12 = 0.0f;
+    float f13 = f1 * acceleration.Z * 0.5f + f7;
+    float f14;
+
+    if (acceleration.X == 0.0f) {
+        if (acceleration.Y == 0.0f && acceleration.Z == 0.0f) {
+            goto l1;
```
xezon: You can rewrite this condition to get rid of the `goto`.

```c++
if (acceleration.X != 0.0f && acceleration.Y != 0.0f && acceleration.Z != 0.0f) {
    if (acceleration.X == 0.0f) {
        f14 = 0.0f;
    } else {
        f14 = -s.Center.X / acceleration.X;
    }
    // ... all the other code until the goto label
}
```

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+        }
+    }
+}
+
+void CGraphicView::UpdateCameraDistance(float distance)
+{
+    m_radius = distance;
+    Matrix3D tm(true);
+    tm.Look_At(
+        Vector3(distance + m_objSphere.Center.X, m_objSphere.Center.Y, m_objSphere.Center.Z), m_objSphere.Center, 0.0f);
```
xezon: Can take out the `Vector3(distance + m_objSphere.Center.X, m_objSphere.Center.Y, m_objSphere.Center.Z)` in separate line to improve readability.

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+    if (document->m_model != nullptr) {
+        Vector3 center = m_objSphere.Center;
+        m_radius = m_objSphere.Radius * 3.0f;
+        m_radius = std::clamp(m_radius, 1.0f, 400.0f);
+        Matrix3D tm(true);
+        Vector3 v;
+
+        switch (direction) {
+            case DIRECTION_FRONT:
+                v.X = m_radius + center.X;
```
xezon: Minor style: `v.X = center.X + m_radius`

Also a few times below.

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+                v.X = center.X - m_radius;
+                v.Y = center.Y;
+                v.Z = center.Z;
+                tm.Look_At(v, center, 0.0f);
+                break;
+            case DIRECTION_TOP:
+                v.X = center.X;
+                v.Y = center.Y;
+                v.Z = m_radius + center.Z;
+                tm.Look_At(v, center, 3.1415927f);
```
xezon: `DEG_TO_RADF(180.f)`

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -0,0 +1,1095 @@
...
+                break;
+            case DIRECTION_BOTTOM:
+                v.X = center.X;
+                v.Y = center.Y;
+                v.Z = center.Z - m_radius;
+                tm.Look_At(v, center, 3.1415927f);
```
xezon: `DEG_TO_RADF(180.f)`

src/tools/w3dviewer/datatreeview.h
jonwil: No it does not.

src/tools/w3dviewer/graphicview.cpp
jonwil: Wrong documentation, that's not MFC.

src/tools/w3dviewer/graphicview.cpp
jonwil: fixed.

src/tools/w3dviewer/graphicview.cpp
jonwil: fixed.

src/tools/w3dviewer/graphicview.cpp
jonwil: fixed.

src/tools/w3dviewer/graphicview.cpp
jonwil: fixed.

src/tools/w3dviewer/graphicview.cpp
jonwil: Yes.

src/tools/w3dviewer/graphicview.cpp
jonwil: fixed.

src/tools/w3dviewer/graphicview.cpp
jonwil: fixed.

src/tools/w3dviewer/graphicview.cpp
jonwil: fixed.

src/tools/w3dviewer/graphicview.cpp
jonwil: fixed.

src/tools/w3dviewer/graphicview.cpp
jonwil: IsRendering is static so its designed to make sure Render isn't run if its already running.

src/tools/w3dviewer/graphicview.cpp
jonwil: fixed.

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+        return CFrameWnd::OnCmdMsg(nID, nCode, pExtra, pHandlerInfo);
+    }
+
+    if (pExtra == nullptr) {
+        return CFrameWnd::OnCmdMsg(nID, nCode, pExtra, pHandlerInfo);
+    }
+
+    CCmdUI *pCmdUI = static_cast<CCmdUI *>(pExtra);
+
+    if (pCmdUI->m_nID < 1000 || pCmdUI->m_nID > 1100) {
```
xezon, line 204: Can we have some named value for the 1000 and 1100 ? Why is it these numbers?

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+}
+
+CMainFrame::~CMainFrame() {}
+
+BOOL CMainFrame::PreCreateWindow(CREATESTRUCT &cs)
+{
+    return CFrameWnd::PreCreateWindow(cs);
+}
+
+BOOL CMainFrame::OnCmdMsg(UINT nID, int nCode, void *pExtra, AFX_CMDHANDLERINFO *pHandlerInfo)
```
xezon, line 192: This function can call `CFrameWnd::OnCmdMsg` up to 3 times in one go. Is this correct?

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+    if (ret) {
+        ret &= m_splitter.CreateView(0, 0, CDataTreeView::GetThisClass(), SIZE{ 340, 10 }, pContext);
+        ret &= m_splitter.CreateView(0, 1, CGraphicView::GetThisClass(), SIZE{ 120, 10 }, pContext);
+
+        if (ret) {
+            CWnd *pane = m_splitter.GetPane(0, 1);
+
+            if (pane != nullptr) {
+                char path[MAX_PATH];
+                GetModuleFileName(nullptr, path, MAX_PATH);
```
xezon, line 261: `GetModuleFileNameA`

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+                char *str = strrchr(path, '\\');
+
+                if (str != nullptr) {
+                    *str = 0;
```
xezon: `'\0'`

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+                    SetCurrentDirectory(path);
+                }
+
+                W3D::Init(pane->m_hWnd);
+                W3D::Enable_Static_Sort_Lists(true);
+                DeviceWidth = theApp.GetProfileInt("Config", "DeviceWidth", 640);
+                DeviceHeight = theApp.GetProfileInt("Config", "DeviceHeight", 480);
+                GetDevice(false);
+                W3D::Enable_Munge_Sort_On_Load(AfxGetApp()->GetProfileInt("Config", "MungeSortOnLoad", 0) == 1);
+                W3D::Enable_Sorting(AfxGetApp()->GetProfileInt("Config", "EnableSorting", 1) == 1);
```
xezon, line 275: You can move the compares out of the function argument list to make it a bit easier on the eyes.

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+
+void CMainFrame::RestoreWindowPos()
+{
+    int left = theApp.GetProfileInt("Window", "Left", -1);
+    int right = theApp.GetProfileInt("Window", "Right", -1);
+    int top = theApp.GetProfileInt("Window", "Top", -1);
+    int bottom = theApp.GetProfileInt("Window", "Bottom", -1);
+    bool maximized = theApp.GetProfileInt("Window", "Maximized", -1) == 1;
+
+    if (left != -1 && right != -1 && top != -1 && bottom != -1) {
```
xezon, line 291: This logic looks odd. When moving window 1 pixel left of the left screen border, does this mean the window could not accept other coordinates?

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+
+void CMainFrame::UpdateMenus(AssetType type)
+{
+    if (m_currentType != type) {
+        if (m_currentType >= ASSET_TYPE_HIERARCHY
+            && (m_currentType <= ASSET_TYPE_ANIMATION || m_currentType == ASSET_TYPE_MORPH)) {
+            CMenu *menu = GetMenu();
+
+            if (menu != nullptr) {
+                menu->RemoveMenu(6, MF_BYPOSITION | MF_DELETE);
```
xezon, line 313: What does this 6 stand for? A few times in this function. Is this like an index? Maybe make it a named constexpr or enum and use it for all occurrences here. It appears to be used in other functions as well, so appears to have specific meaning.

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+        }
+
+        m_currentType = type;
+    }
+}
+
+void CMainFrame::UpdateStatusBar(unsigned int time)
+{
+    static int Time = 0;
+    Time += time;
```
xezon, line 404: `time` is unsigned but `Time` is signed. Why?

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+    static int Frames = 0;
+    Frames++;
+    static int Ticks = 0;
```
xezon: `Ticks` is int, but shouldn't it be DWORD because GetTickCount() returns DWORD?

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+    CString str;
+    str.Format("Frame %d/%d at %.2f fps", frame, framecount, fps);
+    m_statusBar.SetPaneText(4, str);
+}
+
+void CMainFrame::UpdateCameraDistance(float distance)
+{
+    CString str;
+    str.Format("Camera %.3f", distance);
+    m_statusBar.SetPaneText(3, str);
```
xezon, line 454: Maybe create an enum

```c++
enum MainFramePanes
{
    PANE_POLYGON_COUNT = 1,
    PANE_PARTICLE_COUNT = 2,
    PANE_CAMERA_DISTANCE = 3,
    PANE_FRAME_COUNT = 4,
    ...
}
```

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+    m_subMenu = SubMenu;
+    MENUITEMINFO info;
+    memset(&info, 0, sizeof(info));
+    info.cbSize = sizeof(info);
+    info.fMask = MIIM_STRING;
+    char buffer[200];
+    info.dwTypeData = buffer;
+    info.cch = 100;
+    GetMenuItemInfo(SubMenu, 4, TRUE, &info);
+    m_subMenu = GetSubMenu(SubMenu, 4);
```
xezon: I think these submenu indexes also need a named Enum:

```c++
enum MainFrameSubMenu
{
    SUBMENU_SOMETHING = 3,
    SUBMENU_WHATEVER = 4,
}
```

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+    HMENU menu = ::GetMenu(m_hWnd);
+    HMENU SubMenu = GetSubMenu(menu, 3);
+    m_subMenu = SubMenu;
+    MENUITEMINFO info;
+    memset(&info, 0, sizeof(info));
+    info.cbSize = sizeof(info);
+    info.fMask = MIIM_STRING;
+    char buffer[200];
+    info.dwTypeData = buffer;
+    info.cch = 100;
```
xezon: I am very confused by this.

https://docs.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-menuiteminfoa

> cch is used only if the MIIM_TYPE flag is set in the fMask member

and

> The value of cch+1 is the size needed. Then allocate a buffer of this size, place the pointer to the buffer in dwTypeData, increment cch, and call GetMenuItemInfo once again

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+            CWnd::Default();
+            return;
+        }
+
+        RemoveProp(m_hWnd, "Inactive");
+        view->m_time = GetTickCount();
+    }
+
+    CWnd::Default();
+    CWnd::Default();
```
xezon: Why is `CWnd::Default` called twice?

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1616 @@
...
+    CW3DViewDoc *doc = reinterpret_cast<CW3DViewDoc *>(GetActiveDocument());
+
+    if (doc != nullptr) {
+        CFileDialog dlg(TRUE,
+            ".w3d",
+            nullptr,
+            OFN_EXPLORER | OFN_ALLOWMULTISELECT | OFN_HIDEREADONLY | OFN_OVERWRITEPROMPT,
+            "Westwood 3D Files (*.w3d)|*.w3d||",
+            this);
+        char file[2600] = {};
```
xezon, line 607: This buffer is 2600 ? I assume because user can write multiple chained file paths maybe?

src/tools/w3dviewer/mainfrm.cpp
jonwil: No it wont call it 3 times, notice the return statements?

src/tools/w3dviewer/mainfrm.cpp
jonwil: fixed.

src/tools/w3dviewer/mainfrm.cpp
jonwil: fixed.

src/tools/w3dviewer/mainfrm.cpp
jonwil: fixed.

src/tools/w3dviewer/mainfrm.cpp
jonwil: fixed.

src/tools/w3dviewer/mainfrm.cpp
jonwil: Yes.

src/tools/w3dviewer/mainfrm.cpp
jonwil: fixed, the GetMenuItemInfo call wasn't even supposed to be there.

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1607 @@
...
+
+void CMainFrame::OnExportSoundObject()
+{
+    // do later
+}
+
+void CMainFrame::OnWireframeMode()
+{
+    CW3DViewDoc *document = GetCurrentDocument();
+    document->m_scene->Set_Polygon_Mode((SceneClass::PolyRenderType)((document->m_scene->Get_Polygon_Mode() == 1) + 1));
```
xezon:
```c++
(document->m_scene->Get_Polygon_Mode() == 1) + 1)
```
reads like something a compiler would write and only works given the specific enum values of 
```c++
    enum PolyRenderType
    {
        POINT = 0x0,
        LINE = 0x1,
        FILL = 0x2,
    };
```
Rewrite as:
```c++
SceneClass::PolyRenderType curType = document->m_scene->Get_Polygon_Mode();
SceneClass::PolyRenderType newType = curType == SceneClass::LINE ? SceneClass::FILL : SceneClass::LINE;
document->m_scene->Set_Polygon_Mode(newType);
```

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1607 @@
...
+}
+
+void CMainFrame::OnUpdateImportFacial(CCmdUI *pCmdUI)
+{
+    // do later
+}
+
+void CMainFrame::OnUpdateRestrictAnims(CCmdUI *pCmdUI)
+{
+    bool restrict = true;
```
xezon: `restrict` is c++ keyword. Maybe use different name.

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1607 @@
...
+    CGraphicView *view = reinterpret_cast<CGraphicView *>(m_splitter.GetPane(0, 1));
+
+    if (view != nullptr) {
+        pCmdUI->SetCheck(view->m_plusXCamera);
+    }
+}
+
+void CMainFrame::OnUpdateMungeSort(CCmdUI *pCmdUI)
+{
+    pCmdUI->SetCheck(W3D::Is_Munge_Sort_On_Load_Enabled() != 0);
```
xezon: `!= 0` ?

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1607 @@
...
+}
+
+void CMainFrame::OnUpdateScaleEmitter(CCmdUI *pCmdUI)
+{
+    // do later
+}
+
+void CMainFrame::OnUpdatePolygonSorting(CCmdUI *pCmdUI)
+{
+    pCmdUI->SetCheck(W3D::Is_Sorting_Enabled() != 0);
```
xezon: `!= 0` ?

src/tools/w3dviewer/toolbar.cpp
```diff
@@ -0,0 +1,265 @@
...
+        }
+    }
+}
+
+BOOL CFancyToolbar::PreCreateWindow(CREATESTRUCT &cs)
+{
+    return CControlBar::PreCreateWindow(cs);
+}
+
+CSize CFancyToolbar::CalcFixedLayout(BOOL bStretch, BOOL bHorz)
```
xezon, line 63: `BOOL bStretch, BOOL bHorz` unused?

src/tools/w3dviewer/toolbar.cpp
```diff
@@ -0,0 +1,265 @@
...
+{
+    return CSize(42 * m_itemCount + 16, 52);
+}
+
+CSize CFancyToolbar::CalcDynamicLayout(int nLength, DWORD nMode)
```
xezon, line 68: `int nLength, DWORD nMode` unused?

src/tools/w3dviewer/toolbar.cpp
```diff
@@ -0,0 +1,265 @@
...
+        if (m_items[id].m_isActive) {
+            bitmap = m_items[id].m_activeBitmap;
+        }
+
+        HDC dc = ::GetDC(m_hWnd);
+
+        if (dc != nullptr) {
+            RECT r;
+            GetClientRect(&r);
+            int cx = ((r.bottom - r.top) >> 1) - 18;
```
xezon, line 156: What is going on here?

src/tools/w3dviewer/toolbar.cpp
```diff
@@ -0,0 +1,265 @@
...
+            HDC dc2 = CreateCompatibleDC(dc);
+
+            if (dc2 != nullptr) {
+                HGDIOBJ obj = SelectObject(dc2, bitmap);
+                BitBlt(dc, 42 * id + 8, cx, 42, 36, dc2, 0, 0, SRCCOPY);
```
xezon, line 161: Can you please add named constants for these magic numbers in this function and class?

src/tools/w3dviewer/mainfrm.cpp
jonwil: fixed.

src/tools/w3dviewer/mainfrm.cpp
jonwil: fixed.

src/tools/w3dviewer/mainfrm.cpp
jonwil: fixed.

src/tools/w3dviewer/toolbar.cpp
jonwil: Yes.

src/tools/w3dviewer/toolbar.cpp
jonwil: Yes.

src/tools/w3dviewer/mainfrm.cpp
jonwil: fixed.

src/tools/w3dviewer/viewerscene.cpp
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    m_visibilityChecked = true;
+}
+
+void ViewerSceneClass::Add_LOD_Object(RenderObjClass *obj)
+{
+    if (obj->Class_ID() == RenderObjClass::CLASSID_HLOD) {
+        AABoxClass box = obj->Get_Bounding_Box();
```
xezon: lod_box

src/tools/w3dviewer/viewerscene.cpp
```diff
@@ -0,0 +1,167 @@
...
+        float f1 = box.m_extent.Y + box.m_extent.Y;
+        AABoxClass box2;
```
xezon: this_box

src/tools/w3dviewer/viewerscene.cpp
```diff
@@ -0,0 +1,167 @@
...
+        Get_Bounding_Box(&box2);
+        float f2 = box2.m_extent.Y + box.m_extent.Y;
```
xezon:
```c++
float f1 = box.m_extent.Y + box.m_extent.Y;
float f2 = box2.m_extent.Y + box.m_extent.Y;
```
is correct for sure?

src/tools/w3dviewer/w3dview.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+CW3DViewApp theApp;
+
+BOOL CW3DViewApp::InitInstance()
+{
+    // do later
+    // ColorPickerClass::RegisterWndClass(AfxGetInstanceHandle());
+    // ColorBarClass::RegisterWndClass(AfxGetInstanceHandle());
+    GameMath::Init();
+    CWinApp::SetRegistryKey("Westwood Studios");
```
xezon, line 31: I don't think we should use this company name.

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+        if (m_scene != nullptr && m_model != nullptr) {
+            RemoveRenderObject(m_model);
+        }
+
+        if (m_scene != nullptr) {
+            m_scene->Remove_All_LOD_Objects();
+            m_scene->Set_Fog_Color(m_fogColor);
+        }
+
+        if (m_model != nullptr) {
```
xezon, line 92: With this test in place it looks like `m_animation` could leak. Maybe omit this test, because `Ref_Ptr_Release` already tests null?

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+            if (view != nullptr) {
+                view->GetTreeCtrl().DeleteItem(TVI_ROOT);
+                view->AddCategories();
+            }
+        }
+
+        Free();
+        W3DAssetManager::Get_Instance()->Free_Assets();
+
+        return TRUE;
```
xezon, line 111: This return can be omitted.

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+            if (tree != nullptr) {
+                tree->AddRenderObjects();
+            }
+        }
+
+        if (view != nullptr) {
+            view->EnableRendering(true);
+        }
+
+        return TRUE;
```
xezon, line 143: This return can be omitted.

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+        m_scene->Remove_All_LOD_Objects();
+        Ref_Ptr_Release(m_scene);
+    }
+
+    Ref_Ptr_Release(m_textureCamera);
+    Ref_Ptr_Release(m_docCamera);
+    Ref_Ptr_Release(m_backgroundBitmap);
+    Ref_Ptr_Release(m_light);
+
+    if (m_model != nullptr) {
```
xezon, line 185: With this test in place it looks like `m_animation` could leak. Maybe omit this test, because `Ref_Ptr_Release` already tests null?

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+    if (m_textureScene != nullptr) {
+        if (m_backgroundBitmap != nullptr) {
+            m_backgroundBitmap->Remove();
+        }
+
+        Ref_Ptr_Release(m_textureScene);
+    }
+
+    if (m_backgroundScene != nullptr) {
+        if (m_backgroundObject != nullptr) {
```
xezon, line 160: Where is this allocated? Unable to find.

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+        return TRUE;
+    }
+
+    return result;
+}
+
+void CW3DViewDoc::Free()
+{
+    if (m_textureScene != nullptr) {
+        if (m_backgroundBitmap != nullptr) {
```
xezon, line 152: Where is this allocated? Unable to find.

src/tools/w3dviewer/viewerscene.cpp
jonwil: fixed.

src/tools/w3dviewer/viewerscene.cpp
jonwil: fixed.

src/tools/w3dviewer/viewerscene.cpp
jonwil: fixed.

src/tools/w3dviewer/w3dviewdoc.cpp
jonwil: m_animation would only be set if m_model is also set.

src/tools/w3dviewer/w3dviewdoc.cpp
jonwil: m_animation would only be set if m_model is also set.

src/tools/w3dviewer/w3dview.cpp
jonwil: This is only so that our w3dviewer is compatible when it comes to stored settings/recent files/etc with the WW one (something I want)

src/tools/w3dviewer/w3dviewdoc.cpp
jonwil: Its allocated by code I haven't cloned yet.

src/tools/w3dviewer/w3dviewdoc.cpp
jonwil: Its allocated by code I haven't cloned yet.

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+
+    CGraphicView *view = GetCurrentGraphicView();
+
+    if (view != nullptr) {
+        view->EnableRendering(false);
+    }
+
+    if (strrchr(filename, '\\') != nullptr) {
+        CString str = filename;
+        str = str.Left(strrchr(filename, '\\') - filename);
```
xezon, line 236: `strrchr(filename, '\\')` is called twice here. Can reduce to one.

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+        g_theSimpleFileFactory->Append_Sub_Directory(str);
+    }
+
+    const char *c = strrchr(filename, '.');
+
+    if (lstrcmpi(c, ".tga") != 0 && lstrcmpi(c, ".dds") != 0) {
+        W3DAssetManager::Get_Instance()->Load_3D_Assets(GetFilenameFromPath(filename));
+    } else {
+        TextureClass *texture = W3DAssetManager::Get_Instance()->Get_Texture(GetFilenameFromPath(filename));
+        Ref_Ptr_Release(texture);
```
xezon, line 247: What is this for? Gets texture and does nothing with it?

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+    if (strrchr(filename, '\\') != nullptr) {
+        CString str = filename;
+        str = str.Left(strrchr(filename, '\\') - filename);
+        SetCurrentDirectory(str);
+        g_theSimpleFileFactory->Append_Sub_Directory(str);
+    }
+
+    const char *c = strrchr(filename, '.');
+
+    if (lstrcmpi(c, ".tga") != 0 && lstrcmpi(c, ".dds") != 0) {
```
xezon, line 243: Maybe invert the condition to read better?

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+            if (o != nullptr) {
+                count += GetNumParticles(o);
+                o->Release_Ref();
+            }
+        }
+
+        if (robj->Class_ID() == RenderObjClass::CLASSID_PARTICLEEMITTER) {
+
+            if (reinterpret_cast<ParticleEmitterClass *>(robj)->Peek_Buffer() != nullptr) {
+                count += reinterpret_cast<ParticleEmitterClass *>(robj)->Peek_Buffer()->Get_Particle_Count();
```
xezon: `static_cast`

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+            m_scene->Remove_All_LOD_Objects();
+            m_scene->Set_Fog_Color(m_fogColor);
+        }
+
+        if (m_model != nullptr) {
+            Ref_Ptr_Release(m_animation);
+            Ref_Ptr_Release(m_model);
+        }
+
+        CMainFrame *frame = reinterpret_cast<CMainFrame *>(AfxGetMainWnd());
```
xezon: `static_cast`

This and all other occurrences.

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1609 @@
...
+        }
+    }
+
+    return CFrameWnd::OnCommand(wParam, lParam);
+}
+
+LRESULT CMainFrame::WindowProc(UINT message, WPARAM wParam, LPARAM lParam)
+{
+    if (message == WM_CLOSE) {
+        CW3DViewDoc *doc = reinterpret_cast<CW3DViewDoc *>(GetActiveDocument());
```
xezon: `static_cast` or no cast if possible. Here and all other occurrences.

xezon (commented):
Code uses reinterpret_cast very often. But by the looks of it just `static_cast` or no cast is required.

src/tools/w3dviewer/w3dviewdoc.cpp
jonwil: I don't know why ww did that but they did.

src/tools/w3dviewer/w3dviewdoc.cpp
jonwil: Changed to static_cast where possible.

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1610 @@
...
+}
+
+BOOL CMainFrame::OnCommand(WPARAM wParam, LPARAM lParam)
+{
+    if (lParam == 0 && LOWORD(wParam) >= 1000 && LOWORD(wParam < 1100)) {
+        MENUITEMINFO info;
+        memset(&info, 0, sizeof(info));
+        info.cbSize = sizeof(info);
+        info.fMask = MIIM_DATA | MIIM_TYPE;
+        info.dwTypeData = reinterpret_cast<char *>(&info.hbmpItem);
```
xezon: Did you check if this is legal? It looks very suspicious. A quick glance at documentation does not make it obvious if this is proper.

https://docs.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-menuiteminfoa

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1610 @@
...
+        return CFrameWnd::OnCmdMsg(nID, nCode, pExtra, pHandlerInfo);
+    }
+
+    pCmdUI->Enable();
+    return TRUE;
+}
+
+BOOL CMainFrame::OnCommand(WPARAM wParam, LPARAM lParam)
+{
+    if (lParam == 0 && LOWORD(wParam) >= 1000 && LOWORD(wParam < 1100)) {
```
xezon, line 214: What does 1000 and 1100 refer to?

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1610 @@
...
+        MENUITEMINFO info;
+        memset(&info, 0, sizeof(info));
+        info.cbSize = sizeof(info);
+        info.fMask = MIIM_DATA | MIIM_TYPE;
+        info.dwTypeData = reinterpret_cast<char *>(&info.hbmpItem);
+        info.cch = 200;
```
xezon: Why is this 200 ?

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+            if (static_cast<ParticleEmitterClass *>(robj)->Peek_Buffer() != nullptr) {
+                count += static_cast<ParticleEmitterClass *>(robj)->Peek_Buffer()->Get_Particle_Count();
+            }
+        }
+    }
+
+    return count;
+}
+
+void CW3DViewDoc::SetRenderObject(RenderObjClass *robj, bool unk1, bool unk2, bool unk3)
```
xezon:
```c++
bool use_regular_camera_reset, bool reset_camera, bool preserve_model
```

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+
+    bool b = unk1 ? this->m_resetCamera : unk2;
+
+    if (b || this->m_forceCameraReset) {
+        view->ResetCamera(robj);
+        m_forceCameraReset = false;
+    }
+}
+
+void CW3DViewDoc::SetParticleEmitter(ParticleEmitterClass *emitter, bool unk1, bool unk2)
```
xezon:
```c++
bool use_regular_camera_reset, bool reset_camera
```

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+    if (unk3) {
+        robj->Set_Animation();
+        Matrix3D tm(true);
+        robj->Set_Transform(tm);
+        SceneClass *scene;
+
+        if (robj->Class_ID() == RenderObjClass::CLASSID_BITMAP2D) {
+            scene = m_textureScene;
+        } else {
+            m_scene->Remove_All_LOD_Objects();
```
xezon, line 451: Why is this called here when it is already called on line 436?

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+        if (index > 0) {
+            *tm = robj->Get_Bone_Transform(index);
+            return true;
+        }
+    }
+
+    return false;
+}
+
+void CW3DViewDoc::SetAnimationByName(RenderObjClass *robj, const char *name, bool unk1, bool unk2)
```
xezon:
```c++
bool use_regular_camera_reset, bool reset_camera
```

src/tools/w3dviewer/w3dviewdoc.h
```diff
@@ -0,0 +1,122 @@
...
+    CameraClass *m_textureCamera;
+    CameraClass *m_docCamera;
+    Vector3 m_fogColor;
+    CString m_backgroundBitmapName;
+    CString m_backgroundObjectName;
+    bool m_isQCompressed;
+    int m_compressType;
+    float m_frameCount;
+    float m_time;
+    BOOL m_blendFrames;
```
xezon, line 110: Is this meant to be type BOOL?

src/tools/w3dviewer/w3dviewdoc.cpp
```diff
@@ -0,0 +1,957 @@
...
+                }
+            }
+        }
+    }
+}
+
+void CW3DViewDoc::ToggleAlternateMaterials(RenderObjClass *robj)
+{
+    if (robj == nullptr) {
+        robj = m_model;
```
xezon, line 823: What if `m_model` is null? Can put an assert if it is not meant to be null here ever by design.

src/tools/w3dviewer/mainfrm.cpp
jonwil: Found out what was going on and fixed it.

src/tools/w3dviewer/mainfrm.cpp
jonwil: Because that's the size of the buffer.

src/tools/w3dviewer/w3dviewdoc.cpp
jonwil: fixed.

src/tools/w3dviewer/w3dviewdoc.cpp
jonwil: fixed.

src/tools/w3dviewer/w3dviewdoc.cpp
jonwil: fixed.

src/tools/w3dviewer/w3dviewdoc.h
jonwil: yes.

src/tools/w3dviewer/w3dviewdoc.cpp
jonwil: I don't know, ask Westwood.

src/tools/w3dviewer/w3dviewdoc.cpp
jonwil: fixed.

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1611 @@
...
+                CMenu *menu = GetMenu();
+
+                if (menu != nullptr) {
+                    MENUITEMINFO info;
+                    memset(&info, 0, sizeof(info));
+                    info.cbSize = sizeof(info);
+                    info.hSubMenu = GetSubMenu(LoadMenu(AfxGetResourceHandle(), MAKEINTRESOURCE(IDR_HIERARCHYMENU)), 0);
+                    info.fMask = MIIM_TYPE | MIIM_DATA | MIIM_SUBMENU;
+                    info.fType = MF_STRING;
+                    info.dwTypeData = "&Hierarchy";
```
OmniBlade: These assignments of string constants to dwTypeData is assigning const data type to a none const pointer type, at least in 32bit bit builds (not looked at 64bit). You could create some global constants or static constants such as char g_hierachyString[] = "&Hierarchy"; and then assign that to ensure defined behaviour.

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1611 @@
...
+                CMenu *menu = GetMenu();
+
+                if (menu != nullptr) {
+                    MENUITEMINFO info;
+                    memset(&info, 0, sizeof(info));
+                    info.cbSize = sizeof(info);
+                    info.hSubMenu = GetSubMenu(LoadMenu(AfxGetResourceHandle(), MAKEINTRESOURCE(IDR_AGGREGATEMENU)), 0);
+                    info.fMask = MIIM_TYPE | MIIM_DATA | MIIM_SUBMENU;
+                    info.fType = MF_STRING;
+                    info.dwTypeData = "&Aggregate";
```
OmniBlade: Another const string assigned to dwTypeData.

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1611 @@
...
+                CMenu *menu = GetMenu();
+
+                if (menu != nullptr) {
+                    MENUITEMINFO info;
+                    memset(&info, 0, sizeof(info));
+                    info.cbSize = sizeof(info);
+                    info.hSubMenu = GetSubMenu(LoadMenu(AfxGetResourceHandle(), MAKEINTRESOURCE(IDR_LODMENU)), 0);
+                    info.fMask = MIIM_TYPE | MIIM_DATA | MIIM_SUBMENU;
+                    info.fType = MF_STRING;
+                    info.dwTypeData = "&LOD";
```
OmniBlade: Another const string assigned to dwTypeData.

src/tools/w3dviewer/mainfrm.cpp
```diff
@@ -0,0 +1,1611 @@
...
+                CMenu *menu = GetMenu();
+
+                if (menu != nullptr) {
+                    MENUITEMINFO info;
+                    memset(&info, 0, sizeof(info));
+                    info.cbSize = sizeof(info);
+                    info.hSubMenu = GetSubMenu(LoadMenu(AfxGetResourceHandle(), MAKEINTRESOURCE(IDR_ANIMATIONMENU)), 0);
+                    info.fMask = MIIM_TYPE | MIIM_DATA | MIIM_SUBMENU;
+                    info.fType = MF_STRING;
+                    info.dwTypeData = "&Animation";
```
OmniBlade: Another const string assigned to dwTypeData.

OmniBlade (dismissed):
Looks like there are a couple of potentially UB string assignments which will trigger build errors when using /Zc:strictStrings to enforce constant strings not being assigned to pointers to none const data types.

src/tools/w3dviewer/mainfrm.cpp
jonwil: Ok so I looked at the MFC source code to see how it did this and it does an explicit cost_cast<LPSTR> to remove the constness. This then compiles fine with /Zc:strictStrings enabled (no warning) and since its in the official MFC source code from Microsoft is clearly an acceptable way to do it.

src/tools/w3dviewer/toolbar.cpp
tomsons26: ^

xezon:
There is an assert hit at
```
>    w3dview.exe!W3D::Render(SceneClass * scene, CameraClass * cam, bool clear, bool clearz, const Vector3 & color) Line 312    C++
     w3dview.exe!CGraphicView::Render(int update, unsigned int time) Line 590    C++
     w3dview.exe!CGraphicView::WindowProc(unsigned int message, unsigned int wParam, long lParam) Line 87    C++
```

when doing the following:

1. Load W3D Model with missing textures
2. Click on Model in hierarchy to see model with pink textures
3. Symlink or copy textures next to W3D files so model can find its textures now
4. Click on Tool Window

Edit:

Actually, the assert also triggers without doing the above. I had the w3d viewer idle for a few minutes doing something else on the computer. And once I clicked on the tool, the assert hit. An object was loaded in the viewport.

xezon (approved):
Good from my side. There is still an assert being hit, but that can be fixed in isolation.

PR #717, opened 2022-04-18
Various fixes and corrections for DX8Caps class.

tomsons26:
Fixes support for fog and multitexture
Corrects working string name to what Renegade shows
Clarifies that it's hardware Transform and Lighting
Adds missing default case for vendor
Corrects Check_Max_Texture_Support to match binary
Corrects Supported format functions, first entry of the array was set to false in original.

### DIFF
--- a/src/w3d/renderer/dx8caps.cpp
+++ b/src/w3d/renderer/dx8caps.cpp
@@ -21,7 +21,7 @@ using std::memcpy;
 using std::memset;
 
 #ifndef GAME_DLL
-StringClass DX8Caps::s_videoCardDetails;
+StringClass DX8Caps::s_capsWorkStrings;
 #else
 #include "hooker.h"
 #endif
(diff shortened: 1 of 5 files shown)


src/w3d/renderer/dx8caps.cpp
```diff
@@ -366,12 +379,14 @@ bool DX8Caps::Is_Valid_Display_Format(int w, int h, WW3DFormat format)
 void DX8Caps::Check_Texture_Format_Support(WW3DFormat format, const w3dcaps_t &caps)
 {
     if (format == WW3D_FORMAT_UNKNOWN) {
-        memset(m_supportTextureFormat, 0, sizeof(m_supportTextureFormat));
+        memset(m_supportTextureFormat, false, sizeof(m_supportTextureFormat));
```
xezon: Actually 0 is correct. memset expects integer.

If you really want to pass bool, then you can use `std::fill`.

xezon (approved):

PR #722, opened 2022-08-05
Import wdump clone from TT scripts

jonwil:


src/tools/wdump/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+add_executable(wdump WIN32)
```
duncanspumpkin, line 1: Shouldn't you add this executable to the list of artifacts to be uploaded during ci compile

src/w3d/renderer/w3d_file.h
```diff
@@ -542,8 +544,8 @@ struct W3dTriStruct
 
 struct W3dVertInfStruct
 {
-    uint16_t BoneIdx;
-    uint8_t Pad[6];
+    uint16_t BoneIdx[2];
+    uint16_t Weight[2];
```
xezon, line 548: It was 6 pad bytes before, now its 8 bytes, does this fit into game.dat binary constraints?

src/w3d/renderer/w3d_obsolete.h
```diff
@@ -0,0 +1,479 @@
+/**
+ * @file
+ *
+ * @author Jonathan Wilson
+ *
+ * @brief Obsolete W3D stuff
```
xezon, line 6: Why is this file called w3d_obsolete? If it is obsolete, why do we have it? Maybe there is a better name for it?

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+#include <commdlg.h>
+#include <unordered_map>
+
+HWND mainwnd;
+HWND treewnd;
+HWND listwnd;
+HMENU menu;
+HACCEL accel;
+int mainwidth;
+int mainheight;
```
xezon: We can 0 initialize all these for robustness.

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+            iter->second.function(cload, d);
+        }
+
+        cload.Close_Chunk();
+    }
+}
+
+#define FUNC(id) void dump##id(ChunkLoadClass &cload, ChunkData *data)
+
+FUNC(O_W3D_CHUNK_MATERIALS)
```
xezon: Any particular reason this is written with a macro? Can we just type it out instead? Would make it easier to debug if functions are clearly visible.

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+            StringClass str3;
+
+            for (unsigned int i = 0; i < cload.Cur_Chunk_Length(); i++) {
+                unsigned char n = chunkdata[i];
+                str3 += table[n >> 4];
+                str3 += table[n & 0xf];
+                str3 += ' ';
+            }
+
+            AddString(d, "Chunk Data", str3, "Unknown");
```
xezon, line 486: Leaking `chunkdata` here?

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+{
+    char c[256];
+    sprintf(c, "%d %d %d", value->I, value->J, value->K);
+    AddString(data, name, c, "IJK");
+}
+
+char *ReadChunkData(ChunkLoadClass &cload)
+{
+    if (!cload.Cur_Chunk_Length()) {
+        return "";
```
xezon: This function either returns an address to heap allocated memory or an address to static memory. The first must be freed by user, the second must not be freed. The user does not know what is returned, so should delete always. This can cause crash, if "" was returned and deleted.

Why return `""`? Can it not return `nullptr`? `""` should be allocated if `nullptr` is not possible.

I highly recommend to return a `std::unique_ptr<char[]>()` here and then remove all `delete[] chunk` from function callers. `std::unique_ptr` will automatically cleanup on scope exit. This way nothing can leak or crash.

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+    }
+};
+
+struct ChunkDumper
+{
+    const char *name;
+    void (*function)(ChunkLoadClass &cload, ChunkData *data);
+};
+
+std::unordered_map<int, ChunkDumper> chunks;
```
xezon: Please name this `g_chunks`. Otherwise it is really confusing in functions as to what `chunks` refers to, especially here in review. Additionally, please give that `int` a typedef. It is not clear what it refers to. Or clarify in name, e.g. `g_chunksIdMap`, if it is an ID.

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+{
+    char *chunkdata = ReadChunkData(cload);
+    W3dEmitterPropertyStruct *props = (W3dEmitterPropertyStruct *)chunkdata;
+    AddInt32(data, "ColorKeyframes", props->ColorKeyframes);
+    AddInt32(data, "OpacityKeyframes", props->OpacityKeyframes);
+    AddInt32(data, "SizeKeyframes", props->SizeKeyframes);
+    AddRGBA(data, "ColorRandom", &props->ColorRandom);
+    AddFloat(data, "OpacityRandom", props->OpacityRandom);
+    AddFloat(data, "SizeRandom", props->SizeRandom);
+    W3dEmitterColorKeyframeStruct *color = (W3dEmitterColorKeyframeStruct *)(chunkdata + sizeof(W3dEmitterPropertyStruct));
```
xezon, line 1143: This is mind boggling. How do you know this?

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+        AddString(data, "Attributes", "W3D_MESH_FLAG_COLLISION_TYPE_USER1", "flag");
+    }
+
+    if (header->Attributes & W3D_MESH_FLAG_COLLISION_TYPE_USER2) {
+        AddString(data, "Attributes", "W3D_MESH_FLAG_COLLISION_TYPE_USER2", "flag");
+    }
+
+    if (header->Attributes & 0x00000800) {
+        StringClass str;
+        str.Format("W3D_CHUNK_MESH_HEADER3 Unknown Attribute 0x00000800");
```
xezon, line 1677: What does this mean?

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+    AddString(data, "Texture Name:", chunkdata, "string");
+    delete[] chunkdata;
+}
+
+FUNC(W3D_CHUNK_TEXTURE_REPLACER_INFO)
+{
+    char *chunkdata = ReadChunkData(cload);
+    W3dTextureReplacerHeaderStruct *header = (W3dTextureReplacerHeaderStruct *)chunkdata;
+    AddInt32(data, "ReplacedTexturesCount", header->ReplacedTexturesCount);
+    W3dTextureReplacerStruct *replacer = (W3dTextureReplacerStruct *)(chunkdata + sizeof(W3dTextureReplacerHeaderStruct));
```
xezon, line 2225: How do you know this?

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+            break;
+        case W3D_MESH_FLAG_GEOMETRY_TYPE_CAMERA_Z_ORIENTED:
+            AddString(data, "Attributes", "W3D_MESH_FLAG_GEOMETRY_TYPE_CAMERA_Z_ORIENTED", "flag");
+            break;
+        default: {
+            StringClass str;
+            str.Format("W3D_CHUNK_MESH_HEADER3 Unknown Mesh Type %x", type);
+            data->unknowndata.Add(str);
+            AddString(data, "Attributes", "Unknown", "string");
+        } break;
```
xezon: Can write `break` in scope like the others.

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+            AddString(data, "Flags", "W3D_ELINE_UNIFORM_LENGTH_TEXTURE_MAP", "flag");
+            break;
+        case W3D_ELINE_TILED_TEXTURE_MAP:
+            AddString(data, "Flags", "W3D_ELINE_TILED_TEXTURE_MAP", "flag");
+            break;
+        default: {
+            StringClass str;
+            str.Format("W3D_CHUNK_EMITTER_LINE_PROPERTIES Unknown Emitter Mapping Mode %x", mapmode);
+            data->unknowndata.Add(str);
+            AddString(data, "Flags", "Unknown", "string");
```
xezon, line 2720: Minor: Can add `break` here.

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+                AddFloat(data, "OuterScale.X", RingStruct.OuterScale.x);
+                AddFloat(data, "OuterScale.Y", RingStruct.OuterScale.y);
+                AddFloat(data, "InnerExtent.X", RingStruct.InnerExtent.x);
+                AddFloat(data, "InnerExtent.Y", RingStruct.InnerExtent.y);
+                AddFloat(data, "OuterExtent.X", RingStruct.OuterExtent.x);
+                AddFloat(data, "OuterExtent.Y", RingStruct.OuterExtent.y);
+                AddString(data, "TextureName", RingStruct.TextureName, "string");
+                AddShader(data, "Shader", &RingStruct.Shader);
+                AddInt32(data, "TextureTiling", RingStruct.TextureTiling);
+            } break;
```
xezon: Can write `break` inside scope like the others.

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+        cload.Close_Chunk();
+    }
+}
+
+void DoVector3Channel(ChunkLoadClass &cload, ChunkData *data, const char *name)
+{
+    int i = 0;
+
+    for (; cload.Open_Chunk(); cload.Close_Chunk()) {
+        if (cload.Cur_Chunk_ID() == 51709961) {
```
xezon: What is this magic number here?

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+    delete[] chunkdata;
+}
+
+const char *Types[] = { "Texture", "Float", "Vector2", "Vector3", "Vector4", "Int", "Bool" };
+
+FUNC(W3D_CHUNK_FX_SHADER_CONSTANT)
+{
+    char *chunkdata = ReadChunkData(cload);
+    uint32_t type = *(uint32_t *)chunkdata;
+    uint32_t strlen = *(uint32_t *)(chunkdata + 4);
```
xezon: Could you give this a name that does not look like C `strlen` function?

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+
+        uint32_t *values = (uint32_t *)(chunkdata + pos);
+
+        for (int i = 0; i < datalen; i++) {
+            StringClass str;
+            str.Format("Data[%d]", i);
+            AddInt32(data, str, values[i]);
+        }
+    } else {
+        float *scale = (float *)(chunkdata + sizeof(W3dCompressedMotionChannelStruct));
```
xezon, line 3605: How do you know this?

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+    StringClass str = data->name;
+    HTREEITEM newitem = TreeViewInsertItem(treewnd, str, item, TVI_LAST);
+    TreeViewSetItem(treewnd, newitem, (LPARAM)data);
+
+    for (int i = 0; i < data->subchunks.Count(); i++) {
+        AddItems(data->subchunks[i], newitem);
+    }
+}
+
+ChunkData *master = nullptr;
```
xezon: `g_mainChunk` perhaps?

xezon (changes requested):
The ReadChunkData function certainly needs fixing.

src/w3d/renderer/w3d_file.h
duncanspumpkin: It is correct because its gone from `2 + 6*1` to `2*2 + 2*2`

xezon: Ah right. Somehow did not see the BoneIdx

src/w3d/renderer/w3d_obsolete.h
tomsons26: it's old no longer used at some point chunks in W3D, we still need to know about them, this file existed in actual source code, the public w3d file header refs it

src/tools/wdump/wdump.cpp
tomsons26: We don't know what its for, so it would be good if we could spot such attribs like that

src/tools/wdump/wdump.cpp
tomsons26: i'd guess CHUNKID_PARENT of PrimitiveAnimationChannelClass

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4138 @@
...
+        AddFloat(data, "Scale", *scale);
+        float *initial = (float *)(chunkdata + sizeof(W3dCompressedMotionChannelStruct) + 4);
+
+        for (int i = 0; i < channel->VectorLen; i++) {
+            StringClass str;
+            str.Format("Initial[%d]", i);
+            AddFloat(data, str, initial[i]);
+        }
+
+        int count = (cload.Cur_Chunk_Length() - sizeof(W3dCompressedMotionChannelStruct) - 4 - 4 * channel->VectorLen) / 4;
```
tomsons26, line 3615: () prob here

src/tools/wdump/CMakeLists.txt
jonwil: Fixed.

src/tools/wdump/wdump.cpp
jonwil: fixed.

src/tools/wdump/wdump.cpp
jonwil: Fixed.

src/tools/wdump/wdump.cpp
jonwil: fixed.

src/tools/wdump/wdump.cpp
jonwil: Fixed to allocate the "".

src/tools/wdump/wdump.cpp
jonwil: fixed.

src/tools/wdump/wdump.cpp
jonwil: This is exactly what the original code in the WW binary does.

src/tools/wdump/wdump.cpp
jonwil: This is exactly what the original code in the WW binary does.

src/tools/wdump/wdump.cpp
jonwil: fixed.

src/tools/wdump/wdump.cpp
jonwil: fixed.

src/tools/wdump/wdump.cpp
jonwil: fixed.

src/tools/wdump/wdump.cpp
jonwil: Yes it is.

src/tools/wdump/wdump.cpp
jonwil: fixed.

src/tools/wdump/wdump.cpp
jonwil: This is exactly what the original code in the WW binary does.

src/tools/wdump/wdump.cpp
jonwil: fixed.

src/tools/wdump/wdump.cpp
jonwil: Not sure what you mean here, it looks good to me.

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4142 @@
...
+#include <commdlg.h>
+#include <unordered_map>
+
+HWND mainwnd = nullptr;
+HWND treewnd = nullptr;
+HWND listwnd = nullptr;
+HMENU menu = nullptr;
+HACCEL accel = nullptr;
+int mainwidth = 0;
+int mainheight = 0;
```
xezon: I suggest name these globals with `g_` prefix too for visibility.

src/tools/wdump/wdump.cpp
```diff
@@ -0,0 +1,4142 @@
...
+    }
+
+    delete[] chunkdata;
+}
+
+#define CHUNK(id) \
+    { \
+        ChunkDumper c; \
+        c.name = #id; \
+        c.function = dump##id; \
```
xezon: I suggest build function name as `Dump_##id` here, then the above functions would look a bit more gracious as
`Dump_W3D_CHUNK_COMPRESSED_ANIMATION_MOTION_CHANNEL`
instead of
`dumpW3D_CHUNK_COMPRESSED_ANIMATION_MOTION_CHANNEL`

src/tools/wdump/wdump.cpp
xezon: Can this be written out with an enum value, or at least commented?

src/tools/wdump/wdump.cpp
jonwil: fixed

src/tools/wdump/wdump.cpp
jonwil: fixed.

src/tools/wdump/wdump.cpp
jonwil: fixed.

xezon (approved):
Looks ok to me.

PR #723, opened 2022-09-05
Refactors how wxwidgets dependency is handled.

OmniBlade:
wxWidgets is now fetched and built by cmake when not found.


jonwil (approved):
Works and builds on all platforms. Approving.

PR #726, opened 2022-09-08
Cleanup RenderObjClass properly

xezon:
Untested. ~~Not applicable for `GAME_DLL` I think, because touches vtable.~~

The idea is call `Remove` function before deletion, to make sure that the deleted object is being disconnected always. If it is not disconnected, then dangling pointer can cause game crash. Essentially this change makes code more robust.

### DIFF
--- a/src/w3d/renderer/rendobj.cpp
+++ b/src/w3d/renderer/rendobj.cpp
@@ -107,6 +107,16 @@ RenderObjClass &RenderObjClass::RenderObjClass::operator=(const RenderObjClass &
     return *this;
 }
 
+RenderObjClass::~RenderObjClass()
+{
+    if (m_unknown) {
+        delete m_unknown;
+    }
+
+    // BUGFIX Cleanup always to avoid dangling pointers.
+    RenderObjClass::Remove();
+}
+
 float RenderObjClass::Get_Screen_Size(CameraClass &camera)
 {
     Vector3 cam = camera.Get_Position();
--- a/src/w3d/renderer/rendobj.h
+++ b/src/w3d/renderer/rendobj.h
@@ -123,12 +123,7 @@ public:
     RenderObjClass(const RenderObjClass &src);
     RenderObjClass &operator=(const RenderObjClass &that);
 
-    virtual ~RenderObjClass()
-    {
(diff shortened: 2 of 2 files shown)


xezon:
Updated with implementation that does not touch vtable.

src/w3d/renderer/rendobj.cpp
```diff
@@ -107,6 +107,16 @@ RenderObjClass &RenderObjClass::RenderObjClass::operator=(const RenderObjClass &
...
 }
 
+RenderObjClass::~RenderObjClass()
+{
+    if (m_unknown) {
+        delete m_unknown;
+    }
+
+    // Thyme specific. Cleanup always to avoid dangling pointers.
+    Remove_Internal();
```
tomsons26: do RenderObjClass::Remove instead and comment this as a bugfix

tomsons26 (changes requested):

src/w3d/renderer/rendobj.cpp
xezon: I advise against this. Calling virtual function in destructor can lead to unexpected behaviour. It will make code less intuitive to understand, as you would need to exactly understand how destruction in virtual context works to anticipate the exact flow of execution that results from it. It is much safer to call non virtual function instead.

See:
https://www.artima.com/articles/never-call-virtual-functions-during-construction-or-destruction

tomsons26: prefixing it with the class name calls it directly

xezon: Ah. Meaning `RenderObjClass::Remove` will make it a CALL Fixed Address, yes?

tomsons26: yes

xezon: Ok we can do that. I will get back to it when the compile errors are fixed and I can test this change properly.

xezon: Fixed and tested. Loaded big test map, killed a bunch of units with SCUD Storm, all good.

tomsons26 (approved):

PR #729, opened 2022-09-16
Corrects various DirectInput return checks, assigns proper key enum to Keyboard class.

### LINKED ISSUE #123: Investigate DirectInputKeyboard DirectX interface returns.
If (m_inputDevice->Acquire() < 0) causes the game to log "Failed to acquire device" when the game is not in focus when starting, if (m_inputDevice->Acquire() == 0) logs that it inited successfully, keyboard works in both cases.

We should investigate what are the proper return checks for calls to the DirectX interfaces in this class.

tomsons26:
fixes #123

### DIFF
--- a/src/game/client/input/keyboard.cpp
+++ b/src/game/client/input/keyboard.cpp
@@ -1,4 +1,4 @@
-/**
+﻿/**
  * @file
  *
  * @author OmniBlade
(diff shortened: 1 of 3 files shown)


jonwil (approved):
Looks good to me, approved.

PR #733, opened 2022-09-17
Fixes WB compilation.

tomsons26:


src/w3d/lib/sharebuf.h
```diff
@@ -66,3 +69,9 @@ template<class T> class ShareBufferClass : public W3DMPO, public RefCountClass
     T *m_array;
     int m_count;
 };
+
+#ifndef BUILD_EDITOR
+#define NEW_SHARE_BUFFER(T, V) (new ShareBufferClass<T>(V))
+#else
+#define NEW_SHARE_BUFFER(T, V, N) (new ShareBufferClass<T>(V, N))
+#endif
```
xezon, line 79: I suggest to create small helper function instead.

```c++
template <typename T>
ShareBufferClass<T>* New_Share_Buffer_Class(int count, const char* name)
{
#ifdef BUILD_EDITOR
    return new ShareBufferClass<T>(count, name);
#else
    return new ShareBufferClass<T>(count);
#endif
}
```

Same approach for the other new macros. This will be inlined and `name` will be omitted in Release because not used.

xezon: Further I suggest to use `#ifdef` instead of `#ifndef`. It is easier to the eyes to use non negated conditions. Suggested code updated accordingly.

src/w3d/lib/sharebuf.h
```diff
@@ -66,3 +69,12 @@ template<class T> class ShareBufferClass : public W3DMPO, public RefCountClass
...
 };
+
+template<class T> inline ShareBufferClass<T> *New_Share_Buffer(int count, const char *name = nullptr)
+{
+#ifdef BUILD_EDITOR
+    return new ShareBufferClass<T>(count, name);
+#else
+    return new ShareBufferClass<T>(count);
+#endif
+}
```
xezon: Can add EOL here.

xezon (approved):
Looks good.

PR #734, opened 2022-09-17
Fixes some wrong and missing code.

tomsons26:


src/game/common/system/gamememory.cpp
```diff
@@ -41,12 +41,27 @@ void *New_New(size_t bytes)
     return g_dynamicMemoryAllocator->Allocate_Bytes(bytes);
 }
 
+void *New_New_Dbg(size_t bytes, char const *file, int line)
+{
+    ++(Make_Global<int>(PICK_ADDRESS(0x00A29B9C, 0x00E1B368)));
```
xezon, line 46: May I suggest to separate these into
```c++
auto &var_name = (Make_Global<int>(PICK_ADDRESS(0x00A29B9C, 0x00E1B368)))
++var_name;
```
This way it will be possible to inspect variable easily in debugger.

src/w3d/renderer/dx8wrapper.cpp
```diff
@@ -2000,17 +2000,25 @@ void DX8Wrapper::Set_Render_Target(w3dsurface_t render_target, bool use_default_
             DX8CALL(SetRenderTarget(s_defaultRenderTarget, s_defaultDepthBuffer));
             s_defaultRenderTarget->Release();
             s_defaultRenderTarget = nullptr;
-            if (s_defaultDepthBuffer) {
+
+            if (s_defaultDepthBuffer != nullptr) {
```
xezon, line 2004: Is this meant to be inside scope of `if (s_defaultRenderTarget != nullptr)` condition ?

src/game/common/system/mempool.cpp
```diff
@@ -180,6 +182,8 @@ int MemoryPool::Release_Empties()
 
 void MemoryPool::Reset()
 {
+    ScopedCriticalSectionClass scs(g_memoryPoolCriticalSection);
```
xezon, line 185: Are all locks now placed in this class? Asking because the lock coverage looks wild...

xezon (approved):
Looks ok from cpp side

PR #739, opened 2022-09-24
Sets correct memory functions adds missing globals for WB.

tomsons26:
Not sure how to go about the crit section stuff, maybe it should be its own function or something
Or we need to implement a main before merging this

### DIFF
--- a/src/hooker/setupglobals_zh.cpp
+++ b/src/hooker/setupglobals_zh.cpp
@@ -51,7 +51,7 @@ class TerrainRoadCollection;
 TerrainRoadCollection *&g_theTerrainRoads = Make_Global<TerrainRoadCollection *>(PICK_ADDRESS(0x00A2BE4C, 0x00E23A34));
 
 // view.cpp
-uint32_t &View::s_idNext = Make_Global<uint32_t>(PICK_ADDRESS(0x009D4178, 0));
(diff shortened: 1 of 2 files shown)


src/hooker/setupglobals_zh.cpp
```diff
@@ -580,19 +580,20 @@ W3DProjectedShadowManager *&g_theW3DProjectedShadowManager =
 class FrustumClass;
 FrustumClass *&g_shadowCameraFrustum = Make_Global<FrustumClass *>(PICK_ADDRESS(0x00A3AE3C, 0x00E1A604));
 #ifdef BUILD_WITH_D3D8
-IDirect3DVertexBuffer8 *&g_shadowDecalVertexBufferD3D = Make_Global<IDirect3DVertexBuffer8 *>(0x00A3ACE8);
-IDirect3DIndexBuffer8 *&g_shadowDecalIndexBufferD3D = Make_Global<IDirect3DIndexBuffer8 *>(0x00A3ACEC);
+    IDirect3DVertexBuffer8 *&g_shadowDecalVertexBufferD3D =
```
xezon: Is this meant to be indented?

tomsons26: no, clangformat produced that, but ran it again and this time it didn't...

xezon (approved):

PR #743, opened 2022-09-27
Reverts test for success in dinputkeybd.cpp.

OmniBlade:
Reverts a test that checks if Acquire failed and closed the keyboard if it did. Seems this breaks the keyboard input as it doesn't seem to matter if it fails here.

### DIFF
--- a/src/platform/directx/dinputkeybd.cpp
+++ b/src/platform/directx/dinputkeybd.cpp
@@ -127,12 +127,7 @@ void DirectInputKeyboard::Open_Keyboard()
         return;
     }
 
-    if (FAILED(m_inputDevice->Acquire())) {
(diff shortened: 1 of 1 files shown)


jonwil (approved):
Approved.

src/platform/directx/dinputkeybd.cpp
```diff
@@ -127,12 +127,7 @@ void DirectInputKeyboard::Open_Keyboard()
         return;
     }
 
-    if (FAILED(m_inputDevice->Acquire())) {
-        captainslog_error("Failed to acquire device.");
```
xezon, line 131: Maybe keep the error log here, but do not close and return?

PR #749, opened 2022-09-30
W3DSupplyDraw

jonwil:


src/game/client/draw/w3dsupplydraw.cpp
```diff
@@ -0,0 +1,103 @@
...
+void W3DSupplyDraw::Load_Post_Process()
+{
+    W3DModelDraw::Load_Post_Process();
+}
+
+ModuleData *W3DSupplyDraw::Friend_New_Module_Data(INI *ini)
+{
+    W3DSupplyDrawModuleData *data = new W3DSupplyDrawModuleData();
+
+    if (ini) {
```
xezon: != nullptr

src/game/client/draw/w3dsupplydraw.cpp
```diff
@@ -0,0 +1,103 @@
...
+    return nk;
+}
+
+void W3DSupplyDraw::Update_Draw_Module_Supply_Status(int max, int current)
+{
+    W3DModelDraw::Update_Draw_Module_Supply_Status(max, current);
+    Utf8String prefix = Get_W3D_Supply_Draw_Module_Data()->m_supplyBonePrefix;
+
+    if (m_totalSupplyBoneCount == -1) {
+        m_totalSupplyBoneCount = Get_Drawable()->Get_Pristine_Bone_Positions(prefix, 1, nullptr, nullptr, 0x7FFFFFFF);
```
xezon: Can use INT_MAX instead of 0x7FFFFFFF

src/game/client/draw/w3dsupplydraw.cpp
```diff
@@ -0,0 +1,103 @@
...
+    if (newcount != m_currentSupplyBoneCount) {
+        int minbone = std::min(m_currentSupplyBoneCount, newcount);
+        int maxbone = std::max(m_currentSupplyBoneCount, newcount);
+        bool hide = newcount < m_currentSupplyBoneCount;
+        int bonecount = minbone + 1;
+        std::vector<ModelConditionInfo::HideShowSubObjInfo> subobjs;
+
+        while (bonecount <= maxbone) {
+            char name[16];
+            sprintf(name, "%s%02d", prefix.Str(), bonecount);
```
xezon: This is bug. It means the logic will not just break if the bone name is longer than 14 or so characters, but that the buffer will also be overrun and cause memory corruption.

Using `sprintf_s` here would only fix it half. I suggest using `Utf8String` for `name` with its `Format` function. That way it will be robust. Mark as BUGFIX.

Bugfix can be done in separate change, to preserve original history if so necessary.

src/game/client/draw/w3dsupplydraw.cpp
```diff
@@ -0,0 +1,103 @@
...
+NameKeyType W3DSupplyDraw::Get_Module_Name_Key() const
+{
+    static const NameKeyType nk = g_theNameKeyGenerator->Name_To_Key("W3DSupplyDraw");
+    return nk;
+}
+
+void W3DSupplyDraw::Update_Draw_Module_Supply_Status(int max, int current)
+{
+    W3DModelDraw::Update_Draw_Module_Supply_Status(max, current);
+    Utf8String prefix = Get_W3D_Supply_Draw_Module_Data()->m_supplyBonePrefix;
```
xezon: No need to "copy" here. `const Utf8String &prefix` will suffice.

xezon (commented):
Tested on Big Supply Stash and Small Supply Piles. Both work.

src/game/client/draw/w3dsupplydraw.cpp
jonwil: fixed.

src/game/client/draw/w3dsupplydraw.cpp
jonwil: fixed.

src/game/client/draw/w3dsupplydraw.cpp
jonwil: sprintf_s makes sense but bone names are fixed by the w3d format at 16 characters therefore using a 16 character array is correct here.

src/game/client/draw/w3dsupplydraw.cpp
jonwil: fixed.

src/game/client/draw/w3dsupplydraw.cpp
xezon: But we append digits, max 2, so it can be longer.

xezon:
```
{ "SupplyBonePrefix", &INI::Parse_AsciiString, nullptr, offsetof(W3DSupplyDrawModuleData, m_supplyBonePrefix) },
```

From my understanding, `m_supplyBonePrefix` can be any length at this point. There is nothing that caps the input.

jonwil: If someone puts a prefix in the ini that's too long, things will break regardless of the size of the array because it would produce bone names that are longer than the valid length of a bone in a w3d file so it's impossible for those to be valid bones. We should not be setting subobj.sub_obj_name to a name that is longer than the valid length of a bone in a w3d file.

xezon: Ok that is fair. Then we should add an assert in the field parse for `m_supplyBonePrefix <= 13` (16 - null terminator - 2 digits) to make it obvious that things break then.

xezon (approved):
Good from my side. Check with tom if he wants to add anything.

tomsons26 (approved):

PR #752, opened 2022-10-03
Fix diagonal movement speed bug

xezon:
This change fixes the diagonal movement speed bug.

Reference https://github.com/TheSuperHackers/GeneralsGameCode/issues/123

### Demo

https://user-images.githubusercontent.com/4720891/193651852-ba618719-c85c-4b9b-8ad4-7058da911af4.mp4

### DIFF
--- a/src/game/logic/object/update/physicsupdate.cpp
+++ b/src/game/logic/object/update/physicsupdate.cpp
@@ -662,33 +662,43 @@ float PhysicsBehavior::Get_Velocity_Magnitude() const
     return m_velMag;
 }
 
-// TODO investigate doesn't account for diagonal movement
 float PhysicsBehavior::Get_Forward_Speed_2D() const
 {
     const Coord3D *dir = Get_Object()->Get_Unit_Dir_Vector2D();
-    float x = m_vel.x * dir->x;
-    float y = m_vel.y * dir->y;
-    float xy = x + y;
-    float len = GameMath::Sqrt(x * x + y * y);
 
-    if (xy < 0.0f) {
+    // #BUGFIX square root the velocities separately to fix issue where units move faster diagonally.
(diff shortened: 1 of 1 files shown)


xezon:
I suggest to merge this fix later when we no longer need game state to be consistent with original. For now it will be good to have Thyme work similar to Original, so we can use Original Replays to backtest and find issues. This change would break this ability.

xezon:
Fix is done, but converting to Draft for the time being.

xezon:
Functions were moved from header to cpp file. Change updated accordingly.

src/game/logic/object/update/physicsupdate.cpp
```diff
@@ -660,33 +660,43 @@ float PhysicsBehavior::Get_Velocity_Magnitude() const
...
-    float xy = x + y;
-    float len = GameMath::Sqrt(x * x + y * y);
 
-    if (xy < 0.0f) {
+    // #BUGFIX square root the velocities separately to fix issue where units move faster diagonally.
+    const float vel_len = GameMath::Sqrt(m_vel.x * m_vel.x + m_vel.y * m_vel.y);
+    const float dir_len = GameMath::Sqrt(dir->x * dir->x + dir->y * dir->y);
+    float len = vel_len * dir_len;
+    // Scale len by (1 + sqrt(2)) / 2 to adjust to the average of the former min/max movement speed.
+    len /= 1.20710678f;
```
commy2, line 674: May as well multiply by the inverse (may be faster).
Would need to think about if this factor is the correct one. It should be scaled by:

    speed *= average_before/average_after

What is average_before though?

xezon: Multiplication indeed is favourable here. That would be 1/1.20... = 0.82842712 then.

PR #768, opened 2022-10-16
Fix compile warning in TurretAI::Update_Turret_AI

xezon:
This change fixes compile warning

src\game\logic\ai\turretai.cpp(868,22): warning C4018: '<': signed/unsigned mismatch

### DIFF
--- a/src/game/logic/ai/turretai.cpp
+++ b/src/game/logic/ai/turretai.cpp
@@ -865,7 +865,7 @@ UpdateSleepTime TurretAI::Update_Turret_AI()
         }
 
         if (type > 0) {
-            if (type < time) {
+            if (static_cast<UpdateSleepTime>(type) < time) {
                 time = static_cast<UpdateSleepTime>(type);
             }
         } else {


tomsons26:
not sure actually what to do here
```
time = 0x3FFFFFFF;
....
        if ( (signed int)type > 0 )
        {
            if ( (signed int)type < 0x3FFFFFFF ) //this actually is checking time in asm as its currently written
            {
                time = type;
            }
        }
        else
        {
            time = 1;
        }
```
as pseudo it shows that, maybe UpdateSleepTime should be signed, maybe we should just use casts

xezon:
`type` is larger than zero, so casting it to unsigned is logically correct. Casting `time` to signed would not be correct, because it would break if UPDATE_SLEEP_TIME_MAX would change to be above 7FFFFFFF.

xezon:
Fixed

jonwil:
This PR is obsolete, warning no longer occurs. Closing.

PR #770, opened 2022-10-17
Fix animation jumping back one frame when cross ConditionState using MAINTAIN_FRAME_ACROSS_STATES

xezon:
This is a possible fix for
https://github.com/TheSuperHackers/GeneralsGameCode/issues/157

### DIFF
--- a/src/game/client/draw/w3dmodeldraw.cpp
+++ b/src/game/client/draw/w3dmodeldraw.cpp
@@ -1501,6 +1501,13 @@ float W3DModelDraw::Get_Current_Anim_Fraction() const
         return 0.0f;
     }
 
+    // #BUGFIX Fix animation jumping back one frame when crossing ConditionState using MAINTAIN_FRAME_ACROSS_STATES.
+    // Loop frame to beginning of animation if supported.
+    frame += 2.0f;
+    if (Test_Animation_Flag(m_curState->m_flags, RESTART_ANIM_WHEN_COMPLETE) && frame > frames) {
+        frame -= frames;
+    }
+
     if (frames > frame) {
         return frame / (frames - 1.0f);
     }


src/game/client/draw/w3dmodeldraw.cpp
```diff
@@ -1501,6 +1501,13 @@ float W3DModelDraw::Get_Current_Anim_Fraction() const
         return 0.0f;
     }
 
+    // #BUGFIX Fix animation jumping back one frame when crossing ConditionState using MAINTAIN_FRAME_ACROSS_STATES.
+    // Loop frame to beginning of animation if supported.
+    frame += 2.0f;
+    if (Test_Animation_Flag(m_curState->m_flags, RESTART_ANIM_WHEN_COMPLETE) && frame > frames) {
+        frame -= frames;
```
xezon, line 1508: This is done to gracefully support edge case where it should get to begin of animation if it crossed the end of it.

src/game/client/draw/w3dmodeldraw.cpp
```diff
@@ -1501,6 +1501,13 @@ float W3DModelDraw::Get_Current_Anim_Fraction() const
         return 0.0f;
     }
 
+    // #BUGFIX Fix animation jumping back one frame when crossing ConditionState using MAINTAIN_FRAME_ACROSS_STATES.
+    // Loop frame to beginning of animation if supported.
+    frame += 2.0f;
```
xezon, line 1506: Here I would have expected to add just one frame to advance the animation after we copy it from the previous animation state. But adding just one will duplicate the animation frame when crossing the ConditionState. It is unclear why it goes back one frame. Could be another bug somewhere else. However, if we are unable to observe other issues, then this fix is sufficient, otherwise just add one here if we find another place to avoid the deduction. Perhaps it is supposed to work like this and +2 frames is the correct approach.

As of right now this fix works for the observed issue. It would be good to test more scenarios to gain more confidence.

PR #771, opened 2022-10-17
Refactor AnimationStateFlag handling

xezon:
A few minor improvements to w3dmodeldraw.

### DIFF
--- a/src/game/client/draw/w3dmodeldraw.cpp
+++ b/src/game/client/draw/w3dmodeldraw.cpp
@@ -144,7 +144,7 @@ void ModelConditionInfo::Validate_Stuff(RenderObjClass *robj, float scale, std::
     Validate_Weapon_Barrel_Info();
 }
 
-bool Test_Animation_Flag(int flags, char flag)
+bool Test_Animation_Flag(int flags, AnimationStateFlag flag)
 {
     return ((1 << flag) & flags) != 0;
 }
@@ -1378,7 +1378,7 @@ void W3DModelDraw::Adjust_Transform_Mtx(Matrix3D &transform) const
         transform.Adjust_Z_Translation(v.Z);
     }
 
-    if ((m_curState->m_flags & 1 << ADJUST_HEIGHT_BY_CONSTRUCTION_PERCENT) != 0) {
+    if (Test_Animation_Flag(m_curState->m_flags, ADJUST_HEIGHT_BY_CONSTRUCTION_PERCENT)) {
         const Object *object = Get_Drawable()->Get_Object();
 
         if (object != nullptr) {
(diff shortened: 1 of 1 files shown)


src/game/client/draw/w3dmodeldraw.cpp
```diff
@@ -144,7 +144,7 @@ void ModelConditionInfo::Validate_Stuff(RenderObjClass *robj, float scale, std::
     Validate_Weapon_Barrel_Info();
 }
 
-bool Test_Animation_Flag(int flags, char flag)
+bool Test_Animation_Flag(int flags, AnimationStateFlag flag)
```
xezon, line 147: This is more explicit. No need for it to be a `char` when we pass `AnimationStateFlag` to it.

src/game/client/draw/w3dmodeldraw.cpp
```diff
@@ -1472,14 +1472,21 @@ const ModelConditionInfo *W3DModelDraw::Find_Transition_For_Sig(uint64_t sig) co
...
 bool Maintain_Frame_Across_States(int flags)
 {
-    return (flags & 0x3A0) != 0;
+    return (flags & g_maintain_frame_across_states_flags) != 0;
 }
 
 bool Maintain_Frame_Across_States_2(int flags, int flags2)
 {
-    return (flags2 & 0x3A0 & flags & 0x3A0) != 0;
+    return (flags2 & flags & g_maintain_frame_across_states_flags) != 0;
```
xezon, line 1489: No need to mask it twice here. Once will do.

src/game/client/draw/w3dmodeldraw.cpp
```diff
@@ -1502,7 +1509,7 @@ float W3DModelDraw::Get_Current_Anim_Fraction() const
     }
 
     if (frames > frame) {
-        return frame / (frames - 1.0f);
+        return frame / (frames - 1);
```
xezon: `frames` is integer, so can deduct integer.

xezon (commented):
I tested and animations looked as usual.

src/game/client/draw/w3dmodeldraw.cpp
xezon: Reverted.

tomsons26:
change if ((m_curState->m_flags & 1 << ADJUST_HEIGHT_BY_CONSTRUCTION_PERCENT) != 0) { to Test_Animation_Flag as well while you are at it
and check if there's any other direct uses of it and do the same, imo it shouldn't be directly checked like this at all

xezon:
Fixed. I did not see any others.

tomsons26 (approved):

PR #775, opened 2022-10-28
Remove public mutable string functions in Utf8String and Utf16String

xezon:
Make mutable string functions in Utf8String and Utf16String private.

Fix discrepancies with code game.

### DIFF
--- a/src/game/common/system/archivefile.cpp
+++ b/src/game/common/system/archivefile.cpp
@@ -154,8 +154,8 @@ bool ArchiveFile::Search_String_Matches(Utf8String string, Utf8String search)
         return false;
     }
 
-    const char *cstring = string.Peek();
-    const char *csearch = search.Peek();
+    const char *cstring = string.Str();
+    const char *csearch = search.Str();
 
     // ? is wildcard for a single character.
     // * is wildcard for a run of characters.
--- a/src/game/common/system/datachunk.cpp
+++ b/src/game/common/system/datachunk.cpp
@@ -271,9 +271,10 @@ Utf8String DataChunkInput::Read_AsciiString()
 
     captainslog_dbgassert(m_chunkStack->data_left >= size, "Read past end of chunk reading Utf8String string.");
 
-    m_file->Read(string.Get_Buffer_For_Read(size), size);
+    char *buf = string.Get_Buffer_For_Read(size);
+    m_file->Read(buf, size);
     Decrement_Data_Left(size);
-    string.Peek()[size] = '\0';
+    buf[size] = '\0';
 
     return string;
 }
@@ -305,7 +306,6 @@ Utf16String DataChunkInput::Read_UnicodeString()
     }
 
     Decrement_Data_Left(sizeof(ch) * size);
-    string.Peek()[size] = L'\0';
 
     return string;
 }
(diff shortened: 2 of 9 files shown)


src/game/common/system/datachunk.cpp
```diff
@@ -271,9 +271,10 @@ Utf8String DataChunkInput::Read_AsciiString()
 
     captainslog_dbgassert(m_chunkStack->data_left >= size, "Read past end of chunk reading Utf8String string.");
 
-    m_file->Read(string.Get_Buffer_For_Read(size), size);
+    char* buf = string.Get_Buffer_For_Read(size);
+    m_file->Read(buf, size);
     Decrement_Data_Left(size);
-    string.Peek()[size] = '\0';
+    buf[size] = '\0';
```
xezon, line 277: As seen in World Builder.

src/game/common/system/datachunktoc.cpp
```diff
@@ -111,8 +111,9 @@ void DataChunkTableOfContents::Read(ChunkInputStream &stream)
 
             // If we have a name length, read it and null terminate it.
             if (name_len > 0) {
-                stream.Read(map->m_name.Get_Buffer_For_Read(name_len), name_len);
-                map->m_name.Peek()[name_len] = '\0';
+                char* buf = map->m_name.Get_Buffer_For_Read(name_len);
+                stream.Read(buf, name_len);
+                buf[name_len] = '\0';
```
xezon, line 116: As seen in World Builder.

src/game/network/lanapi.cpp
```diff
@@ -561,7 +561,7 @@ void LANAPI::Send_Message(LANMessage *msg, uint32_t addr)
  *
  * 0x005E1D20
  */
-uint32_t LANAPI::Resolve_IP(Utf8String addr)
+uint32_t LANAPI::Resolve_IP(const Utf8String &addr)
```
xezon, line 564: As seen in Mac.

src/game/common/system/archivefile.cpp
```diff
@@ -154,8 +154,8 @@ bool ArchiveFile::Search_String_Matches(Utf8String string, Utf8String search)
         return false;
     }
 
-    const char *cstring = string.Peek();
-    const char *csearch = search.Peek();
+    const char *cstring = string.Str();
+    const char *csearch = search.Str();
```
xezon, line 158: As seen in World Builder.

src/game/common/system/datachunk.cpp
```diff
@@ -305,7 +306,6 @@ Utf16String DataChunkInput::Read_UnicodeString()
     }
 
     Decrement_Data_Left(sizeof(ch) * size);
-    string.Peek()[size] = L'\0';
```
xezon, line 308: This is a bug and would crash the game if `size` happened to be zero, because no string would be allocated in `string` then. The code above does not match original code. I assume OmniBlade tried to implement it differently to accommodate charset better? Anyway, with the implementation of `string += ch` above a string null terminator is not necessary to be set, and original code also does not call `Peek` to begin with.

tomsons26 (approved):

PR #780, opened 2022-11-07
Fixes build error for newer ffmpeg versions

feliwir:
Enforces const correctness and avoid deprecated members

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -44,16 +44,34 @@ jobs:
 
       - name: Install dependencies
         run: |
-          sudo apt-get update && sudo apt-get install -y libgtk-3-dev ninja-build clang-format-10 libwxgtk3.0-gtk3-dev libavcodec-dev libavformat-dev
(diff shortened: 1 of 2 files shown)


feliwir:
This only seems to work for FFmpeg 5.0 and great. Will add a version check later

xezon (approved):

xezon (approved):

xezon:
2 targets did not build. probably because old ubuntu was removed, but is still required?

feliwir:
> 2 targets did not build. probably because old ubuntu was removed, but is still required?

yes, i removed those. I guess it's because the "develop" branch still has it

feliwir:
Can this please get merged? It's really causing me a lot of headache under linux - i need to manually apply this fix on each branch i work on

xezon:
It cannot be merged, because 2 checks do not trigger here.

feliwir:
> It cannot be merged, because 2 checks do not trigger here.

I removed those 2 checks. They will never trigger

PR #786, opened 2022-11-09
Fix file function constness

xezon:
Adds `const` to File member functions, but not yet the virtual functions.

### DIFF
--- a/src/game/common/system/file.h
+++ b/src/game/common/system/file.h
@@ -89,9 +89,9 @@ public:
 
     bool Eof();
 
-    const char *Get_Name() { return m_name.Str(); }
+    const char *Get_Name() const { return m_name.Str(); }
     void Set_Name(const char *name) { m_name = name; }
-    int Get_Access() { return m_access; }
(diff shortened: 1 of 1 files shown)


feliwir (approved):

feliwir (approved):

xezon:
Looks like your approval is yet non consequential :D

feliwir:
> Looks like your approval is yet non consequential :D

i thought approving twice can't hurt :) It asked me to add my review again:
image

jonwil:
I have evidence that these were not const in the original. Is there a reason to deviate from that?

xezon:
Yes. const correctness is important.

xezon:
To elaborate on const correctness, here is article:
https://isocpp.org/wiki/faq/const-correctness

In a nutshell, anything that is meant to be immutable should be marked `const`, most importantly on

* immutable global variables
* immutable public struct members
* immutable private class members
* immutable member functions, that promise to not change state of callee
* immutable function arguments

`const` does not change the runtime.

PR #787, opened 2022-11-09
Simplify complicated iterator tests

xezon:
Simplification for programmer sanity sake.

### DIFF
--- a/src/game/logic/object/locomotor.cpp
+++ b/src/game/logic/object/locomotor.cpp
@@ -298,7 +298,7 @@ void LocomotorStore::Reset()
     auto it = m_locomotorTemplates.begin();
 
     for (;;) {
-        if (!(it != m_locomotorTemplates.end())) {
+        if (it == m_locomotorTemplates.end()) {
             break;
         }
 
--- a/src/game/common/rts/team.cpp
+++ b/src/game/common/rts/team.cpp
@@ -828,7 +828,7 @@ bool Team::Remove_Override_Team_Relationship(unsigned int id)
 
     auto it = m_teamRelations->m_relationships.find(id);
 
-    if (!(it != m_teamRelations->m_relationships.end())) {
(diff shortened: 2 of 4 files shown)


src/game/logic/object/locomotor.cpp
```diff
@@ -298,7 +298,7 @@ void LocomotorStore::Reset()
     auto it = m_locomotorTemplates.begin();
 
     for (;;) {
-        if (!(it != m_locomotorTemplates.end())) {
+        if (it == m_locomotorTemplates.end()) {
```
feliwir, line 301: all of this could be simplified with `std::remove_if`:
```cpp
m_locomotorTemplates.erase(std::remove_if(m_locomotorTemplates.begin(), m_locomotorTemplates.end(), [](auto t) { return t.second->DeleteOverrides() == nullptr; });
```

xezon: On that note, this algorithm looks suspicious.

Down below I would have expected
```c++
it = m_locomotorTemplates.erase(it);
```

Essentially loop will continue and then reference invalidated iterator. Looks like bug.

jonwil:
PR is obsolete, code has been changed and now seems correct to me. Closing.

xezon:
I do not think this needs to be closed. These double negative tests still occur and it would be nice to refactor this.

```
src\game\common\rts\team.cpp(831):if (!(it != m_teamRelations->m_relationships.end())) {
src\game\common\rts\team.cpp(1658):if (!(it != m_playerRelations->m_relationships.end())) {
src\game\common\system\archivefile.cpp(35):if (!(it != dirp->directories.end())) {
src\game\common\system\archivefile.cpp(100):if (!(it != dirp->directories.end())) {
src\game\logic\scriptengine\scriptengine.cpp(6851):if (!(it != m_toppleDirections.end())) {
```

PR #788, opened 2022-11-09
Reduce file string allocations

xezon:
Reduces unnecessary string allocations in file class. Optimization.

The way it works is `Utf8String` is reference counted. So by having a static string, constructor can borrow this static and simply increase its ref count when assigning it. It no longer needs to allocate this `"<no file>"` string, effectively elimitating 1 wasted allocation per file open.


tomsons26 (approved):

xezon:
Tested and works correctly.

PR #791, opened 2022-11-09
Implement zlib compression

feliwir:
Use zlib library to optional de/compress data


src/game/common/compression/zlibcompr.cpp
```diff
@@ -0,0 +1,62 @@
...
+    return dstSize;
+}
+
+/**
+ * Returns the max size for the specified amount of bytes.
+ */
+int Zlib_MaxSize(int size)
+{
+    return compressBound(size);
+}
```
xezon: eol

src/game/common/compression/zlibcompr.cpp
```diff
@@ -0,0 +1,62 @@
...
+{
+    if (src_size == 0) {
+        captainslog_error("Missing src_size for Zlib data");
+        return 0;
+    }
+
+    uLongf dstSize = dst_size;
+    uLong srcSize = src_size;
+
+    int result = uncompress((Bytef *)dst, &dstSize, (const Bytef *)src, srcSize);
```
xezon: static cast, same in next function

src/game/common/compression/zlibcompr.h
```diff
@@ -0,0 +1,21 @@
...
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#pragma once
+
+#include "always.h"
+
+int Zlib_Uncompress(void *dst, int dst_size, const void *src, int src_size);
+int Zlib_Compress(void *dst, const void *src, int size, int level);
+int Zlib_MaxSize(int size);
```
xezon: eol

src/platform/audio/ffmpegaudiofilecache.cpp
```diff
@@ -134,7 +134,7 @@ bool FFmpegAudioFileCache::Open_FFmpeg_Contexts(FFmpegOpenAudioFile *open_audio,
         return false;
     }
 
-    AVCodec *input_codec = avcodec_find_decoder(open_audio->fmt_ctx->streams[0]->codecpar->codec_id);
+    const AVCodec *input_codec = avcodec_find_decoder(open_audio->fmt_ctx->streams[0]->codecpar->codec_id);
```
xezon: Why is this here?

src/game/common/compression/zlibcompr.cpp
```diff
@@ -0,0 +1,62 @@
...
+ */
+int Zlib_Uncompress(void *dst, int dst_size, const void *src, int src_size)
+{
+    if (src_size == 0) {
+        captainslog_error("Missing src_size for Zlib data");
+        return 0;
+    }
+
+    uLongf dstSize = dst_size;
+    uLong srcSize = src_size;
```
xezon: Variable names need to be "name_name" format. Despite, these have the same name. Different name would be better.

src/game/common/compression/zlibcompr.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "zlibcompr.h"
+#include <captainslog.h>
+#include <zlib.h>
+
+/**
+ * Decompresses zlib format.
+ */
+int Zlib_Uncompress(void *dst, int dst_size, const void *src, int src_size)
+{
+    if (src_size == 0) {
```
xezon: <= 0 perhaps?

tests/test_compr.cpp
```diff
@@ -0,0 +1,121 @@
...
+    size_t src_size = src_file.Size();
+    uint8_t *src_data = new uint8_t[src_size];
+    src_file.Read(src_data, src_size);
+
+    uint8_t *dst_data = new uint8_t[src_size + 8];
+    uint32_t dst_size = RefPack_Compress(dst_data, src_data, src_size, true);
+    EXPECT_GT(dst_size, 0);
+    EXPECT_LT(dst_size, src_size);
+
+    delete[] src_data;
```
xezon: You can use `std::unique_ptr<>` to skip on `delete`. In several functions.

tests/test_compr.cpp
```diff
@@ -0,0 +1,121 @@
...
+    auto filepath = Utf8String(TESTDATA_PATH) + "/compr/uncompr.txt";
+    BufferedFileClass src_file(filepath.Str());
+    ASSERT_TRUE(src_file.Open(FM_READ));
+    size_t src_size = src_file.Size();
+    uint8_t *src_data = new uint8_t[src_size];
+    src_file.Read(src_data, src_size);
+
+    uint8_t *dst_data = new uint8_t[src_size + 8];
+    uint32_t dst_size = RefPack_Compress(dst_data, src_data, src_size, true);
+    EXPECT_GT(dst_size, 0);
```
xezon: This compression result check looks quite generous, as it would allow error margin on compression result. We do expect compressed data to be always the same right? Perhaps hash contents of `dst_data` and compare hash value against expected hash value.

In multiple functions.

tests/test_compr.cpp
```diff
@@ -0,0 +1,121 @@
...
+    size_t src_size = src_file.Size();
+    uint8_t *src_data = new uint8_t[src_size];
+    src_file.Read(src_data, src_size);
+
+    uint32_t dst_size = CompressionManager::Get_Max_Compressed_Size(src_size, m_type);
+    uint8_t *dst_data = new uint8_t[dst_size];
+    dst_size = CompressionManager::Compress_Data(m_type, src_data, src_size, dst_data, dst_size);
+    EXPECT_GT(dst_size, 0);
+    EXPECT_LT(dst_size, src_size);
+#if WRITE_DATA
```
xezon, line 70: You can add a new line before to make this block a bit easier on the eyes.

tests/test_compr.cpp
```diff
@@ -0,0 +1,121 @@
...
+        }
+    };
+
+protected:
+    CompressionType m_type;
+};
+
+TEST_P(CompressionTest, encode)
+{
+    auto filepath = Utf8String(TESTDATA_PATH) + "/compr/uncompr.txt";
```
xezon: I am not sure whether or not it is good to use `Utf8String` and `BufferedFileClass` in these compression test functions. It means these tests can fail if there is a problem in `Utf8String` & Co. If you intend to write separate tests for `Utf8String` and `BufferedFileClass`, then perhaps it would make sense to use `std::string` and `std::fstream` here instead. That way compression test is pure. Or we run tests for strings and files before this, so that they would fail before this.

src/game/common/compression/compressionmanager.h
```diff
@@ -35,12 +35,22 @@ enum CompressionType : int32_t
...
+};
+
 class CompressionManager
 {
 public:
     static bool Is_Data_Compressed(const void *data, int size);
+    static CompressionType Get_Prefered_Compression();
     static CompressionType Get_Compression_Type(const void *data, int size);
+    static int Get_Max_Compressed_Size(int size, CompressionType type);
+    static const char *Get_Compression_FourCC(CompressionType type);
```
xezon: Are these new functions as per original or are these custom for Thyme?

src/game/common/compression/compressionmanager.cpp
```diff
@@ -124,26 +193,80 @@ int CompressionManager::Get_Uncompressed_Size(const void *data, int size)
...
+                header->uncomp_size = src_size;
+                return compr_size + sizeof(ComprHeader);
+            }
+#endif
         case COMPRESSION_NONE:
         case COMPRESSION_NOX:
+        case COMPRESSION_EAB:
+        case COMPRESSION_EAH:
+        default:
+            captainslog_error("Compression format '%s' unhandled, file a bug report.\n",
```
xezon: Omit the "\n" and "file a bug report".

In multiple log messages.

Also, perhaps it should be promoted to Assert? Could be separate change.

src/platform/audio/ffmpegaudiofilecache.cpp
feliwir: Well i accidentally committed this since i need it locally to compile. However the diff should disappear once #780 is merged

tests/test_compr.cpp
feliwir: It's true that these function don't really check a lot, but i wanted to avoid that different zlib library version may fail. E.g. it's valid that a newer zlib version compresses better than an older one

src/game/common/compression/compressionmanager.h
feliwir: Apart from `Get_Compression_FourCC` all of these new functions exist in the original. Atleast that's the information @tomsons26  provided me with. I could probably get rid of `Get_Compression_FourCC` but i find it easier to read if there's a single place for that, instead of putting it everywhere

tests/test_compr.cpp
xezon: Hmm, is it true that compression result can differ that much, but then can still be read back with any zlib implementation? I would have expected that results will remain identical with identical settings, as otherwise it would not be able to decompress it again. I could be wrong.

feliwir: Well, depending on the compression level the size can vary from 728 bytes (level 1) -> 656 bytes (level 9) with different sizes inbetween accordingly. When the zlib maintainer finds a way to further decrease the size while maintaining compability with the DEFLATE algorithm he will certainly implement it. Also it might be possible that the flags for the different levels change and the size may change.

I know this is very unlikely, but the library doesn't give any guarantees that compression ratio doesn't change with different versions. E.g. if we wanted to switch to `libdeflate`in the future (a better zlib pretty much) we could get even better ratios: https://github.com/ebiggers/libdeflate#compression-levels

See here for a nice blog post: https://aras-p.info/blog/2021/08/09/EXR-libdeflate-is-great/

tests/test_compr.cpp
feliwir: Done

tests/test_compr.cpp
feliwir: Done

tests/test_compr.cpp
xezon: Is there a RefPack Decompress function? Perhaps pair that?

feliwir: What do you mean?

src/game/common/compression/compressionmanager.cpp
feliwir: I kept it as an error, changed the log message however

tests/test_compr.cpp
xezon: Original data -> RefPack Compress -> RefPack Decompress -> Compare decompressed with original.

src/game/common/compression/compressionmanager.cpp
```diff
@@ -43,12 +46,53 @@ bool CompressionManager::Is_Data_Compressed(const void *data, int size)
...
+ */
+CompressionType CompressionManager::Get_Prefered_Compression()
+{
+    return COMPRESSION_EAR;
+}
+
+/**
+ * @brief Get the FourCC for the corresponding compression type
+ */
+const char *CompressionManager::Get_Compression_FourCC(CompressionType type)
```
xezon: I do not agree with this function contract. `const char*` should not be used for non-null terminated string buffer. Right now, function user makes assumption about buffer length (4). Imo this should return a struct with char[4] or a uint32_t. Also, we do actually have function and macros for proper FourCC. You can look around other code to see how it is used.

tests/test_compr.cpp
feliwir: I kind of do that with the current tests - i compress the files (and save them on disk) and then decompress those files on disk again. I don't compare the hashes, but atleast if the size matches.

However it's a good idea do make a roundtrip test, that does a `memcmp` on the decompressed data with the original

src/game/common/compression/compressionmanager.cpp
feliwir: Fixed

src/game/common/compression/compressionmanager.cpp
```diff
@@ -43,12 +48,54 @@ bool CompressionManager::Is_Data_Compressed(const void *data, int size)
...
+            return FourCC<'Z', 'L', '7', '\0'>::value;
+        case COMPRESSION_ZL8:
+            return FourCC<'Z', 'L', '8', '\0'>::value;
+        case COMPRESSION_ZL9:
+            return FourCC<'Z', 'L', '9', '\0'>::value;
+        case COMPRESSION_NONE:
+        default:
+            captainslog_error("Compression format '%s' unhandled", Get_Compression_Name(type));
+            return FourCC<0, 0, 0, 0>::value;
+            ;
```
xezon: Spare semicolon.

tests/test_compr.cpp
xezon: Yes. I have seen the decompression in the other test. The simpler test would be to compress, compare, decompress, compare again, without writing compressed result to file. This reduces test file set. You would only need uncompr.txt. Or not even, you could make the text part of the cpp file and make everything in memory, skipping file IO entirely.

xezon (approved):
Merge #780 before this.

src/game/common/compression/compressionmanager.cpp
```diff
@@ -124,26 +196,80 @@ int CompressionManager::Get_Uncompressed_Size(const void *data, int size)
...
+        case COMPRESSION_ZL6:
+        case COMPRESSION_ZL7:
+        case COMPRESSION_ZL8:
+        case COMPRESSION_ZL9:
+#if BUILD_WITH_ZLIB
+            compr_size = Zlib_Compress(dst_data, dst_size - sizeof(ComprHeader), src, src_size, type + 1 - COMPRESSION_ZL1);
+            if (compr_size > 0) {
+                header->uncomp_size = src_size;
+                return compr_size + sizeof(ComprHeader);
+            }
```
OmniBlade, line 244: Are we not missing a break here in case Zlib_Compress returns < 0?

OmniBlade (dismissed):

src/game/common/compression/compressionmanager.cpp
feliwir: It can't return anything lower than 0

src/game/common/compression/zlibcompr.cpp
```diff
@@ -0,0 +1,62 @@
...
+        return 0;
+    }
+
+    uLongf zlib_dst_size = dst_size;
+    uLong zlib_src_size = src_size;
+
+    int result = uncompress(static_cast<Bytef *>(dst), &zlib_dst_size, static_cast<const Bytef *>(src), zlib_src_size);
+    if (result != Z_OK) {
+        captainslog_error("Failed to compress Zlib data. Error=%i", result);
+        return 0;
```
xezon, line 54: Optional: Log more detailed info about the type of error, and/or return different negative values depending on error type.

```
Z_BUF_ERROR
    The buffer dest was not large enough to hold the uncompressed data.

Z_MEM_ERROR
    Insufficient memory.

Z_DATA_ERROR
    The compressed data (referenced by source) was corrupted.
```

src/game/common/compression/zlibcompr.cpp
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Compresses to zlib data.
+ */
+int Zlib_Compress(void *dst, int dst_size, const void *src, int src_size, int level)
+{
+    uLongf zlib_dst_size = dst_size;
+    int result = compress2(static_cast<Bytef *>(dst), &zlib_dst_size, static_cast<const Bytef *>(src), src_size, level);
+    if (result != Z_OK) {
+        captainslog_error("Failed to compress Zlib data. Error=%i", result);
+        return 0;
```
xezon, line 69: Optional: Log more detailed info about the type of error, and/or return different negative values depending on error type.

```
Z_BUF_ERROR
    The buffer dest was not large enough to hold the compressed data.

Z_MEM_ERROR
    Insufficient memory.

Z_STREAM_ERROR
    The level was not Z_DEFAULT_LEVEL, or was not between 0 and 9.
```

src/game/common/compression/compressionmanager.h
xezon: Could you perhaps add a comment in the header for `Get_Compression_FourCC` that it is Thyme specific?

PR #793, opened 2022-11-11
Address sanitize

feliwir:
Try using direct allocations whenever possible - this makes it easier to find leaks than going through the memory allocators


src/game/common/system/memdynalloc.cpp
```diff
@@ -65,11 +66,12 @@ void DynamicMemoryAllocator::Init(MemoryPoolFactory *factory, int subpools, Pool
     for (int i = 0; i < m_poolCount; ++i) {
         m_pools[i] = m_factory->Create_Memory_Pool(&init_list[i]);
     }
+#endif
 }
 
 DynamicMemoryAllocator::~DynamicMemoryAllocator()
 {
-    captainslog_dbgassert(m_usedBlocksInDma, "Destroying none empty DMA.");
+    captainslog_dbgassert(m_usedBlocksInDma == 0, "Destroying none empty DMA.");
```
xezon: This should be separate change.

src/w3d/lib/hash.cpp
```diff
@@ -34,7 +34,7 @@ HashTableClass::HashTableClass(int size)
 HashTableClass::~HashTableClass()
 {
     if (m_hashTable) {
-        delete (m_hashTable);
+        delete[] m_hashTable;
```
xezon: This should be separate change.

src/game/common/system/gamememory.cpp
```diff
@@ -192,3 +194,4 @@ void operator delete[](void *ptr)
     Init_Memory_Manager_Pre_Main();
     g_dynamicMemoryAllocator->Free_Bytes(ptr);
 }
+#endif
```
xezon: eol

src/game/common/system/mempoolobj.h
```diff
@@ -32,9 +32,10 @@ class MemoryPoolObject
     virtual ~MemoryPoolObject() {}
     virtual MemoryPool *Get_Object_Pool() = 0;
 
+#ifndef __SANITIZE_ADDRESS__
```
xezon, line 35: Did you test if this runs properly with `GAME_DLL` configuration?

src/w3d/lib/iniclass.cpp
```diff
@@ -99,6 +99,14 @@ INIClass::INIClass(FileClass &file) : m_fileName(nullptr)
...
 {
     Clear();
+    Shutdown();
+}
+
+void INIClass::Shutdown()
+{
+    delete m_sectionList;
+    delete m_sectionIndex;
+    delete[] m_fileName;
```
xezon: This should be separate change.

src/w3d/lib/w3dmpo.h
```diff
@@ -29,6 +29,7 @@ class W3DMPO
     virtual ~W3DMPO() {}
 };
 
+#ifndef __SANITIZE_ADDRESS__
```
xezon: Prefer positive condition over negative one. Makes conditions easier to read.

tests/main.cpp
```diff
@@ -11,5 +13,13 @@ int main(int argc, char **argv)
...
+    Init_Memory_Manager();
+
+    int result = RUN_ALL_TESTS();
+
+    Shutdown_Memory_Manager();
+
+    captainslog_deinit();
+
+    return result;
 }
```
xezon: eol

tests/test_filesystem.cpp
```diff
@@ -122,10 +127,10 @@ TEST_P(FileSystemTest, list_dir_unfiltered)
     EXPECT_EQ(files.size(), 2);
 }
 
-LocalFileSystem *filesystem_list[] = {
-    new Win32LocalFileSystem,
+std::shared_ptr<LocalFileSystem> filesystem_list[] = {
```
xezon, line 130: std::unique_ptr should also do

src/game/common/system/mempoolobj.h
feliwir: No, but i hoped that noone would combine `USE_SANITIZE` with anything other than `STANDALONE`. However i can try to enforce this in CMake if preferred

tests/test_filesystem.cpp
feliwir: Unfortunately not - the pointer is being copied, which is deleted for std::unique_ptr. Probably could be avoided somehow, but shared_ptr works without extra effort

src/game/common/system/mempoolobj.h
xezon: You could also do a 

```c++
#if defined(__SANITIZE_ADDRESS__) && !defined(GAME_DLL)
#define USE_ADDRESS_SANITIZER 1
#endif
```

feliwir: I did a CMake check that throws an error when enabling both. It should not be able to generate such a configuration, since code with address sanitizing also can't be injected

feliwir:
I did a smaller PR (#799) which fixes the leaks. When we merge that first everything should be good here

CMakeLists.txt
```diff
@@ -104,6 +105,18 @@ if(WIN32 OR "${CMAKE_SYSTEM}" MATCHES "Windows")
     endif()
 endif()
 
+# Only use GCC for now, we can support the others later
```
xezon, line 108: Can remove this comment.

xezon (approved):
Tested with GAME_DLL. Game appears to run normally.

PR #796, opened 2022-11-13
Fixed bugs and memleaks

GermanAizek:
Hello everyone, review my bug catches. I didn't look at bugtracker, so I don't know if bug with report was fixed.

### DIFF
--- a/src/game/client/worldheightmap.cpp
+++ b/src/game/client/worldheightmap.cpp
@@ -693,7 +693,7 @@ bool WorldHeightMap::Parse_Blend_Tile_Data(DataChunkInput &file, DataChunkInfo *
         m_textureClasses[i].first_tile = file.Read_Int32();
         m_textureClasses[i].num_tiles = file.Read_Int32();
         m_textureClasses[i].width = file.Read_Int32();
-        file.Read_Int32();
+        m_textureClasses[i].is_blend_edge_tile = file.Read_Int32();
         m_textureClasses[i].name = file.Read_AsciiString();
         Read_Tex_Class(&m_textureClasses[i], m_sourceTiles);
     }
--- a/src/game/common/audio/audiomanager.cpp
+++ b/src/game/common/audio/audiomanager.cpp
@@ -507,7 +507,7 @@ void AudioManager::Lose_Focus()
     m_savedVolumes[0] = m_initialMusicVolume;
     m_savedVolumes[1] = m_initialSoundVolume;
     m_savedVolumes[2] = m_initial3DSoundVolume;
-    m_savedVolumes[3] = m_initialMusicVolume;
+    m_savedVolumes[3] = m_initialSpeechVolume;
     Set_Volume(0.0f, AUDIOAFFECT_MUSIC | AUDIOAFFECT_SOUND | AUDIOAFFECT_3DSOUND | AUDIOAFFECT_SPEECH | AUDIOAFFECT_BASEVOL);
 }
 
--- a/src/game/common/gamelod.cpp
+++ b/src/game/common/gamelod.cpp
@@ -303,7 +303,7 @@ StaticGameLODLevel GameLODManager::Find_Static_LOD_Level()
     }
 
     for (int i = STATLOD_HIGH; i >= STATLOD_LOW; --i) {
-        for (int j = 0; j < m_staticLODPresetCount[j]; ++j) {
+        for (int j = 0; j < m_staticLODPresetCount[i]; ++j) {
             if (m_cpuType == m_LODPresets[i][j].cpu_type) {
                 if ((double)m_cpuMHz / (double)m_LODPresets[i][j].mhz >= 0.94 && m_gpuType >= m_LODPresets[i][j].video_type
                     && (double)(m_physicalMem / 0x100000) / (double)m_LODPresets[i][j].video_mem >= 0.94) {
--- a/src/game/common/system/gamememoryinit.cpp
+++ b/src/game/common/system/gamememoryinit.cpp
@@ -732,5 +732,6 @@ void User_Memory_Init_Pools()
                 }
             }
         }
+        fclose(fp);
     }
 }
--- a/src/platform/w3dengine/client/w3droadbuffer.cpp
+++ b/src/platform/w3dengine/client/w3droadbuffer.cpp
@@ -25,6 +25,7 @@
 #include "terrainroads.h"
 #include "texture.h"
 #include "worldheightmap.h"
+#include <cmath>
 
 using GameMath::Fabs;
 bool g_dynamic;
@@ -2020,7 +2021,7 @@ bool W3DRoadBuffer::Insert_Y(Vector2 loc, int index1, float scale)
         Vector2 v4(-v1.Y, v1.X);
         if (xp_sign(v4, v2) == 1 && xp_sign(v4, v3) == 1) {
             b3 = true;
-            f6 = Fabs(f2 + 0.70700002f) + Fabs(f1 + 0.70700002f);
+            f6 = Fabs(f2 + M_SQRT1_2) + Fabs(f1 + M_SQRT1_2);
         }
     }
 
@@ -2031,7 +2032,7 @@ bool W3DRoadBuffer::Insert_Y(Vector2 loc, int index1, float scale)
         Vector2 v4(-v3.Y, v3.X);
         if (xp_sign(v4, v2) == 1 && xp_sign(v4, v1) == 1) {
             b1 = true;
-            f4 = Fabs(f3 + 0.70700002f) + Fabs(f2 + 0.70700002f);
+            f4 = Fabs(f3 + M_SQRT1_2) + Fabs(f2 + M_SQRT1_2);
         }
     }
 
@@ -2042,7 +2043,7 @@ bool W3DRoadBuffer::Insert_Y(Vector2 loc, int index1, float scale)
         Vector2 v4(-v3.Y, v3.X);
         if (xp_sign(v4, v3) == 1 && xp_sign(v4, v1) == 1) {
             b2 = true;
-            f5 = Fabs(f3 + 0.70700002f) + Fabs(f1 + 0.70700002f);
+            f5 = Fabs(f3 + M_SQRT1_2) + Fabs(f1 + M_SQRT1_2);
         }
     }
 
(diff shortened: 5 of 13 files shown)


src/platform/win32localfilesystem.cpp
```diff
@@ -64,7 +64,7 @@ File *Win32LocalFileSystem::Open_File(const char *filename, int mode)
     // Try and open the file, if not, delete instance and return null.
     if (!file->Open(filename, mode)) {
         file->Delete_Instance();
-        return nullptr;
+        file = nullptr;
```
xezon, line 67: This is just refactor right? No fix.

src/w3d/math/tri.h
```diff
@@ -97,7 +97,7 @@ inline bool Point_In_Triangle_2D(const Vector3 &tri_point0,
     } else {
         float p0p1dist2 = p0p1.Length2();
         float p1p2dist2 = p1p2.Length2();
-        float p2p0dist2 = p1p2.Length2();
+        float p2p0dist2 = p2p0.Length2();
```
xezon, line 100: This is also addressed in
* #319

src/w3d/renderer/inttest.h
```diff
@@ -66,6 +66,7 @@ inline bool AABoxIntersectionTestClass::Cull(const Vector3 &cull_min, const Vect
     if ((box_min.Z > cull_max.Z) || (box_max.Z < cull_min.Z)) {
         return true;
     }
+    return false;
```
xezon, line 69: Makes me wonder how did this ever compile?

xezon (commented):
Looks like many different kind of fixes. Maybe would be better to split this into multiple changes.

src/w3d/renderer/inttest.h
feliwir: int and bool functions do compile without return statements, but should atleast emit a warning. It's UB however

tomsons26: unfortunately, modern C++ standards allow this, on older MSVC it was a error, would love if we could make this a error not a warning somehow

src/platform/w3dengine/client/w3droadbuffer.cpp
```diff
@@ -25,6 +25,7 @@
 #include "terrainroads.h"
 #include "texture.h"
 #include "worldheightmap.h"
+#include <cmath>
```
tomsons26, line 28: should include our gmath then

src/w3d/math/vector2.h
```diff
@@ -308,8 +308,8 @@ __forceinline float Quick_Distance(float x1, float y1, float x2, float y2)
 {
     float x_diff = x1 - x2;
     float y_diff = y1 - y2;
-    GameMath::Fabs(x_diff);
-    GameMath::Fabs(y_diff);
+    x_diff = GameMath::Fabs(x_diff);
```
tomsons26, line 311: should be commented as a bugfix as original code is missing this

src/w3d/math/tri.h
tomsons26: yeah im not accepting bugfixes that we don't know the consequences of, for all we know the maps may be layed out to avoid this issue they couldn't find

src/game/client/worldheightmap.cpp
```diff
@@ -693,7 +693,7 @@ bool WorldHeightMap::Parse_Blend_Tile_Data(DataChunkInput &file, DataChunkInfo *
         m_textureClasses[i].first_tile = file.Read_Int32();
         m_textureClasses[i].num_tiles = file.Read_Int32();
         m_textureClasses[i].width = file.Read_Int32();
-        file.Read_Int32();
+        m_textureClasses[i].is_blend_edge_tile = file.Read_Int32();
```
tomsons26, line 696: indeed it's a bug in original, not too sure if this should be accepted, don't know the implications of fixing this..

src/w3d/renderer/inttest.h
tomsons26: for these two likely need to check what code uses them and how fixing this can impact the game

tomsons26:
another issue i noticed checking some other code
SysTimer Get should have
```
    if ( time > m_startTime ) {
        return time - m_startTime;
    }

    return time + m_negTime;
```

src/w3d/renderer/inttest.h
xezon: Ok but original was using old MSVC, so how was it in original then?

Here is list of disabled MSVC warnings.
https://learn.microsoft.com/en-us/cpp/preprocessor/compiler-warnings-that-are-off-by-default?view=msvc-170

I did not spot anything regarding function return statements.

src/w3d/renderer/sortingrenderer.cpp
```diff
@@ -564,7 +564,7 @@ void SortingRendererClass::Deinit()
 
     while (g_cleanList.Head() != nullptr) {
         auto head = g_cleanList.Head();
-        g_sortedList.Remove_Head();
+        g_cleanList.Remove_Head();
```
xezon, line 567: Is this BUGFIX or not?

src/w3d/renderer/scene.cpp
```diff
@@ -70,7 +70,7 @@ void SceneClass::Render(RenderInfoClass &rinfo)
             DX8Wrapper::Set_DX8_Render_State(D3DRS_ZBIAS, 7); // BFME2 calls Set_DX8_ZBias(7.0f) here
 #endif
             Customized_Render(rinfo);
-        } else if (m_extraPassPolyRenderMode == EXTRA_PASS_LINE) {
+        } else if (m_extraPassPolyRenderMode == EXTRA_PASS_CLEAR_LINE) {
```
xezon, line 73: Is this BUGFIX or not?

src/w3d/renderer/ddsfile.cpp
```diff
@@ -592,6 +592,7 @@ bool DDSFileClass::Load()
 
     if (data == 0) {
         captainslog_warn("DDSFile '%s' appears to contain no data, size is %d.", m_name, fp->Size());
+        fp->Close();
```
xezon, line 595: Is this BUGFIX or not?

src/game/common/system/gamememoryinit.cpp
```diff
@@ -732,5 +732,6 @@ void User_Memory_Init_Pools()
                 }
             }
         }
+        fclose(fp);
```
xezon, line 735: Is this BUGFIX or not?

src/game/common/gamelod.cpp
```diff
@@ -303,7 +303,7 @@ StaticGameLODLevel GameLODManager::Find_Static_LOD_Level()
     }
 
     for (int i = STATLOD_HIGH; i >= STATLOD_LOW; --i) {
-        for (int j = 0; j < m_staticLODPresetCount[j]; ++j) {
+        for (int j = 0; j < m_staticLODPresetCount[i]; ++j) {
```
xezon, line 306: Is this BUGFIX or not?

src/game/common/audio/audiomanager.cpp
```diff
@@ -507,7 +507,7 @@ void AudioManager::Lose_Focus()
     m_savedVolumes[0] = m_initialMusicVolume;
     m_savedVolumes[1] = m_initialSoundVolume;
     m_savedVolumes[2] = m_initial3DSoundVolume;
-    m_savedVolumes[3] = m_initialMusicVolume;
+    m_savedVolumes[3] = m_initialSpeechVolume;
```
xezon, line 510: Is this BUGFIX or not?

src/w3d/renderer/sortingrenderer.cpp
tomsons26: not, should be that

src/w3d/renderer/scene.cpp
tomsons26: not, should be that

src/w3d/renderer/ddsfile.cpp
tomsons26: not, should be that

src/game/common/system/gamememoryinit.cpp
tomsons26: not, should be that

src/game/common/gamelod.cpp
tomsons26: not, should be that

src/game/common/audio/audiomanager.cpp
tomsons26: not, should be that

src/game/client/worldheightmap.cpp
GermanAizek: @tomsons26 isn't point of restoring game sources in modifying them and fixing developer errors, as well as improving support for modern systems?

tomsons26: it is eventually, but we need to know what impact it has on the game, we don't want to break things or change behavior of things unintentionally or unknowingly

xezon:
I will split this change in smaller chunks.

src/platform/win32localfilesystem.cpp
xezon: Is not necessary to change, although is reasonable refactor.

src/platform/w3dengine/client/w3droadbuffer.cpp
```diff
@@ -2020,7 +2021,7 @@ bool W3DRoadBuffer::Insert_Y(Vector2 loc, int index1, float scale)
         Vector2 v4(-v1.Y, v1.X);
         if (xp_sign(v4, v2) == 1 && xp_sign(v4, v3) == 1) {
             b3 = true;
-            f6 = Fabs(f2 + 0.70700002f) + Fabs(f1 + 0.70700002f);
+            f6 = Fabs(f2 + M_SQRT1_2) + Fabs(f1 + M_SQRT1_2);
```
xezon, line 2024: Gamemath equivalent would be 
```c++
#define GAMEMATH_OOSQRT2 0.707106781f
```

However, value is slightly different. Perhaps should not change.

src/game/client/worldheightmap.cpp
xezon: With current code Thyme has, `is_blend_edge_tile` is not used anywhere.

xezon (commented):
I have split contents of this pull into separate ones so each one can be discussed individually. This one can be closed accordingly.

PR #798, opened 2022-11-15
A couple of small fixes

jonwil:


jonwil:
ModuleTemplate is confirmed to be a class and not a struct thanks to the rts.map file and its constructor is at 0x007A0F30 in zh wb.

m_locomotorTemplates is confirmed via AIUpdateModuleData::parseLocomotorSet in mac which calls std::map<LocomotorSetType,std::vector<LocomotorTemplate const*>>::operator[]

m_playerObjectCounts, m_completedSpeech and m_completedAudio are confirmed via the ScriptEngine constructor on mac (which calls std::map<AsciiString,int>::map and std::list<std::pair<AsciiString,unsigned int>>::list)

The replacement of 16 with MAX_PLAYER_COUNT seems logical given the use of MAX_PLAYER_COUNT for other arrays in the same class.

src/game/common/modules/modulefactory.h
```diff
@@ -34,11 +34,13 @@ typedef ModuleData *(*moddatacreateproc_t)(INI *);
...
 {
     ALLOW_HOOKING
-    struct ModuleTemplate
+    class ModuleTemplate
     {
+    public:
         modcreateproc_t create_proc;
         moddatacreateproc_t create_data_proc;
         int which_interfaces;
+        ModuleTemplate() : create_proc(0), create_data_proc(0), which_interfaces(0) {}
```
xezon: `create_proc(nullptr), create_data_proc(nullptr)` perhaps?

jonwil: fixed.

xezon (approved):
Looks logical to me.

PR #799, opened 2022-11-16
Fix memory leaks 

feliwir:


src/w3d/lib/hash.cpp
```diff
@@ -34,7 +34,7 @@ HashTableClass::HashTableClass(int size)
 HashTableClass::~HashTableClass()
 {
     if (m_hashTable) {
```
xezon, line 36: While at it, can also write `m_hashTable != nullptr` for style.

xezon (approved):

src/w3d/lib/hash.cpp
feliwir: Well, to be honest the if check makes no sense - delete on nullptr is valid

PR #800, opened 2022-11-17
Remove unused variables / statements

feliwir:
Those variables are never read after being written to - compiler already optimized that away, but it makes no sense to keep the code


xezon:
Could indicate error if there is mismatch with original.

feliwir:
> Could indicate error if there is mismatch with original.

Yes, either those variables should be used somewhere or the compiler back then didn't care about removing them

feliwir:
@tomsons26 could you please double check why these variables did exist? Is there an issue with the algorithm or was this just some artifact from the past?

feliwir:
deprecated due to #835

PR #802, opened 2022-11-23
Fix CI warnings

feliwir:
Need to use the latest checkout action version, since old NodeJS actions will be removed

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -26,7 +26,7 @@ jobs:
         os: [ubuntu-20.04, ubuntu-22.04]
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: "recursive"
 
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -35,11 +35,11 @@ jobs:
         run: |
           if [ "${{ matrix.platform }}" = "clang" ]
           then
-            echo "::set-output name=cc::clang"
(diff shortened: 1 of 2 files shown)


feliwir:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

OmniBlade (commented):
This looks fine to me, but given the time that has passed I'd like to rebase it on the latest develop and retest to make sure there is nothing else that needs a version bump.

.github/workflows/linux.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         os: [ubuntu-20.04, ubuntu-22.04]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
```
janisozaur: You need v4 by now.

```suggestion
      - uses: actions/checkout@v4
```

https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

.github/workflows/windows.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         platform: [x86, amd64]
           
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
```
janisozaur:
```suggestion
    - uses: actions/checkout@v4
```

janisozaur (changes requested):

OmniBlade (approved):

PR #809, opened 2022-12-07
Free the audiofilecache in destructor

feliwir:


src/platform/audio/milesaudiomanager.cpp
```diff
@@ -60,6 +60,7 @@ MilesAudioManager::~MilesAudioManager()
     Unselect_Provider();
```
tomsons26: indeed this is correct to pseudo, missed this when reviewing miles :X

can you also replace first if with Release_Bink_Handle and these 3 calls with Close_Device

feliwir: Done

tomsons26 (approved):

PR #810, opened 2022-12-08
Add codecov badge

feliwir:
Ignore specific folders that should not count into coverage


codecov.yml
```diff
@@ -0,0 +1,3 @@
+ignore:
+    - "deps/**/*"
+    - "tests/**/"
```
xezon: eol

feliwir: Fixed

xezon (commented):
Why are yml files removed when this change is about adding codecov widget?

feliwir:
> Why are yml files removed when this change is about adding codecov widget?

Those are artifacts from the past - travis and appveyor are not used anymore. Removing them made sense to me

xezon (approved):

PR #815, opened 2022-12-14
Update README.md

mountaser:
since github actions are being used link to it instead of the releases


xezon:
Is this the easiest way to access it? Is quite hidden.

PR #817, opened 2023-01-07
Implement WeaponStore

jonwil:


src/game/logic/object/weapon.cpp
```diff
@@ -155,6 +148,172 @@ Weapon *WeaponStore::Allocate_New_Weapon(const WeaponTemplate *tmpl, WeaponSlotT
...
+        m_weaponTemplateVector[i]->Reset();
+    }
+}
+
+void WeaponStore::Reset()
+{
+    for (unsigned int i = 0; i < m_weaponTemplateVector.size(); i++) {
+        WeaponTemplate *tmplate = m_weaponTemplateVector[i];
+
+        if (tmplate->Get_Next_Template()) {
```
tomsons26: should actually be a isOverride call 006DB780 wb
Friend_Clear_Next_Template returns what it popped, return isn't used in this function but i noticed its void

src/game/logic/object/weapon.cpp
```diff
@@ -155,6 +148,172 @@ Weapon *WeaponStore::Allocate_New_Weapon(const WeaponTemplate *tmpl, WeaponSlotT
...
+    WeaponTemplate *t = new WeaponTemplate();
+    t = tmplate;
+    t->Friend_Set_Next_Template(tmplate);
+    return t;
+}
+
+void WeaponStore::Update()
+{
+    for (auto i = m_weaponDDI.begin(); i != m_weaponDDI.end();) {
+        if (g_theGameLogic->Get_Frame() < i->m_delayDamageFrame) {
```
tomsons26: flip

src/game/logic/object/weapon.cpp
```diff
@@ -155,6 +148,172 @@ Weapon *WeaponStore::Allocate_New_Weapon(const WeaponTemplate *tmpl, WeaponSlotT
...
+    weapon->Load_Ammo_Now(source_obj);
+    weapon->Fire_Projectile_Detonation_Weapon(source_obj, victim_pos, bonus_condition, do_damage);
+    weapon->Delete_Instance();
+}
+
+void WeaponStore::Create_And_Fire_Temp_Weapon(
+    const WeaponTemplate *tmplate, const Object *source_obj, const Coord3D *victim_pos)
+{
+    if (tmplate) {
+        Weapon *weapon = g_theWeaponStore->Allocate_New_Weapon(tmplate, WEAPONSLOT_PRIMARY);
```
tomsons26, line 182: Allocate_New_Weapon was makeWeapon

src/game/logic/object/weapon.cpp
```diff
@@ -155,6 +148,172 @@ Weapon *WeaponStore::Allocate_New_Weapon(const WeaponTemplate *tmpl, WeaponSlotT
...
+    }
+}
+
+void WeaponStore::Parse_Weapon_Template_Definition(INI *ini)
+{
+    Utf8String str;
+    str.Set(ini->Get_Next_Token());
+    WeaponTemplate *tmplate = g_theWeaponStore->Find_Weapon_Template_Private(g_theNameKeyGenerator->Name_To_Key(str.Str()));
+
+    if (tmplate == nullptr) {
```
tomsons26, line 288:
```
    if (tmplate == nullptr) {
        tmplate = g_theWeaponStore->New_Weapon_Template(str);
    } else if (ini->Get_Load_Type() == INI_LOAD_CREATE_OVERRIDES) {
        tmplate = g_theWeaponStore->New_Override(tmplate);
    } else {
        captainslog_dbgassert(false, "Weapon '%s' already exists, but OVERRIDE not specified", str.Str());
    }
    
    if (tmplate)
        ini->Init_From_INI(tmplate, WeaponTemplate::Get_Field_Parse());
    }
    ....
```
i think?

xezon (commented):
Nothing to add from my side.

src/game/logic/object/weapon.cpp
jonwil: fixed.

src/game/logic/object/weapon.cpp
jonwil: fixed.

src/game/logic/object/weapon.cpp
jonwil: allocateNewWeapon exists in rts.map as public: class Weapon near * thiscall WeaponStore::allocateNewWeapon(class WeaponTemplate const near *,enum  WeaponSlotType)const which matches the function prototype in ZH WB.
makeWeapon has a different prototype (no WeaponSlotType and it returns Weapon rather than Weapon *) so I think allocateNewWeapon is the correct name.

src/game/logic/object/weapon.cpp
jonwil: fixed.

tomsons26 (approved):

PR #821, opened 2023-01-17
Disable HOVER Locomotor assert for ease of testing

xezon:
As per code comment: Comments assert because it triggers every frame when using Fire Wall ability of China Dragon Tank.

We will fix it in Game Patch Project.


xezon:
I tested game with fixed INI setup, but Flame effect looks bad then. It does not travel quickly enough. I think this assert should be removed. It likely is outdated. HOVER Locomotor with minSpeed > 0 works properly.

PR #831, opened 2023-01-28
Continue implementing GameEngine::Init and the things it needs.

jonwil:


src/game/common/gameengine.cpp
```diff
@@ -72,28 +82,82 @@ void GameEngine::Init(int argc, char *argv[])
 {
     INI ini;
 
+#ifdef GAME_DEBUG_STRUCTS
+    if (g_theVersion != nullptr) {
+        captainslog_debug("================================================================================");
+        captainslog_debug("Generals version %s (%s)", g_theVersion->Get_Ascii_Version().Str(), "Internal");
```
xezon: Make this "Thyme version" to hide the brand name.

src/game/common/gameengine.cpp
```diff
@@ -72,28 +82,82 @@ void GameEngine::Init(int argc, char *argv[])
...
+        captainslog_debug("Branch: %s", g_theVersion->Get_Ascii_Branch().Str());
+        captainslog_debug("Build commit: %s", g_theVersion->Get_Ascii_Commit_Hash().Str());
+        captainslog_debug("================================================================================");
+    }
+
+    captainslog_debug("Calculating CPU frequency for performance timers.");
+    // TODO processor frequency stuff
+#endif
+
+    m_maxFPS = 45;
```
xezon, line 99: 45 is odd FPS value. Not 30 or 60?

src/game/common/gameengine.cpp
```diff
@@ -72,28 +82,82 @@ void GameEngine::Init(int argc, char *argv[])
...
+
     g_theSubsystemList = new SubsystemInterfaceList;
     g_theSubsystemList->Add_Subsystem(this);
+
     Init_Random();
+
     g_theFileSystem = Create_File_System();
+    // DeleteFile("Data\\INI\\INIZH.big"); used to exist here, not sure if we want to copy that or not
```
xezon: It is an obsolete instruction, because there is no big such .big file at that location on a fresh game install.

Maybe write like so:

```c++
#if 0
    // Obsolete file deletion.
    DeleteFile("Data\\INI\\INIZH.big");
#endif
```

src/game/common/gameengine.cpp
```diff
@@ -72,28 +82,82 @@ void GameEngine::Init(int argc, char *argv[])
...
+
     g_theNameKeyGenerator = new NameKeyGenerator;
     g_theNameKeyGenerator->Init();
+#ifdef GAME_DEBUG_STRUCTS
+    captainslog_debug("----------------------------------------------------------------------------After "
```
xezon, line 114: The vast amount of dashes is a bit unsightly. Let us reduce them to just a few.

src/game/common/rts/specialpower.h
```diff
@@ -0,0 +1,63 @@
...
+#include "always.h"
+#include "audioeventrts.h"
+#include "subsysteminterface.h"
+
+class SpecialPowerTemplate : public Overridable
+{
+    IMPLEMENT_POOL(SpecialPowerTemplate)
+
+public:
+    virtual ~SpecialPowerTemplate() override;
```
xezon, line 25: Constructor is missing. Is that meant to be added in follow up change?

src/game/common/gameengine.cpp
jonwil: fixed.

src/game/common/gameengine.cpp
jonwil: It's what the code does.

src/game/common/gameengine.cpp
jonwil: Fixed.

src/game/common/rts/specialpower.h
jonwil: None of the functions of this class need implementing yet so this just defines the virtual functions and the member variables.

xezon (approved):
Tested some Armor, Damage Effects and Special Powers. Looked as usual.

PR #835, opened 2023-02-01
Fix RefPack errors

feliwir:


src/game/common/compression/refpack.cpp
```diff
@@ -47,7 +47,7 @@ static uint32_t RefPack_Matchlen(const uint8_t *s, const uint8_t *d, uint32_t ma
 /**
  * Compresses data using refpack LZ method
  */
-static int RefPack_Encode(const void *src, int src_len, void *dst, int dst_len, bool quick)
+static int RefPack_Encode(const void *src, int src_len, void *dst, int dst_len, int *opts)
```
tomsons26: opts isn't passed on further, this is quick in this function
and quick in RefPack_Uncompress is passed as false

xezon:
I suggest test Ref Pack compress and uncompress against original compressed .map file. Uncompress it first, then compress it again, then binary compare original .map with new .map file. If files are a match, then Ref Pack is success.

feliwir:
> I suggest test Ref Pack compress and uncompress against original compressed .map file. Uncompress it first, then compress it again, then binary compare original .map with new .map file. If files are a match, then Ref Pack is success.

compression tests are added in my zlib PR (#791)

xezon:
> compression tests are added in my zlib PR (#791)

To clarify, the request here is to check Ref Pack correctness against original game map data. Testing Ref Pack against artificial test data will only check correctness against itself, but not against third party compressed data.

tomsons26 (approved):

xezon:
Merge this?

PR #848, opened 2023-02-08
Refpack/ZLib de-/compressor tool

feliwir:


src/tools/CMakeLists.txt
```diff
@@ -10,3 +10,6 @@ if(WIN32 OR CMAKE_SYSTEM_NAME STREQUAL "Windows")
     add_subdirectory(wdump)
     add_subdirectory(wndedit)
 endif()
+
+# These tools are generic and API independent
+add_subdirectory(compressor)
```
xezon: eol

src/tools/compressor/compressor.cpp
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        if (output_file.Write(output_data.get(), output_size) != output_size) {
+            std::cerr << "Failed to write output data completly" << std::endl;
+            return EXIT_FAILURE;
+        }
+    }
+
+    return EXIT_SUCCESS;
+}
```
xezon: eol

tests/CMakeLists.txt
```diff
@@ -28,4 +29,15 @@ else()
...
+if(BUILD_TOOLS)
+  # Test compression & decompression
+  set(COMPR_TYPES EAR ZL1 ZL2 ZL3 ZL4 ZL5 ZL6 ZL7 ZL8 ZL9)
+  foreach(TYPE ${COMPR_TYPES})
+    add_test(NAME "compr_${TYPE}" COMMAND compressor -t ${TYPE} -i ${THYME_TESTDATA_PATH}/compr/uncompr.txt -o ${CMAKE_CURRENT_BINARY_DIR}/compr_${TYPE}.data)
+    set_tests_properties("compr_${TYPE}" PROPERTIES FIXTURES_SETUP DATA_${TYPE})
+    add_test(NAME "decompr_${TYPE}" COMMAND compressor -d -i ${CMAKE_CURRENT_BINARY_DIR}/compr_${TYPE}.data -o ${CMAKE_CURRENT_BINARY_DIR}/uncompr_${TYPE}.txt)
+    set_tests_properties("decompr_${TYPE}" PROPERTIES FIXTURES_REQUIRED DATA_${TYPE})
+  endforeach()
+endif()
```
xezon: eol

src/w3d/lib/rawfileclass.cpp
```diff
@@ -182,10 +182,10 @@ bool RawFileClass::Open(int rights)
 
     switch (rights) {
         case FM_READ:
-            m_handle = open(m_filename, O_RDONLY, S_IREAD);
+            m_handle = open(m_filename, O_RDONLY, S_IREAD | S_IWRITE);
```
xezon, line 185: If I am not mistaken this means no 2 reads can access this file at the same time. This does not seem right.

src/w3d/lib/rawfileclass.cpp
```diff
@@ -182,10 +182,10 @@ bool RawFileClass::Open(int rights)
...
             break;
         case FM_WRITE:
-            m_handle = open(m_filename, O_WRONLY | O_CREAT | O_TRUNC, S_IWRITE);
+            m_handle = open(m_filename, O_WRONLY | O_CREAT | O_TRUNC, S_IREAD | S_IWRITE);
```
xezon, line 188: Why is this added?

CMakeLists.txt
```diff
@@ -313,31 +314,41 @@ if(USE_CRASHPAD)
...
+            set_property(DIRECTORY ${wxWidgets_SOURCE_DIR} PROPERTY EXCLUDE_FROM_ALL YES)
+        endif()
+
+        set(wxWidgets_FOUND TRUE)
+        set(wxWidgets_LIBRARIES wxcore wxbase wxxrc wxxml wxhtml wxadv)
+
+        # Will need manually setting for cross compiles for windows or macos from another OS.
+        if(NOT wxWidgets_wxrc_EXECUTABLE)
+            set(wxWidgets_wxrc_EXECUTABLE wxrc)
+        endif()
```
xezon, line 378: GitHub does not do a good job showing the diff for this file. It looks like you touched wxWidgets and added new cxxopts, but GitHub diff shows it mingled. Perhaps it should be a 2 step change for clarity of changes?

src/w3d/lib/rawfileclass.cpp
feliwir: No, that should have no impact. It's only about the permission flags that the files get created with. We're still opening the file with `O_RDONLY` which is the readmode

CMakeLists.txt
feliwir: I also updated to most recent wxWidgets version (v3.2.1), since v3.2.0 had bugs on linux

src/w3d/lib/rawfileclass.cpp
xezon: Ok. Is this how original Thyme does it? If not, we need BUGFIX comment explaining the fix.

src/w3d/lib/rawfileclass.cpp
feliwir: The file is not readable on the system without this flag.

src/w3d/lib/rawfileclass.cpp
feliwir: It's a bug that we introduced. This codepath does not exist in the original

src/tools/compressor/compressor.cpp
```diff
@@ -0,0 +1,110 @@
+#include "bufffileclass.h"
+#include "compressionmanager.h"
+#include <cxxopts.hpp>
+
+int main(int argc, char **argv)
+{
+    cxxopts::Options options(
+        "Compressor", "A commandline tool to encode/decode from all SAGE formats. By default compress FILES");
```
xezon, line 9: "with all SAGE formats"

src/tools/compressor/compressor.cpp
```diff
@@ -0,0 +1,110 @@
...
+    BufferedFileClass output_file(output_path.c_str());
+    if (!output_file.Open(FM_WRITE)) {
+        std::cerr << "Failed to open output file:" << output_path << std::endl;
+        return EXIT_FAILURE;
+    }
+
+    auto input_size = input_file.Size();
+    std::unique_ptr<uint8_t[]> input_data(new uint8_t[input_size]);
+    if (input_file.Read(input_data.get(), input_size) != input_size) {
+        std::cerr << "Failed to read input data completly" << std::endl;
```
xezon: completely misspelled

src/tools/compressor/compressor.cpp
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    // Decompression
+    if (result["decompress"].as<bool>()) {
+        if (!CompressionManager::Is_Data_Compressed(input_data.get(), input_size)) {
+            std::cerr << "Input data is not compressed with a supported compression" << std::endl;
+            return EXIT_FAILURE;
+        }
+
+        auto output_size = CompressionManager::Get_Uncompressed_Size(input_data.get(), input_size);
```
xezon: Can this be zero?

src/tools/compressor/compressor.cpp
```diff
@@ -0,0 +1,110 @@
...
+    ("i,input", "input file", cxxopts::value<std::string>())
+    ("o,output", "output file", cxxopts::value<std::string>())
+    ("d,decompress", "decompress the input")
+    ("t,type", "specify the encoding format (encode only)", cxxopts::value<std::string>())
+    ("h,help", "print usage")
+    ("v,verbose", "verbose output", cxxopts::value<bool>()->default_value("false"))
+    ;
+    // clang-format on
+
+    auto result = options.parse(argc, argv);
```
xezon: Many of these vars in this function can be const.

src/tools/compressor/compressor.cpp
```diff
@@ -0,0 +1,110 @@
...
+        "Compressor", "A commandline tool to encode/decode from all SAGE formats. By default compress FILES");
+
+    options.allow_unrecognised_options();
+
+    // clang-format off
+    options.add_options()
+    ("i,input", "input file", cxxopts::value<std::string>())
+    ("o,output", "output file", cxxopts::value<std::string>())
+    ("d,decompress", "decompress the input")
+    ("t,type", "specify the encoding format (encode only)", cxxopts::value<std::string>())
```
xezon: How does user know accepted types?

src/tools/compressor/compressor.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+        if (output_file.Write(output_data.get(), output_size) != output_size) {
+            std::cerr << "Failed to write output data completly" << std::endl;
+            return EXIT_FAILURE;
+        }
+    } else {
+        CompressionType type = CompressionManager::Get_Prefered_Compression();
+        if (result.count("type") > 0) {
+            auto compr_name = result["type"].as<std::string>();
+            type = CompressionManager::Get_Compression_Type(compr_name.c_str(), compr_name.size() + 5);
```
xezon: `compr_name.size() + 5` ??

src/tools/compressor/compressor.cpp
```diff
@@ -0,0 +1,110 @@
...
+        std::unique_ptr<uint8_t[]> output_data(new uint8_t[output_size]);
+
+        output_size = CompressionManager::Compress_Data(type, input_data.get(), input_size, output_data.get(), output_size);
+        if (output_size == 0) {
+            std::cerr << "Failed to compress data correctly" << std::endl;
+            return EXIT_FAILURE;
+        }
+
+        if (output_file.Write(output_data.get(), output_size) != output_size) {
+            std::cerr << "Failed to write output data completly" << std::endl;
```
xezon: completely misspelled

src/tools/compressor/compressor.cpp
```diff
@@ -76,14 +80,19 @@ int main(int argc, char **argv)
...
+            std::cerr << "Failed to write output data completely" << std::endl;
             return EXIT_FAILURE;
         }
     } else {
         CompressionType type = CompressionManager::Get_Prefered_Compression();
         if (result.count("type") > 0) {
-            auto compr_name = result["type"].as<std::string>();
-            type = CompressionManager::Get_Compression_Type(compr_name.c_str(), compr_name.size() + 5);
+            const auto compr_name = result["type"].as<std::string>();
+            if (compr_name.size() < 4) {
```
xezon: Is this check really necessary? Wouldn't `CompressionManager::Get_Compression_Type` already cover this by returning `COMPRESSION_NONE` when size is smaller 4?

feliwir: We're not using this function as intended (it should be used on compressed data). However we use it on the `--type` parameter to just get the compression type.

We need this check since we do a memcmp on the `compr_name`. The size parameter we pass is actually not related to the size of `compr_name`

xezon: How about we improve `Get_Compression_Type` to take a char* and do a strcmp instead?

src/game/rtsutils.h
```diff
@@ -89,6 +89,17 @@ template<int a, int b, int c, int d> struct FourCC
 #endif
 };
 
+inline uint32_t FourCC_By_String(const char *s)
+{
+    if (strnlen(s, 5) > 4)
```
xezon: This will fail if s is smaller than 4, which is a valid input for things like "EAR\0" -> length of 3.

I suggest to create a `char buffer[4]`, init to 0, copy `s` into it for a maximum of 4 chars, and then convert that to the `uint32_t`.

This way `s` can be any length and this will work.

src/game/rtsutils.h
```diff
@@ -89,6 +89,17 @@ template<int a, int b, int c, int d> struct FourCC
...
+        return 0;
+#ifdef SYSTEM_LITTLE_ENDIAN
+    return (s[3] << 24) | (s[2] << 16) | (s[1] << 8) | s[0];
+#else
+    return (s[0] << 24) | (s[1] << 16) | (s[2] << 8) | s[3];
```
xezon: This means on Big Endian system, compression tool user would need to supply "\0RAE" string to select this option. Is that what you want?

xezon: On second thought, I could be mistaken. Best to check in debugger.

src/game/common/compression/compressionmanager.cpp
```diff
@@ -98,61 +130,7 @@ CompressionType CompressionManager::Get_Compression_Type(const void *data, int s
         return COMPRESSION_NONE;
     }
 
-    CompressionType type = COMPRESSION_NONE;
```
xezon: I suggest to leave the legacy code here as is and put it in a `#ifdef GAME_DLL`.

```c++
#ifdef GAME_DLL

CompressionType CompressionManager::Get_Compression_Type(const void *data, int size)
{
... Legacy code here
}

#else // !GAME_DLL

CompressionType CompressionManager::Get_Compression_Type(const void *data, int size)
{
    if (size < sizeof(ComprHeader)) {
        return COMPRESSION_NONE;
    }

    const char *data_str = static_cast<const char *>(data);
    const uint32_t id = rts::FourCC_By_String(data_str);
    return Get_Compression_Type_By_FourCC(id);
}

#endif
```

src/game/common/compression/compressionmanager.h
```diff
@@ -50,6 +50,8 @@ class CompressionManager
     static int Get_Max_Compressed_Size(int size, CompressionType type);
     // Thyme specific: This is a utility function
     static uint32_t Get_Compression_FourCC(CompressionType type);
+    // Thyme specific: This is a utility function
```
xezon: Perhaps describe the purpose of the utility function instead of writing it is a utility function.

src/game/rtsutils.h
```diff
@@ -89,6 +89,18 @@ template<int a, int b, int c, int d> struct FourCC
 #endif
 };
 
+inline uint32_t FourCC_By_String(const char *str)
+{
+    char buf[4] = {};
+    if (strlcpy(buf, str, 4) > 4)
```
xezon: This test is obsolete.

xezon: Additionally, `strlcpy` on `buf[4]` is not correct. That is because `strlcpy` will always write null character to target buffer to close it. This means, buf will have at most 3 non-zero chars. Either make buf size 5 or use different method for copying from source string.

Additionally, you can simplify calls to `strlcpy` on buffers by using `strlcpy_tpl`. This way you skip on passing buffer size.

src/game/rtsutils.h
```diff
@@ -89,6 +89,18 @@ template<int a, int b, int c, int d> struct FourCC
...
+        return 0;
+#ifdef SYSTEM_LITTLE_ENDIAN
+    return (buf[3] << 24) | (buf[2] << 16) | (buf[1] << 8) | buf[0];
+#else
+    return (buf[0] << 24) | (buf[1] << 16) | (buf[2] << 8) | buf[3];
```
xezon, line 99: Did you test this swap? Does it invoke expected behaviour down the line?

src/tools/compressor/compressor.cpp
```diff
@@ -0,0 +1,115 @@
...
+        "Compressor", "A commandline tool to encode/decode from all SAGE formats. By default compress FILES");
+
+    options.allow_unrecognised_options();
+
+    // clang-format off
+    options.add_options()
+    ("i,input", "input file", cxxopts::value<std::string>())
+    ("o,output", "output file", cxxopts::value<std::string>())
+    ("d,decompress", "decompress the input")
+    ("t,type", "specify the encoding format (EAR, ZL1-ZL9)", cxxopts::value<std::string>())
```
xezon, line 18: Not sure if `ZL1-ZL9` is entirely clear to every reader.

src/tools/compressor/compressor.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+        if (output_file.Write(output_data.get(), output_size) != output_size) {
+            std::cerr << "Failed to write output data completely" << std::endl;
+            return EXIT_FAILURE;
+        }
+    } else {
+        CompressionType type = CompressionManager::Get_Prefered_Compression();
+        if (result.count("type") > 0) {
+            const auto compr_name = result["type"].as<std::string>();
+            type = CompressionManager::Get_Compression_Type_By_FourCC(rts::FourCC_By_String(compr_name.c_str()));
```
xezon: I suggest break this line in 2 lines for readability.

src/game/rtsutils.h
```diff
@@ -89,6 +89,18 @@ template<int a, int b, int c, int d> struct FourCC
 #endif
 };
 
+inline uint32_t FourCC_By_String(const char *str)
```
xezon: Not sure if `FourCC_By_String` is proper name. Perhaps `FourCC_From_String`?

src/game/rtsutils.h
feliwir: On little endian systems it works - i can't really check with big endian. The swap might be unnecessary

xezon: You can test by just faking big endian system - simply swap the macros.

src/game/common/compression/compressionmanager.cpp
```diff
@@ -89,6 +89,51 @@ uint32_t CompressionManager::Get_Compression_FourCC(CompressionType type)
...
+            return COMPRESSION_ZL8;
+        case FourCC<'Z', 'L', '9', '\0'>::value:
+            return COMPRESSION_ZL9;
+        default:
+            captainslog_error("Compression fourcc '%i' unhandled", fourcc);
+            return COMPRESSION_NONE;
+    }
+}
+
+#ifndef GAME_DLL
```
xezon: Minor: Prefer non negated conditions.

Aka favor `#ifdef GAME_DLL` over `#ifndef GAME_DLL`.

This way there are less mental gymnastics required when looking at conditions.

src/tools/compressor/compressor.cpp
```diff
@@ -76,14 +81,15 @@ int main(int argc, char **argv)
...
+            std::cerr << "Failed to write output data completely" << std::endl;
             return EXIT_FAILURE;
         }
     } else {
         CompressionType type = CompressionManager::Get_Prefered_Compression();
         if (result.count("type") > 0) {
-            auto compr_name = result["type"].as<std::string>();
-            type = CompressionManager::Get_Compression_Type(compr_name.c_str(), compr_name.size() + 5);
+            const auto compr_name = result["type"].as<std::string>();
+            uint32_t fourCC = rts::FourCC_From_String(compr_name.c_str());
```
xezon: const

src/game/common/compression/compressionmanager.cpp
```diff
@@ -89,6 +89,39 @@ uint32_t CompressionManager::Get_Compression_FourCC(CompressionType type)
...
+        case FourCC<'Z', 'L', '6', '\0'>::value:
+            return COMPRESSION_ZL6;
+        case FourCC<'Z', 'L', '7', '\0'>::value:
+            return COMPRESSION_ZL7;
+        case FourCC<'Z', 'L', '8', '\0'>::value:
+            return COMPRESSION_ZL8;
+        case FourCC<'Z', 'L', '9', '\0'>::value:
+            return COMPRESSION_ZL9;
+        default:
+            captainslog_error("Compression fourcc '%i' unhandled", fourcc);
```
xezon: %u for unsigned integer.

src/game/rtsutils.h
xezon: feliwir
> i'll test what you suggest
> seems to work

xezon (approved):
Looks good to me.

PR #869, opened 2023-02-20
Implement a couple things that will make it easier to see what we need to implement to move forward with standalone

jonwil:


xezon (approved):
Looks good to me. Game launches, Shell Map runs, Test map with all faction things loads and runs. Tested in RelWithDebInfo.

PR #870, opened 2023-02-21
Updates GameMath version used.

### LINKED ISSUE #864: Need to add fast math functions
There are many places in our code where we call into slow deterministic gm_xxxx functions from GameMath but that don't need to be deterministic. We should add faster versions of Max, Min, Pow, Cos, Sin, Sqrt, Acos, Asin, Atan, Atan2, Tan, Ceil, Floor, Fast_To_Int_Floor, Fast_To_Int_Ceil, Fast_To_Int_Truncate, Fast_Float_Floor, Fast_Float_Ceil and Lrintf and use those in places where we are confident that the results are not going to be sent over the network or stored in any files (save games, replays, maps etc)

OmniBlade:
* Fixes #864

GameMath has been updated to make use of CPU instructions in some functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,7 +248,7 @@ if(USE_GAMEMATH)
     FetchContent_Declare(
         gamemath_git
         GIT_REPOSITORY https://github.com/TheAssemblyArmada/GameMath.git
-        GIT_TAG        f133b2cc8f85de6c2967ae77177daf0b8bbf39c0
+        GIT_TAG        59f7ccd494f7e7c916a784ac26ef266f9f09d78d
     )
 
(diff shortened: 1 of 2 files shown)


feliwir:
Could be maybe merge #867 before, so we can see the perf change?

jonwil (approved):
Looks OK to me.

xezon:
Related change: https://github.com/TheAssemblyArmada/GameMath/commit/4d2d8150717f4fa9dc429f8227c81a454ec2ffc6

xezon (approved):
Tested in RelWithDebInfo. Works. Performance is through the roof. On my machine it now runs with 300 FPS where it ran with 50 FPS prior this fix.

Fixes #864.

duncanspumpkin (approved):

PR #896, opened 2023-04-15
Implement AcademyStats

jonwil:


src/game/common/rts/academystats.cpp
```diff
@@ -13,34 +13,761 @@
...
 
 void AcademyStats::Xfer_Snapshot(Xfer *xfer)
 {
-#ifdef GAME_DLL
-    Call_Method<void, SnapShot, Xfer *>(PICK_ADDRESS(0x00576BF0, 0x008DF4B4), this, xfer);
-#endif
+    unsigned char version = 1;
+    xfer->xferVersion(&version, 1);
+    xfer->xferUnsignedInt(&m_lastUpdateFrame);
+    xfer->xferBool(&m_needsUpdate);
```
xezon, line 731: m_commandSet, m_commandCenterTemplate, m_supplyCenterTemplate are not serialized?

src/game/common/rts/academystats.cpp
```diff
@@ -13,34 +13,761 @@
...
+    if (power->Get_Academy_Classify() == ACT_SUPERPOWER) {
+        m_supersUsedCount++;
+    }
+}
+
+void AcademyStats::Record_Income()
+{
+    unsigned int frame = g_theGameLogic->Get_Frame();
+
+    unsigned int income_frame = m_incomeFrame1 - frame;
```
xezon, line 281: Looks different to `unsigned int power_frame = frame - m_powerFrame2;`. I wonder if there is mistake.

src/game/common/rts/academystats.cpp
```diff
@@ -13,34 +13,761 @@
...
+    if (info == nullptr) {
+        captainslog_dbgassert(false, "AcademyStats::Calculate_Academy_Advice() was passed in NULL AcademyAdviceInfo.");
+        return false;
+    } else {
+        info->m_adviceCount = 0;
+
+        for (int i = 0; i < max_advice_count; i++) {
+            Utf16String str;
+            str = "\n\n";
+            info->m_adviceString[i].Format(str);
```
xezon, line 709: This looks odd. It effectively just assigns \n\n to advice string?

src/game/common/rts/academystats.cpp
jonwil: No they are not.

src/game/common/rts/academystats.cpp
jonwil: This is correct.

src/game/common/rts/academystats.cpp
jonwil: This is correct.

xezon:
I do not know how to test this code.

xezon:
Code can be tested by looking at "Tactical Analysis" text in Game Score screen. It shows different text presets depending on what the player has done in a match.

This works in principle.

xezon (approved):
Looks good.

PR #917, opened 2023-05-16
Improves readability in BaseHeightMapRenderObjClass::Get_Height_Map_Height

bobtista:


jonwil (approved):
Looks good to me.

xezon:
The variable name convention for non-member non-global is lowercase with underscore: my_variable_name.

bobtista:
> The variable name convention for non-member non-global is lowercase with underscore: my_variable_name.

Updated, thanks!

src/game/client/baseheightmap.cpp
```diff
@@ -968,86 +968,88 @@ void BaseHeightMapRenderObjClass::Do_The_Light(VertexFormatXYZDUV2 *vb,
...
+                unsigned char height_6 = data[index_3 + 1];
+                unsigned char height_7 = data[index_4 + 1];
+                unsigned char height_8 = data[index_4];
+                float diff_1 = height_2 - height_3;
+                float diff_2 = height_3 - height_1;
+                float diff_3 = height_8 - height_2;
+                float diff_4 = height_2 - height_5;
+                float diff_5 = height_3 - height_6;
+                float diff_6 = height_7 - height_2;
+                float interpolated_diff_x = (1.0f - fract_x) * diff_
```
xezon: Despite this not compiling now, I don't like how this code was refactored beyond the name changes. it is difficult to tell whether or not it will behave the same.

xezon (dismissed):
Name changes need to not further refactor code.

src/game/client/baseheightmap.cpp
bobtista: Good catch, I re-asserted to gpt that it should not refactor anything. It should be back to identical logic now, with the only change being the if (pos) check being if (pos != nullptr)

src/game/client/baseheightmap.cpp
```diff
@@ -968,89 +968,93 @@ void BaseHeightMapRenderObjClass::Do_The_Light(VertexFormatXYZDUV2 *vb,
...
+        float floor_y = GameMath::Fast_Float_Floor(scaled_y);
+        float fract_x = scaled_x - floor_x;
+        float fract_y = scaled_y - floor_y;
+        int index_x = height_map->Border_Size() + GameMath::Lrintf(floor_x);
+        int index_y = height_map->Border_Size() + GameMath::Lrintf(floor_y);
+        int x_extent = height_map->Get_X_Extent();
+
+        // Check if the indices are within valid bounds
+        if (index_x <= x_extent - 3 && index_y <= height_map->Get_Y_Extent() - 3 && index_y >= 1 && index_x >= 1) {
+            unsigned char *data = height_map->Get_Data_Ptr();
```
xezon: Perhaps call it height_data

src/game/client/baseheightmap.cpp
```diff
@@ -968,89 +968,93 @@ void BaseHeightMapRenderObjClass::Do_The_Light(VertexFormatXYZDUV2 *vb,
...
+
+            // Calculate the height based on interpolation
+            if (fract_y <= fract_x) {
+                float height_3 = data[index_1 + 1];
+                interpolated_height = ((height_2 - height_3) * fract_y + height_3 + (1.0f - fract_x) * (height_1 - height_3)) * HEIGHTMAP_SCALE;
             } else {
-                float f10 = data[x_extent + i3];
-                f9 = ((1.0f - f6) * (f7 - f10) + f10 + (f8 - f10) * f5) * HEIGHTMAP_SCALE;
+                float height_3 = data[x_extent + index_1];
+                interpolated_height = ((1.0f - fract_y) * (height_1 - height_3) + height_3 + (height_2 - height_3) * fract_x) * HEIGHTMAP_SCALE;
```
xezon: Any chance we can come up with better names than height_1, height_2, height_3 etc? New names are better, but still leave questions.

I do not have better name suggestions at this time. Code is hard to understand.

src/game/client/baseheightmap.cpp
```diff
@@ -966,91 +966,96 @@ void BaseHeightMapRenderObjClass::Do_The_Light(VertexFormatXYZDUV2 *vb,
         alpha);
 }
 
-float BaseHeightMapRenderObjClass::Get_Height_Map_Height(float x, float y, Coord3D *pos) const
+float BaseHeightMapRenderObjClass::Get_Height_Map_Height(float input_x, float input_y, Coord3D *pos) const
```
xezon: Perhaps can leave x, y as is. It is short and also in the declaration of the function. Generally if function definition argument names are changed, then those of function declaration should be changed too.

bobtista: Done. Also formatted.

src/game/client/baseheightmap.cpp
```diff
@@ -968,89 +968,100 @@ void BaseHeightMapRenderObjClass::Do_The_Light(VertexFormatXYZDUV2 *vb,
...
+                int next_row_cell_index = map_width + current_cell_index;
+                int next_next_row_cell_index = map_width * (index_y + 2) + index_x;
+                unsigned char current_cell_height_data = height_data[current_cell_index];
+                unsigned char right_neighbor_height_data = height_data[current_cell_index + 1];
+                unsigned char bottom_right_neighbor_height_data = height_data[map_width + 1 + current_cell_index];
+                unsigned char bottom_neighbor_height_data = height_data[map_width + current_cell_index];
+                unsigned char top_neighbor_height_data = height_data[prev_row_cell_index];
+                unsigned char top_right_neighbor_height_data = height_data[prev_row_cell_index + 1];
+                unsigned char down_two_right_neighbor_height_data = height_data[next_next_row_cell_index + 1];
+                unsigned char down_two_neighbor_height_data = height_data[next_next_row_cell_index];
```
xezon: Could you explain how you came up with these names? Perhaps visualize it?

src/game/client/baseheightmap.cpp
```diff
@@ -968,89 +968,100 @@ void BaseHeightMapRenderObjClass::Do_The_Light(VertexFormatXYZDUV2 *vb,
...
+        int index_x = height_map->Border_Size() + GameMath::Lrintf(floor_x);
+        int index_y = height_map->Border_Size() + GameMath::Lrintf(floor_y);
+        int map_width = height_map->Get_X_Extent();
+
+        // Check if the indices are within valid bounds
+        if (index_x <= map_width - 3 && index_y <= height_map->Get_Y_Extent() - 3 && index_y >= 1 && index_x >= 1) {
+            unsigned char *height_data = height_map->Get_Data_Ptr();
+            int current_point_index = map_width * index_y + index_x;
+            float top_left_height = height_data[current_point_index];
+            float bottom_right_height = height_data[map_width + 1 + current_point_index];
```
xezon: You have written all variables assuming the origin is top left correct? Are you sure the origin is not at bottom left? I think it is there, though would need to double check as well.

xezon (commented):
Just realized I did not submit my pending review.

src/game/client/baseheightmap.cpp
bobtista: Yes, I was under the assumption that 0,0 was top left. As I understand it, map_width * index_y + index_x is the index for (x, y). Rather than using a 2D array, we think of map_width as one row worth of items, so y*number_of_rows gets us which row we're in, and +/- x moves us around within a given row. I'll rewrite it so 0,0 is bottom left.

src/game/client/baseheightmap.cpp
bobtista: GPT said it should be 0,0 is top left, but it also kept trying to name the neighbors incorrectly and inconsistently, so I don't have much confidence in what it thought was going on. I'll rewrite it so 0,0 is bottom left. I've heard that's what the debugger shows anyway

src/game/client/baseheightmap.cpp
bobtista: Done

xezon (approved):
Let us submit that.

PR #919, opened 2023-05-18
Enable warnings for wrong order of constructor initialization and for variable shadowing

jonwil:
overrides previous variable of the same name" and fix all the results.


xezon:
Perhaps in future we can also enable this?
https://learn.microsoft.com/en-us/cpp/code-quality/c26495?view=msvc-170

Though not sure what happens to POD structs. Some structs should not have initialization.

src/w3d/renderer/boxrobj.cpp
```diff
@@ -117,7 +117,7 @@ BoxRenderObjClass::BoxRenderObjClass(const W3dBoxStruct &src) : m_opacity(0.25)
 }
 
 BoxRenderObjClass::BoxRenderObjClass() :
-    m_color(1, 1, 1), m_opacity(0.25), m_objSpaceCenter(0, 0, 0), m_objSpaceExtent(1, 1, 1)
+    m_color(1, 1, 1), m_objSpaceCenter(0, 0, 0), m_objSpaceExtent(1, 1, 1), m_opacity(0.25)
```
xezon: f

xezon (approved):

PR #920, opened 2023-05-20
Add Is_Nan function to build with/without gamemath enabled

abramann:


xezon (approved):

src/w3d/math/gamemath.h
```diff
@@ -547,4 +530,31 @@ inline int Lrintf(float val)
 //    return true;
 //}
 
+inline bool Is_Nan(float val)
+{
+#ifdef BUILD_WITH_GAMEMATH
+    return gm_isnanf(val);
+#else
+    return isnan(val);
```
xezon: Something fails to compile here. I suggest using std::isnan.

src/w3d/math/gamemath.h
```diff
@@ -547,4 +530,31 @@ inline int Lrintf(float val)
 //    return true;
 //}
 
+inline bool Is_Nan(float val)
+{
+#ifdef BUILD_WITH_GAMEMATH
+    return gm_isnanf(val);
```
xezon:
```
D:\a\Thyme\Thyme\src\w3d\math\gamemath.h(536): error C4800: Implicit conversion from 'int' to bool. Possible information loss
D:\a\Thyme\Thyme\src\w3d\math\gamemath.h(536): note: consider using explicit cast or comparison to 0 to avoid this warning
D:\a\Thyme\Thyme\src\w3d\math\gamemath.h(561): fatal error C1004: unexpected end-of-file found
```

abramann: I compile with MSVC v140 and no such errors,.

xezon: See build below. Compile errors on some targets.

abramann:
It should build ok now. Can you check?

xezon:
Still errors with
```
D:\a\Thyme\Thyme\src\w3d\math\gamemath.h(536): error C4800: Implicit conversion from 'int' to bool. Possible information loss
D:\a\Thyme\Thyme\src\w3d\math\gamemath.h(536): note: consider using explicit cast or comparison to 0 to avoid this warning
```

src/w3d/math/gamemath.h
```diff
@@ -533,7 +533,7 @@ inline int Lrintf(float val)
 inline bool Is_Nan(float val)
 {
 #ifdef BUILD_WITH_GAMEMATH
-    return gm_isnanf(val);
+    return static_cast<bool>(gm_isnanf(val));
```
xezon: If `gm_isnanf(val)` returns int, then I would write it as `return gm_isnanf(val) != 0;`

xezon:
There is formatting issue left. Make sure to format your code with the provided .clang-format at the root of this project.

abramann:
Seems build (gcc, ubuntu-20.04) has been stuck.

Issue #888, opened 2023-04-01
GLA Bomb Truck can no longer attack


xezon:
GLA Bomb Truck can no longer attack. When disguised and parked close to enemy unit, the Truck will not even undisguise.

Happens with 3018c18860270f436edc42c5fe3336a98e0ed994

![shot_20230401_231645_1](https://user-images.githubusercontent.com/4720891/229314960-ac614522-904d-4f28-89d1-44c8dd817f56.jpg)

abramann:
Terrorist is same.

xezon:
I narrowed it to `WeaponTemplate::Fire_Weapon_Template`. Everything else for WeaponTemplate, WeaponStore and Weapon is unhooked.

PR #928, opened 2023-05-24
Fix errors in Weapon::Get_Percent_Ready_To_Fire

### LINKED ISSUE #916: Assert in Weapon::Get_Percent_Ready_To_Fire() is hit after using clear mines ability
Assert is hit after using clear mines ability:

```
captainslog_dbgassert(time_left <= total_time, "time_left <= total_time");
```

```
>	thyme.dll!Weapon::Get_Percent_Ready_To_Fire() Line 1026	C++
 	game.dat!005a45c4()	Unknown
 	game.dat![Frames below may be incorrect and/or missing, no symbols loaded for game.dat]	Unknown
 	game.dat!005a3cf9()	Unknown
 	game.dat!0045e9e8()	Unknown
 	game.dat!0050a6fa()	Unknown
 	game.dat!004ad301()	Unknown
 	thyme.dll!GameEngine::Update() Line 176	C++
 	thyme.dll!Win32GameEngine::Update() Line 58	C++
 	thyme.dll!GameEngine::Execute() Line 556	C++
 	thyme.dll!Game_Main(int argc, char * * argv) Line 41	C++
 	thyme.dll!main(int argc, char * * argv) Line 700	C++
 	thyme.dll!Main_Func(HINSTANCE__ * hInstance, HINSTANCE__ * hPrevInstance, char * lpCmdLine, int nCmdShow) Line 728	C++
```

bobtista:
* Fixes #916

### DIFF
--- a/src/game/logic/object/weapon.cpp
+++ b/src/game/logic/object/weapon.cpp
@@ -1017,25 +1017,25 @@ float Weapon::Get_Percent_Ready_To_Fire() const
             unsigned int next_shot = Get_Next_Shot();
             captainslog_dbgassert(now >= m_whenLastReloadStarted, "now >= m_whenLastReloadStarted");
 
-            if (now < next_shot) {
-                captainslog_dbgassert(next_shot >= m_whenLastReloadStarted, "next_shot >= m_whenLastReloadStarted");
-                unsigned int time_left = next_shot - m_whenLastReloadStarted;
-
-                if (time_left != 0) {
-                    unsigned int total_time = next_shot - now;
-                    captainslog_dbgassert(time_left <= total_time, "time_left <= total_time");
-
-                    if (time_left - total_time < time_left) {
-                        return (float)(time_left - total_time) / (float)time_left;
-                    } else {
-                        return 1.0f;
-                    }
-                } else {
-                    return 1.0f;
-                }
-            } else {
+            if (now >= next_shot) {
                 return 1.0f;
             }
+
+            captainslog_dbgassert(next_shot >= m_whenLastReloadStarted, "next_shot >= m_whenLastReloadStarted");
+            unsigned int total_time = next_shot - m_whenLastReloadStarted;
+
+            if (total_time == 0) {
+                return 1.0f;
+            }
+
+            unsigned int time_left = next_shot - now;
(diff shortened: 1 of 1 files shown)


src/game/logic/object/weapon.cpp
```diff
@@ -1019,10 +1019,10 @@ float Weapon::Get_Percent_Ready_To_Fire() const
 
             if (now < next_shot) {
                 captainslog_dbgassert(next_shot >= m_whenLastReloadStarted, "next_shot >= m_whenLastReloadStarted");
-                unsigned int time_left = next_shot - m_whenLastReloadStarted;
```
bobtista, line 1022: Should we also replace m_whenLastReloadStarted with the getter Get_Last_Reload_Started()?

jonwil:
Having reviewed the original code, the correct fix is to rename time_left to total_time and total_time to time_left.
So
unsigned int time_left = next_shot - m_whenLastReloadStarted; 
becomes
unsigned int total_time = next_shot - m_whenLastReloadStarted;
and
unsigned int total_time = next_shot - m_whenLastReloadStarted;
becomes
unsigned int time_left = next_shot - m_whenLastReloadStarted;

and the other uses of both variables get swapped as well.

And no we should not use the getter in this case.

src/game/logic/object/weapon.cpp
tomsons26: original code didn't and seems they followed a rule to avoid getters inside classes that define the member

bobtista:
@jonwil do you mean that total_time should be defined first? Like
`unsigned int total_time = next_shot - m_whenLastReloadStarted;

                if (time_left != 0) {
                    unsigned int time_left = next_shot - now;`
If so, I should change the check to if (total_time != 0)
Or, I could both rename and swap the order of the definitions such that the check if (time_left != 0) stays the same

xezon:
Does change fix #916 confirmed?

Does it perhaps also fix #888?

jonwil (approved):
Verified that the change in this PR is correct. Approved.

xezon:
I will check later if this fixes the bugs.

xezon:
* Fixes #916
* Does not fix #888

Issue #915, opened 2023-05-13
Obsolete code in AIGroup::Group_Tighten_To_Position()


xezon:
Call to Get_Min_Max_And_Center appears to be obsolete.

```c++
void AIGroup::Group_Tighten_To_Position(const Coord3D *target_pos, bool append, CommandSourceType cmd_source)
{
    if (cmd_source == COMMANDSOURCE_PLAYER && g_theWriteableGlobalData->m_groupMoveClickToGatherAreaFactor > 0.0f) {
        Coord2D min;
        Coord2D max;
        Coord3D center;
        Get_Min_Max_And_Center(&min, &max, &center);
    }
...
```

PR #937, opened 2023-05-28
Removes obsolete code and adds comments in AIGroup::Group_Tighten_To_Position

bobtista:
Addresses https://github.com/TheAssemblyArmada/Thyme/issues/915

### DIFF
--- a/src/game/logic/ai/aigroup.cpp
+++ b/src/game/logic/ai/aigroup.cpp
@@ -1641,14 +1641,18 @@ void Get_Helicopter_Offset(Coord3D &pos, int count)
     }
 }
 
+/**
+Group_Tighten_To_Position - Performs tightening of the group members towards a target position.
+This method iterates over the group members, calculates their distances to the target position,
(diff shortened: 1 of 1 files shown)


src/game/logic/ai/aigroup.cpp
```diff
@@ -1686,6 +1694,7 @@ void AIGroup::Group_Tighten_To_Position(const Coord3D *target_pos, bool append,
         }
     }
 
+    // Update the positions of the members in the Pathfinder
```
xezon: I think these kind of in-function comments are too much. The code here is rather self-explanatory. The comment outside the function is good enough.

In-function code comments are useful when something not so obvious needs explanation - or in our case, when we deviate from original functionality. However, even that can be documented by code, for example by putting new code into a unique namespace, or by guarding code inside a pre-processor macro.

xezon:
Change looks ok. Does not need to be merged before Standalone.

PR #953, opened 2023-06-09
Add correct enum names for ProductionID, ObjectID, DrawableID, WaypointID and FormationID based on debug strings in BFME2. This is a simple renaming of an enum so should require minimal testing.

jonwil:


src/game/client/draw/drawable.cpp
```diff
@@ -1577,7 +1577,7 @@ void Drawable::Xfer_Snapshot(Xfer *xfer)
     if (m_object != nullptr) {
         object_id = m_object->Get_ID();
     } else {
-        object_id = OBJECT_UNK;
+        object_id = INVALID_ID;
```
xezon: INVALID_ID reads ambiguous. Perhaps call this INVALID_OBJECT_ID?

tomsons26: well indeed its INVALID_ID
BFME1 also shows it

tomsons26: while probably should go with INVALID_OBJECT_ID for consistency with others

jonwil: Fixed.

xezon (approved):

PR #972, opened 2023-08-08
Use override keyword when possible

feliwir:
- Remove `virtual` keyword when not required
- Use `override` keyword when a function overrides


xezon:
I do not like this. I personally would like it to keep virtual always.

xezon:
But adding `override` everywhere necessary is good.

feliwir:
I closed the PRs for now since they‘re pretty outdated. My focus for now is to get standalone working (to some degree). 
I’ll pick those up again later

PR #975, opened 2023-09-09
Use blocking event loop in WDump

feliwir:
The loop would fully load the current thread. There's no need to use a non-blocking event loop in WDump

### DIFF
--- a/src/tools/wdump/wdump.cpp
+++ b/src/tools/wdump/wdump.cpp
@@ -4126,7 +4126,7 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
     MSG msg = {};
 
     while (WM_QUIT != msg.message) {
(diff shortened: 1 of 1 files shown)


src/tools/wdump/wdump.cpp
```diff
@@ -4126,7 +4126,7 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
     MSG msg = {};
 
     while (WM_QUIT != msg.message) {
-        if (PeekMessage(&msg, nullptr, 0, 0, PM_REMOVE)) {
+        if (GetMessage(&msg, nullptr, 0, 0)) {
```
xezon, line 4129: Perhaps add a comment so that future engineer does not change it back to `PeekMessage`, because it mismatches decomp.

jonwil:
WDump isn't an exact decompilation match. The wdump we have is essentially taken from the Renegade work and thrown in there (the original WW wdump is written with MFC)

xezon (approved):

PR #978, opened 2023-09-11
Load freetype font face

feliwir:
- [x] Use fontconfig to locate fonts on all platforms (enabled on non-windows by default)
- [x] Use  windows registry to locate fonts, if fontconfig disabled
- [x]  Use freetype to rasterize fonts and get font metrics
image


src/w3d/renderer/render2dsentence.cpp
```diff
@@ -42,13 +52,39 @@ FontCharsClass::FontCharsClass() :
     m_gdiFont(nullptr),
     m_gdiBitmapBits(nullptr),
     m_memDC(nullptr),
+#endif
+#ifdef BUILD_WITH_FREETYPE
+    m_ftLibrary(nullptr),
```
xezon: Perhaps just use decl init, so you can save on inserting ifdef here.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -42,13 +52,39 @@ FontCharsClass::FontCharsClass() :
...
+        }
+    } else {
+        captainslog_error("Failed to init Fontconfig");
+    }
+#endif
+    // Use Freetype to loader & rasterize fonts
+#ifdef BUILD_WITH_FREETYPE
+    FT_Error error = FT_Init_FreeType(&m_ftLibrary);
+    if (error != FT_Err_Ok) {
+#ifdef FT_CONFIG_OPTION_ERROR_STRINGS
```
xezon: Is there a way to get rid of this ifdef? It is quite a bit too much configurability. Several times in this file.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -42,13 +52,39 @@ FontCharsClass::FontCharsClass() :
...
     m_lastUnicodeChar(0),
     m_isBold(false)
 {
     std::memset(m_asciiCharArray, 0, sizeof(m_asciiCharArray));
+    // Use Fontconfig to locate system fonts
+#ifdef BUILD_WITH_FONTCONFIG
+    if (FcInit() == FcTrue) {
+        m_fc = FcInitLoadConfigAndFonts();
+        if (m_fc == nullptr) {
+            captainslog_error("Failed to load Fontconfig config & fonts ");
```
xezon: fonts"

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,8 +306,149 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+    FcConfigSubstitute(m_fc, pattern, FcMatchPattern);
+    FcDefaultSubstitute(pattern);
+
+    // Find the font
+    FcResult result;
+    FcPattern *font = FcFontMatch(m_fc, pattern, &result);
+    if (font != nullptr) {
+        FcChar8 *file = NULL;
+        if (FcPatternGetString(font, FC_FILE, 0, &file) == FcResultMatch) {
+            // Save the file to another std::string
```
xezon: Comment does not help. Can be removed.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,8 +306,149 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+            font_file_path = (char *)file;
```
xezon: reinterpret_cast?

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,8 +306,149 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+    HKEY hkey;
+    LONG result;
+
+    // Open Windows font registry key
+    result = RegOpenKeyEx(HKEY_LOCAL_MACHINE, font_registry_key, 0, KEY_READ, &hkey);
+    if (result != ERROR_SUCCESS) {
+        return false;
+    }
+
+    DWORD max_value_name_size, max_value_data_size;
```
xezon: Prefer 2 lines.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,8 +306,149 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
     }
 }
 
+bool FontCharsClass::Locate_Font(const char *font_name, StringClass &font_file_path)
```
xezon: I suggest to make this function call 2 sub functions

Locate_Freetype_Font
Locate_Windows_Font

This way it becomes slimmer.

```c++
void FontCharsClass::Locate_Font(...)
{
#if defined(BUILD_WITH_FREETYPE)
  Locate_Freetype_Font(...);
#else defined(PLATFORM_WINDOWS)
  Locate_Windows_Font(...)
#endif
}
```

xezon: Is this function often or rarely called? It looks expensive.

If this function is expensive and called many times, I suggest to build font path catalogue once, and then lookup from that catalogue afterwards. It should bring down computation time.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,8 +306,149 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+    // Open Windows font registry key
+    result = RegOpenKeyEx(HKEY_LOCAL_MACHINE, font_registry_key, 0, KEY_READ, &hkey);
+    if (result != ERROR_SUCCESS) {
+        return false;
+    }
+
+    DWORD max_value_name_size, max_value_data_size;
+    result = RegQueryInfoKey(hkey, 0, 0, 0, 0, 0, 0, 0, &max_value_name_size, &max_value_data_size, 0, 0);
+    if (result != ERROR_SUCCESS) {
+        return false;
```
xezon, line 380: `hkey` is not closed here.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,8 +306,149 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+
+    RegCloseKey(hkey);
+
+    if (font_file.Is_Empty()) {
+        return false;
+    }
+
+    // Build full font file path
+    char windows_dir[MAX_PATH];
+    if (GetWindowsDirectory(windows_dir, MAX_PATH) == 0) {
```
xezon: Perhaps do this test much earlier? No need to do all this stuff above when this already fails.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -274,15 +508,23 @@ void FontCharsClass::Create_GDI_Font(const char *font_name)
...
+#endif
     if (is_generals) {
         m_overhang = 0;
     }
-#endif
 }
 
 void FontCharsClass::Free_GDI_Font()
 {
+#ifdef BUILD_WITH_FREETYPE
```
xezon, line 567: I suggest to make this function call 2 sub functions

Free_Windows_Font
Free_Freetype_Font

This way it becomes slimmer.

```c++
void FontCharsClass::Free_GDI_Font()
{
#if defined(BUILD_WITH_FREETYPE)
  Free_Freetype_Font();
#else defined(PLATFORM_WINDOWS)
  Free_Windows_Font()
#endif
}
```

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,8 +306,149 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+    font_file_path += font_file;
+    return true;
+#else
+    return false;
+#endif
+}
+
 void FontCharsClass::Create_GDI_Font(const char *font_name)
 {
+    bool is_generals = false;
```
xezon, line 447: I suggest to make this function call 2 sub functions

Create_Freetype_Font
Create_Windows_Font

This way it becomes slimmer.

```c++
void FontCharsClass::Create_GDI_Font(...)
{
#if defined(BUILD_WITH_FREETYPE)
  Create_Freetype_Font(...);
#else defined(PLATFORM_WINDOWS)
  Create_Windows_Font(...)
#endif
}
```

This way you also do not need to move the `is_generals` part from original code.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,8 +306,149 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+    StringClass font_file;
+
+    // Look for a matching font name
+    do {
+        value_data_size = max_value_data_size;
+        value_name_size = max_value_name_size;
+
+        result = RegEnumValue(hkey, value_index, value_name, &value_name_size, 0, &value_type, value_data, &value_data_size);
+
+        value_index++;
```
xezon: Prefer preincrement over postincrement.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,8 +306,149 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+        if (FcPatternGetString(font, FC_FILE, 0, &file) == FcResultMatch) {
+            // Save the file to another std::string
+            font_file_path = (char *)file;
+        }
+        FcPatternDestroy(font);
+    }
+
+    FcPatternDestroy(pattern);
+    return result == FcResultMatch;
+#elif defined PLATFORM_WINDOWS
```
xezon: What is this locator for? It is not used on Windows or is it?

src/w3d/renderer/render2dsentence.cpp
feliwir: It is called once per Family + Size combination (e.g. "Arial" & 12).  The fonts get cached afterwards.

The `FontConfig` stuff should be pretty "cheap". Regarding the Windows registry lookup: Ideally we want to use `FontConfig` later aswell, however their buildsystem is not really friendly on windows (= no CMake, only `Meson` or `Autotools`)

src/w3d/renderer/render2dsentence.cpp
feliwir: I find `Windows_Font` misleading. There is `GDI` or `DirectWrite`. I'll add a new function `Free_Freetype_Font` and call both `Free_GDI_Font` and `Free_Freetype_Font`

src/w3d/renderer/render2dsentence.cpp
feliwir: `FontConfig` locator is only used in windows when explicity enabled (since it's a bit cumbersome to build on windows). As a fallback we use the registry to find out our font paths.

src/w3d/renderer/render2dsentence.cpp
feliwir: I don't really know how. We'd need to abstract all of this and split the fontrendering in multiple classes. However i don't think that this works well, without having all surrounding code

src/w3d/renderer/render2dsentence.cpp
xezon: There are a multitude of font and size combinations in the game. A font path catalogue may be useful.

src/w3d/renderer/render2dsentence.cpp
xezon: Well how about:

```c++
void LogFtError(const char *text, FT_Error error)
{
#ifdef FT_CONFIG_OPTION_ERROR_STRINGS
  captainslog_error("%s: %s", text, FT_Error_String(error));
#else
  captainslog_error(text);
#endif
}
```
This would get rid of bulky code.

src/w3d/renderer/render2dsentence.cpp
feliwir: That's exactly what `FontConfig` does internally. See https://www.freedesktop.org/wiki/Software/fontconfig/

src/w3d/renderer/render2dsentence.cpp
xezon: Would be nice. All this `#ifdef` here is difficult to look at.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -134,12 +179,81 @@ void FontCharsClass::Blit_Char(unichar_t ch, uint16_t *dest_ptr, int dest_stride
...
+    }
+    // convert to an anti-aliased bitmap
+    error = FT_Render_Glyph(m_ftFace->glyph, FT_RENDER_MODE_NORMAL);
+    if (error) {
+        captainslog_error("Failed to render character(%i) from font", ch);
+        return nullptr;
+    }
+    int x_pos = 0;
+    if (ch == 'W') {
+        x_pos = 1;
```
xezon, line 206: Why this? Because original does it in the GDI function?

xezon:
How to enable this?

Turning on USE_FREETYPE, USE_FONTCONFIG will produce cmake error:

```
Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE) 
CMake Warning at .solution/vs2019/_deps/freetype-src/builds/cmake/FindHarfBuzz.cmake:99 (message):
  Required version (2.0.0) is higher than found version ()
Call Stack (most recent call first):
  .solution/vs2019/_deps/freetype-src/CMakeLists.txt:258 (find_package)


Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR) 
Could NOT find PNG (missing: PNG_LIBRARY PNG_PNG_INCLUDE_DIR) 
Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR) 
Could NOT find BZip2 (missing: BZIP2_LIBRARIES BZIP2_INCLUDE_DIR) 
Could NOT find BrotliDec (missing: BROTLIDEC_INCLUDE_DIRS BROTLIDEC_LIBRARIES) 
CMake Error at tools/.cmake/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Fontconfig (missing: Fontconfig_LIBRARY
  Fontconfig_INCLUDE_DIR)
Call Stack (most recent call first):
  tools/.cmake/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  tools/.cmake/share/cmake-3.21/Modules/FindFontconfig.cmake:78 (find_package_handle_standard_args)
  CMakeLists.txt:341 (find_package)
```

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -134,12 +179,81 @@ void FontCharsClass::Blit_Char(unichar_t ch, uint16_t *dest_ptr, int dest_stride
...
 
+#if defined BUILD_WITH_FREETYPE
+const FontCharsClass::CharDataStruct *FontCharsClass::Store_Freetype_Char(unichar_t ch)
+{
+    captainslog_assert(m_ftFace != nullptr);
+    FT_UInt glyph_index = FT_Get_Char_Index(m_ftFace, ch);
+    // load glyph image into the slot (erase previous one)
+    FT_Error error = FT_Load_Glyph(m_ftFace, glyph_index, FT_LOAD_DEFAULT);
+    if (error) {
+        captainslog_error("Failed to load character(%i) from font", ch);
```
xezon, line 195: `%lc`, at least on Windows.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -134,12 +179,81 @@ void FontCharsClass::Blit_Char(unichar_t ch, uint16_t *dest_ptr, int dest_stride
...
+        return nullptr;
+    }
+    // convert to an anti-aliased bitmap
+    error = FT_Render_Glyph(m_ftFace->glyph, FT_RENDER_MODE_NORMAL);
+    if (error) {
+        captainslog_error("Failed to render character(%i) from font", ch);
```
xezon, line 201: `%lc`, at least on Windows.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -134,12 +179,81 @@ void FontCharsClass::Blit_Char(unichar_t ch, uint16_t *dest_ptr, int dest_stride
...
+    curr_buffer += m_currPixelOffset;
+    // We might need a memset
+
+    int x_offset = m_ftFace->glyph->bitmap_left;
+    int descent = m_charHeight - m_ascent;
+    int y_offset = (m_charHeight - m_ftFace->glyph->bitmap_top) - descent;
+
+    // Render the bitmap
+    for (unsigned int row = 0; row < m_ftFace->glyph->bitmap.rows; row++) {
+        int index = row * m_ftFace->glyph->bitmap.pitch;
```
xezon, line 226: unsigned

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -134,12 +179,81 @@ void FontCharsClass::Blit_Char(unichar_t ch, uint16_t *dest_ptr, int dest_stride
...
+        int dst_index = (y_offset + row) * char_width;
```
xezon, line 227: unsigned, probably

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,6 +321,171 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+
+    FcPatternDestroy(pattern);
+    return result == FcResultMatch;
+}
+#endif
+
+#if defined PLATFORM_WINDOWS
+bool FontCharsClass::Locate_Font_WinRegistry(const char *font_name, StringClass &font_file_path)
+{
+    char *font_registry_key = "Software\\Microsoft\\Windows NT\\CurrentVersion\\Fonts";
```
xezon: const char*

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,6 +321,171 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+{
+    bool is_generals = false;
+
+    if (std::strcmp(font_name, "Generals") == 0) {
+        font_name = "Arial";
+        is_generals = true;
+    }
+
+    StringClass font_file_path;
+    if (!Locate_Font(font_name, font_file_path)) {
```
xezon, line 455: Should this not just call `Locate_Font_FontConfig` directly?

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,6 +321,171 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+    if (error != FT_Err_Ok) {
+#ifdef FT_CONFIG_OPTION_ERROR_STRINGS
+        captainslog_error("Failed to set character size: %s", FT_Error_String(error));
+#else
+        captainslog_error("Failed to set character size");
+#endif
+        return;
+    }
+
+    // Use the same metric calculation as Wine does, in their GDI emulation layer
```
xezon: Can remove comma.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -305,13 +596,14 @@ void FontCharsClass::Free_GDI_Font()
...
 {
-#ifdef PLATFORM_WINDOWS
     m_name.Format("%s%d", font_name, point_size);
     m_gdiFontName = font_name;
     m_pointSize = point_size;
     m_isBold = is_bold;
-    Create_GDI_Font(font_name);
+#ifdef BUILD_WITH_FREETYPE
+    Create_Freetype_Font(font_name);
 #endif
```
xezon, line 609: `elif defined(PLATFORM_WINDOWS)` ? Does this create 2 fonts on Windows now? Should `PLATFORM_WINDOWS` and `BUILD_WITH_FREETYPE` be mutually exclusive or not?

`Create_GDI_Font`, `Free_GDI_Font`, `Store_GDI_Char` functions could be compiled out entirely in non Windows. Currently they exist in all platforms, but will have empty function bodies.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,6 +321,171 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+
+    // Use the same metric calculation as Wine does, in their GDI emulation layer
+    // See https://github.com/NVIDIA/winex_lgpl/blob/master/winex/dlls/gdi/freetype.c#L3027
+    if (FT_IS_SCALABLE(m_ftFace)) {
+        m_ascent = FT_MulFix(m_ftFace->ascender, m_ftFace->size->metrics.y_scale) >> 6;
+        int descent = -FT_MulFix(m_ftFace->descender, m_ftFace->size->metrics.y_scale) >> 6;
+        m_charHeight = m_ascent + descent;
+        m_overhang = 0;
+    } else {
+        captainslog_error("Require a scalable font");
```
xezon: I suggest to also print the offending font name in the error message.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,6 +321,171 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+
+    // Original calculates with 96 DPI. Do the same
+    int font_height = FT_MulDiv(m_pointSize, 96, 72);
+    m_widthReduction = std::clamp(font_height / 8, 0, 4);
+    error = FT_Set_Pixel_Sizes(m_ftFace, 0, font_height);
+    if (error != FT_Err_Ok) {
+#ifdef FT_CONFIG_OPTION_ERROR_STRINGS
+        captainslog_error("Failed to set character size: %s", FT_Error_String(error));
+#else
+        captainslog_error("Failed to set character size");
```
xezon: I suggest to also print the offending font name in the error message.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,6 +321,171 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+            captainslog_error("Failed to find fallback font: %s", font_name);
+            return;
+        }
+    }
+    FT_Error error = FT_New_Face(m_ftLibrary, font_file_path, 0, &m_ftFace);
+    if (error != FT_Err_Ok) {
+#ifdef FT_CONFIG_OPTION_ERROR_STRINGS
+        captainslog_error("Failed to load Freetype face from file: %s", FT_Error_String(error));
+#else
+        captainslog_error("Failed to load Freetype face from file");
```
xezon: I suggest to also print the offending font name in the error message.

src/w3d/renderer/render2dsentence.cpp
```diff
@@ -207,6 +321,171 @@ void FontCharsClass::Update_Current_Buffer(int char_width)
...
+
+    if (std::strcmp(font_name, "Generals") == 0) {
+        font_name = "Arial";
+        is_generals = true;
+    }
+
+    StringClass font_file_path;
+    if (!Locate_Font(font_name, font_file_path)) {
+        captainslog_error("Failed to locate font: %s", font_name);
+        captainslog_info("Trying Arial as a fallback");
```
xezon: Better be:

```c++
font_name = "Arial";
captainslog_info("Trying '%s' as a fallback", font_name);
if (Locate_Font(font_name, font_file_path)) {
```

Right now there is mismatch in messages.

feliwir:
> How to enable this?
> 
> Turning on USE_FREETYPE, USE_FONTCONFIG will produce cmake error:
> 
> ```
> Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE) 
> CMake Warning at .solution/vs2019/_deps/freetype-src/builds/cmake/FindHarfBuzz.cmake:99 (message):
>   Required version (2.0.0) is higher than found version ()
> Call Stack (most recent call first):
>   .solution/vs2019/_deps/freetype-src/CMakeLists.txt:258 (find_package)
> 
> 
> Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR) 
> Could NOT find PNG (missing: PNG_LIBRARY PNG_PNG_INCLUDE_DIR) 
> Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR) 
> Could NOT find BZip2 (missing: BZIP2_LIBRARIES BZIP2_INCLUDE_DIR) 
> Could NOT find BrotliDec (missing: BROTLIDEC_INCLUDE_DIRS BROTLIDEC_LIBRARIES) 
> CMake Error at tools/.cmake/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
>   Could NOT find Fontconfig (missing: Fontconfig_LIBRARY
>   Fontconfig_INCLUDE_DIR)
> Call Stack (most recent call first):
>   tools/.cmake/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
>   tools/.cmake/share/cmake-3.21/Modules/FindFontconfig.cmake:78 (find_package_handle_standard_args)
>   CMakeLists.txt:341 (find_package)
> ```

You need to install `FontConfig` on your windows system. It does not use `CMake`, so it's not super simple to build. You can try using `VCPKG`: 
`.\vcpkg install fontconfig:x86-windows`

src/w3d/renderer/render2dsentence.cpp
feliwir: Yes. It makes the 'W' char one pixel larger

src/w3d/renderer/render2dsentence.cpp
feliwir: No. Pitch is a signed integer and index can be negative

src/w3d/renderer/render2dsentence.cpp
feliwir: No, why? It works with both versions. E.g. `FreeType` works without `FontConfig`

src/w3d/renderer/render2dsentence.cpp
feliwir: If both are enabled, it will create the windows font handle and the freetype face. For rendering only Freetype will be used

src/w3d/renderer/render2dsentence.cpp
feliwir: Well, if i am using the function above that's a bit hard

xezon:
### Original options menu

options_freetype

### Freetype options menu

The freetype letters are spread wider.
The freetype font is smoother than the original font.

options_original

### Original briefing text

briefing_freetype

### Freetype briefing text

First row of freetype text is not bold. It should be bold.
The freetype font is smoother than the original font.

briefing_original

### Original mission text

mission_original

### Freetype mission text

The freetype font is smoother than the original font.

mission_freetype

xezon (approved):

PR #980, opened 2023-09-12
Various fixes to eliminate use of non-w3d headers from inside w3d code.

jonwil:
Move GameAssetManager to its own file.
Verify and fix all functions in GameMemory.cpp including properly adding Create_W3D_Mem_Pool, Allocate_From_W3D_Mem_Pool and Free_From_W3D_Mem_Pool.
Move array.h to w3d\lib.
Create crcengine.cpp and realcrc.cpp and move things from crc.cpp as needed (and use in w3d where it should be used).
Move some stuff from rtsutils.h to a new rtsutilsw3d.h in w3d\lib so that w3d can use that without dragging in other things. Correctly implement IMPLEMENT_NAMED_W3D_POOL macro.


src/w3d/lib/realcrc.h
```diff
@@ -0,0 +1,21 @@
...
+ *            modify it under the terms of the GNU General Public License
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#pragma once
+
+#include "always.h"
+#include "array.h"
```
xezon, line 18: Is array.h necessary here?

src/game/common/system/mempool.h
```diff
@@ -48,6 +48,7 @@ class MemoryPool
     void Add_To_List(MemoryPool **head);
     void Remove_From_List(MemoryPool **head);
     int Get_Alloc_Size() { return m_allocationSize; }
+    const char *Get_Pool_Name() { return m_poolName; }
```
xezon, line 51: Can be const?

src/game/client/w3dassetmanager.cpp
```diff
@@ -0,0 +1,517 @@
...
+#include "proto.h"
+#include "realcrc.h"
+#include "render2dsentence.h"
+#include "string.h"
+#include "textureloader.h"
+#include "vertmaterial.h"
+#include "w3d_file.h"
+#include "w3dexclusionlist.h"
+#include <cmath>
+#include <stdio.h>
```
xezon, line 39: Did you check if all these headers are still required? I assume it was just copied from other file, which hosts a bit more code unrelated to GameAssetManager.

src/w3d/lib/crcengine.h
```diff
@@ -0,0 +1,29 @@
...
+ *            as published by the Free Software Foundation, either version
+ *            2 of the License, or (at your option) any later version.
+ *            A full copy of the GNU General Public License can be found in
+ *            LICENSE
+ */
+#pragma once
+
+#include "always.h"
+#include "array.h"
+#include "bittype.h"
```
xezon, line 19: This can be removed. always.h already includes it.

xezon (approved):

src/w3d/lib/crcengine.h
xezon: Or remove always.h

PR #981, opened 2023-09-13
Trigger error for unsupported blocks

feliwir:
When a block is not supported in standalone it should point to a nullpointer - otherwise there's no error during parsing but instead a segfault at some random address


xezon (approved):
Squash and merge.

PR #984, opened 2023-09-13
Move some files to the correct source location and rename a few other files.

jonwil:


jonwil:
This is only a simple move/rename of files so will have no impact on functionality.

src/platform/w3dengine/client/w3ddisplay.cpp
```diff
@@ -40,7 +40,7 @@
 #include "network.h"
 #include "particlesysmanager.h"
 #include "physicsupdate.h"
-#include "renderdevicedesc.h"
+#include "rddesc.h"
```
xezon, line 43: Perhaps leave `renderdevicedesc` as is? `rddesc` is not more descriptive.

src/platform/w3dengine/client/drawable/draw/w3dprojectilestreamdraw.h
xezon, line 1: Whats the point of drawable/draw when drawable folder is empty?

xezon (commented):
What about utf8string.h to asciistring.h and utf16string.h to unicodestring.h?

src/platform/w3dengine/client/drawable/draw/w3dprojectilestreamdraw.h
jonwil: I want to stick to the original filenames and paths where we know what they are. drawable probably contained files we don't have yet.

src/platform/w3dengine/client/w3ddisplay.cpp
jonwil: I want to stick to the original filenames and paths where we know what they are.

xezon (approved):

PR #988, opened 2023-09-14
Set default in StartingMoney to false

feliwir:


src/game/common/multiplayersettings.cpp
```diff
@@ -116,6 +116,7 @@ void MultiplayerSettings::Parse_Multiplayer_Starting_Money_Choice_Definition(INI
     captainslog_dbgassert(
         ini->Get_Load_Type() != INI_LOAD_CREATE_OVERRIDES, "Overrides not supported for MultiplayerStartingMoneyChoice");
     StartingMoney money;
+    money.m_default = false;
```
xezon: How about initializing the member to false on `StartingMoney` construction?

xezon (approved):

PR #989, opened 2023-09-14
Implement the rest of Upgrade, UpgradeTemplate and UpgradeCenter.

jonwil:


src/game/common/system/upgrade.cpp
```diff
@@ -102,3 +117,140 @@ bool UpgradeCenter::Can_Afford_Upgrade(Player *player, const UpgradeTemplate *up
...
+    m_buttonImage(nullptr)
+{
+}
+
+int UpgradeTemplate::Calc_Time_To_Build(Player *player) const
+{
+#ifdef GAME_DEBUG_STRUCTS
+    if (player->Is_Instant_Build()) {
+        return 1;
+    } else
```
xezon, line 212: Can remove the `else` to simplify.

src/game/common/system/upgrade.cpp
```diff
@@ -102,3 +117,140 @@ bool UpgradeCenter::Can_Afford_Upgrade(Player *player, const UpgradeTemplate *up
...
+#endif
+    {
+        return m_buildTime * 30.0f;
```
xezon, line 215: This returns `float` on `int`. As designed?

src/game/common/system/upgrade.cpp
```diff
@@ -33,19 +33,23 @@ UpgradeCenter::~UpgradeCenter()
...
 {
-#ifdef GAME_DLL
-    Call_Method<void, UpgradeCenter>(PICK_ADDRESS(0x004AF190, 0x00728DBD), this);
-#endif
+    if (g_theMappedImageCollection != nullptr && !m_buttonImagesCached) {
+        for (UpgradeTemplate *upgrade = m_upgradeList; upgrade != nullptr; upgrade = upgrade->Friend_Get_Next()) {
+            upgrade->Cache_Button_Image();
+        }
+
+        m_buttonImagesCached = true;
```
xezon, line 48: This is surprising. Should the caching not happen in the Init?

xezon (approved):
Tested. Upgrades work normally. Veterancy also works normally.

src/game/common/system/upgrade.cpp
jonwil: This is correct to the original.

src/game/common/system/upgrade.cpp
jonwil: This is correct to the original.

src/game/common/system/upgrade.cpp
jonwil: This is correct to the original.

PR #1037, opened 2023-12-03
Consistently enable INSTALL project in Visual Studio solution

xezon:


xezon:
OmniBlade:
> INSTALL shouldn't run automatically each time the build is redone as part of the default project configuration. It would be the same as a call to make also installing the application at the same time. You could instead create a CMakeUserPresets.json file and set things like that in it and always generate your solutions using your custom preset.

> If you let me know what options you normally configure it for I can probably whip up the presets file for you. https://github.com/TheAssemblyArmada/Vanilla-Conquer/pull/746 has some examples for vanilla conquer that haven't been merged yet due to lack of time and other maintainers to review but allow cleaning up some config in the CMake files.

PR #1105, opened 2024-02-11
Fix error in SysTimeClass::Get

xezon:
* Split off of #796

This claims to be a fix for a Thyme specific issue.

### DIFF
--- a/src/w3d/lib/systimer.cpp
+++ b/src/w3d/lib/systimer.cpp
@@ -38,13 +38,13 @@ int SysTimeClass::Get()
         _is_init = true;
     }
 
-    int time = rts::Get_Time();
+    unsigned int time = rts::Get_Time();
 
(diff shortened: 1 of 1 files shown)


tomsons26 (approved):

tomsons26 (changes requested):
This seems was 
```
int SysTimeClass::Get()
{
#ifdef GAME_DLL
#define _is_init (Make_Global<bool>(0x00A66B30))
#else
    static bool _is_init;
#endif
    if (!_is_init) {
        Reset();
        _is_init = true;
    }

    unsigned time = rts::Get_Time();

    if (time > m_startTime) {
        return time - m_startTime;
    }
    return time + m_negTime;
}
```

xezon:
Fixed.

tomsons26 (approved):

PR #1109, opened 2024-02-11
Fix error in User_Memory_Init_Pools

xezon:
* Split off of #796

This claims to be a fix for a Thyme specific issue.

### DIFF
--- a/src/w3d/renderer/ddsfile.cpp
+++ b/src/w3d/renderer/ddsfile.cpp
@@ -592,7 +592,6 @@ bool DDSFileClass::Load()
 
     if (data == 0) {
         captainslog_warn("DDSFile '%s' appears to contain no data, size is %d.", m_name, fp->Size());
-        fp->Close();
         return false;
     }
 
--- a/src/w3d/renderer/ddsfile.cpp
+++ b/src/w3d/renderer/ddsfile.cpp
@@ -592,6 +592,7 @@ bool DDSFileClass::Load()
 
     if (data == 0) {
         captainslog_warn("DDSFile '%s' appears to contain no data, size is %d.", m_name, fp->Size());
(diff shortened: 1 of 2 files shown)


src/w3d/renderer/ddsfile.cpp
```diff
@@ -592,6 +592,7 @@ bool DDSFileClass::Load()
 
     if (data == 0) {
         captainslog_warn("DDSFile '%s' appears to contain no data, size is %d.", m_name, fp->Size());
+        fp->Close();
```
tomsons26: auto_file_ptr should indirectly be doing that?

jonwil: No it does not.

tomsons26: right, comment as bugfix

xezon: Perhaps the right fix here is then to auto close on destruction of `auto_file_ptr` instead.

I will take a look if there are more places where `auto_file_ptr` is missing call to Close.

jonwil: Original auto_file_ptr didn't call Close, we should keep it that way.

xezon: I have reverted change to DDSFileClass::Load and will make a new Pull for `auto_file_ptr`

xezon: I have taken another look at `auto_file_ptr` and `GameFileClass` in debugger. It does call `Close` on destruction via virtual destructor:

```c++
GameFileClass::~GameFileClass()
{
    Close();
}
```

So this needs no change.

tomsons26 (approved):

PR #1112, opened 2024-02-11
Fix bug in WorldHeightMap::Parse_Blend_Tile_Data function

xezon:
* Split off of #796

This is addressing supposed original game bug.

tomsons26
> indeed it's a bug in original, not too sure if this should be accepted, don't know the implications of fixing this..

I unclear what this does. `is_blend_edge_tile` member is currently not used in Thyme.

Also a few lines lower is another file read and that also does not set `is_blend_edge_tile`.

Because `is_blend_edge_tile` is a legacy thing that would need to be removed entirely.


jonwil (approved):
Looks good to me, approved.

jonwil (changes requested):
Actually no, I mis-read it. Original does not set is_blend_edge_tile. Original code in WB that outputs the value that we are reading and ignoring outputs hardcoded 0. So the fix there is incorrect.

The change from 0xFFFFFFFF to -1 is fine though.

PR #1116, opened 2024-02-11
Avoid stlport compile error with Visual Studio 2022

xezon:
This change ~~fixes~~ avoids stlport compile error with Visual Studio 2022.

`xstddef` file does not exist in Visual Studio 2022. It does exist in Visual Studio 2019 however. But Visual Studio 2019 also compiles without `xstddef` included in `exception` from stlport.

Unclear what is going on.

```
Visual Studio 2019: _STLP_MSVC = 1929
Visual Studio 2022: _STLP_MSVC = 1938
```

This include was already touched prior with 9d9908f3a28a9b46a25c216b41fc43bb383b740b

### DIFF
--- a/deps/stlport/exception
+++ b/deps/stlport/exception
@@ -57,9 +57,7 @@
 # if defined (_STLP_NO_NEW_NEW_HEADER)
 #  include _STLP_NATIVE_CPP_RUNTIME_HEADER(exception.h)
 # else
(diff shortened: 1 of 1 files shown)


OmniBlade (approved):
If its causing issues lets merge this and worry about reverting it if its needed for something down the road.

xezon:
Yes if some platform needs this explicitly then we compile it in for just that.

PR #1117, opened 2024-02-13
Fix vs2022 clang format

xezon:
VS 2022 pops up an error dialog on every format about this duplicate record.


OmniBlade (approved):
Suprised nothing complained about this before VS 2022 did.

PR #1119, opened 2024-02-15
Revert "Remove always.h from hot include files to reduce compile time"

xezon:
This reverts commit 30cc43174d879316e312a6905720fb87ba801ce1.

It broke string class because of missing PLATFORM_WINDOWS define.


jonwil (approved):
Looks good to me, approved.

PR #1121, opened 2024-02-17
Move always.h from headers to cpp files

xezon:
This change aims to further reduce compile times.

On my machine with SSD, this reduces compile time in Debug configuration by 1 second - from 42 seconds to 41 seconds.

## Disadvantage of this change

Most headers will no longer compile on their own when included in a cpp file. Every cpp file now needs to have always.h included at the very top. Our clang-format already sorts it that way.

## Advantage of this change

This principle causes less file includes and a bit faster compilation.

Also, it allows the project to use precompiled headers moving forward. For next steps, we can move always.h into a stdafx.h, and then bundle it with some other hot files such as xfer.h and asciistring.h, and include stdafx.h instead of always.h in all cpp files.

https://en.wikipedia.org/wiki/Precompiled_header

precomp


OmniBlade (commented):
Okay, so my concerns are 1. I think headers compiling on their own is a useful feature and they should include anything they require. 2. The precompiled header benefits only apply to Windows while we aspire to have the code base compile cross platform. In a similar project https://github.com/OmniBlade/LasMarionetas/tree/master/src/compat I've split what always.h encapsulates into multiple headers and just avoided the macros for platforms in favour of the compiler provided ones.

xezon:
I don't feel strongly about this change. We can close it.

As for multi platform, Windows by far is the most important platform because this is what 99% of the actual players use, so any development improvements for Windows platform are worthwhile improvements.

PR #1140, opened 2024-08-11
Fix reference counting in HLodClass::Remove_Sub_Object

xezon:
Original call to single `Release_Ref` looks wrong to me, because `m_lod[i][j].m_model` and `m_additionalModels[i].m_model` are individually ref counted. So this looks like it would leak.


jonwil:
I verified this against original and the code we have right now is correct to the original (i.e. the subobj->Release_Ref(); line)

xezon:
Ok. Then perhaps this is original game bug. It would only lead to memory leak, not corruption, so it is not critical problem.

PR #1141, opened 2024-08-11
Fix operator precedence in DDSFileClass::Merge_Color

xezon:
This gets rid of msvc warning regarding operator precedence.


jonwil (approved):
Looks OK to me, approved.

PR #1142, opened 2024-08-13
Fix w3dview camera movement

xezon:
Left and right camera movement was previously inverted. This fixes it.


src/tools/w3dviewer/graphicview.cpp
```diff
@@ -236,7 +236,7 @@ void CGraphicView::OnMouseMove(UINT nFlags, CPoint point)
 
         v.X = (((float)point.x - (float)(r.right >> 1)) / (float)(r.right >> 1)
                   - ((float)m_mousePos.x - (float)(r.right >> 1)) / (float)(r.right >> 1))
-            * m_radius * -1.0f;
+            * m_radius;
```
jonwil: The * -1.0f does exist in the original binary.

src/tools/w3dviewer/graphicview.cpp
```diff
@@ -236,7 +236,7 @@ void CGraphicView::OnMouseMove(UINT nFlags, CPoint point)
 
         v.X = (((float)point.x - (float)(r.right >> 1)) / (float)(r.right >> 1)
                   - ((float)m_mousePos.x - (float)(r.right >> 1)) / (float)(r.right >> 1))
-            * m_radius * -1.0f;
+            * m_radius;
 
         v.Y = (((float)point.y - (float)(r.bottom >> 1)) / (float)(r.bottom >> 1)
                   - ((float)m_mousePos.y - (float)(r.bottom >> 1)) / (float)(r.bottom >> 1))
```
jonwil: On this line, the expressions are ((float)(r.bottom >> 1) - (float)point.y) and ((float)(r.bottom >> 1) - (float)m_mousePos.y)

src/tools/w3dviewer/graphicview.cpp
xezon: I can absolutely confirm that this `* -1.0f` causes inverted movement (camera moves left when moving right).

jonwil: I tested the latest w3dview from the Thyme trunk alongside the original Westwood w3dview and when I hold down both mouse buttons and move left and right it behaves the same. But if I use the w3dview from this PR, it behaves differently.
That means the left and right movement currently in Thyme is correct to the original as far as I can tell.

src/tools/w3dviewer/graphicview.cpp
jonwil: Up and down camera movement is incorrect in Thyme w3dview compared to stock w3dview, the above-mentioned change should fix that.

src/tools/w3dviewer/graphicview.cpp
xezon: Ok. So the intent was to have inverted movement on both axis. Ok then I will test your proposal.

src/tools/w3dviewer/graphicview.cpp
xezon: Fixed.

src/tools/w3dviewer/graphicview.cpp
xezon: Tested and worked. Both axis are inverted now.

jonwil (approved):
Looks good to me, approved.

PR #1150, opened 2024-09-06
Fix memset size in IMEManager::Update_Candidate_List

xezon:
Looks like memset had wrong size.


tomsons26:
read https://learn.microsoft.com/en-us/windows/win32/api/imm/nf-imm-immgetcandidatelistcounta

xezon:
@tomsons26 
I still think it is wrong in Thyme. We memset too little of the allocated array. It should be sizeof struct * count, no?

Edit:  Ah no.

It should be `CANDIDATELIST *list = new CANDIDATELIST[count];` instead of `CANDIDATELIST *list = new CANDIDATELIST[size];`

xezon:
Let's try this: #1154

PR #1161, opened 2024-09-19
Fix game pools with __SANITIZE_ADDRESS__

xezon:
This change fixes game pools with `__SANITIZE_ADDRESS__`.

Before, textures would just render black in w3dview tool. Unclear why exactly, but this change reinstates game pools and replaces the pool allocator with `malloc` and `free`.

Textures render normally now.


src/game/common/system/mempool.cpp
```diff
@@ -139,7 +143,9 @@ void MemoryPool::Free_Block(void *block)
     if (block == nullptr) {
         return;
     }
-
+#ifdef __SANITIZE_ADDRESS__
+    free(block);
```
feliwir, line 147: I think we should return afterwards? otherwise it will try deleting it otherwise / applying "delete"

xezon: Where is that delete? I do not see that.

src/game/common/system/gamememory.cpp
```diff
@@ -94,8 +94,6 @@ void Init_Memory_Manager()
         captainslog_dbgassert(g_thePreMainInitFlag, "memory manager is already inited");
     }
 
-#if defined GAME_DEBUG && !defined __SANITIZE_ADDRESS__
```
xezon, line 97: This looks wrong.

xezon: Fixed.

feliwir (approved):

src/game/common/system/gamememory.cpp
Suardi1985: Wtf

Suardi1985 (approved):

Suardi1985 (approved):

Suardi1985 (approved):

Suardi1985 (approved):

Issue #911, opened 2023-05-13
China Horde bonus does not show


xezon:
China Horde bonus does not show.

## Original

5 China Rocket men have red stars under them.

![shot_20230513_135032_1](https://github.com/TheAssemblyArmada/Thyme/assets/4720891/699b9a74-6e63-4f50-8fb7-496aee00f121)

## Thyme

5 China Rocket men have no red stars under them.

![shot_20230513_134951_3](https://github.com/TheAssemblyArmada/Thyme/assets/4720891/547bd1c7-2dfc-4e36-b229-5275e6b8d9bf)

PR #1163, opened 2024-10-10
Boolean Fixes

### LINKED ISSUE #884: Cash Bounty reward value is different
Cash Bounty reward value is different than Original. It looks like it is always 1 cash higher than it should be.

Tested on map
https://github.com/TheSuperHackers/GeneralsGamePatch/tree/main/Maps/Any/!TestRank

## Original

shot_20230307_175918_1

## Thyme

shot_20230307_175711_2

### LINKED ISSUE #1081: Fake GLA Buildings do not draw yellow line around after construction completion
Fake GLA Buildings do not draw yellow line around after construction completion.

First noticed in 9cec51148993006659ada48294dbf5da49fe7b65 Implement some stuff for GameClient (#1072)

## Original

shot_20240204_202637_1

## Thyme

shot_20240204_202526_1

6emmes:
Fixes #884
More faithful to Ghidra code would be addition of GameMath::Round() function, but since float errors are << 0.5 there is no difference compared to Floor.

Fixes #1081 and similar problem with America chemsuits. Suprisingly #911 remains broken

### DIFF
--- a/src/game/common/rts/player.cpp
+++ b/src/game/common/rts/player.cpp
@@ -2998,7 +2998,7 @@ void Player::Do_Bounty_For_Kill(const Object *killer, const Object *killed)
 {
     if (killer != nullptr && killed != nullptr && !killed->Get_Status(OBJECT_STATUS_UNDER_CONSTRUCTION)) {
         unsigned int amount =
-            GameMath::Fast_To_Int_Round(killed->Get_Template()->Calc_Cost_To_Build(this) * m_bountyCostToBuild);
+            GameMath::Fast_To_Int_Floor(killed->Get_Template()->Calc_Cost_To_Build(this) * m_bountyCostToBuild);
 
         if (amount != 0) {
             Get_Money()->Deposit(amount, true);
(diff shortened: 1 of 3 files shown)


6emmes:
Fixes #1081 and similar problem with America chemsuits. Suprisingly #911 remains broken

src/game/common/rts/player.cpp
```diff
@@ -2998,7 +2998,7 @@ void Player::Do_Bounty_For_Kill(const Object *killer, const Object *killed)
 {
     if (killer != nullptr && killed != nullptr && !killed->Get_Status(OBJECT_STATUS_UNDER_CONSTRUCTION)) {
         unsigned int amount =
-            GameMath::Fast_To_Int_Ceil(killed->Get_Template()->Calc_Cost_To_Build(this) * m_bountyCostToBuild);
+            GameMath::Fast_To_Int_Floor(killed->Get_Template()->Calc_Cost_To_Build(this) * m_bountyCostToBuild);
```
tomsons26: Fast_To_Int_Ceil was correct

6emmes: Have I missed a flag? FISTP should round to the nearest.

tomsons26: it rounds to currently set rounding mode actually
Issue with this math will be somewhere else as that's def ceil called there in the binaries

PR #1164, opened 2024-10-23
Removed incorrect assert

6emmes:
In the original code this branch leads to assert("WARNING: Script doCommandButton for button %s not implemented. Doing nothing"). This If exists to check if an AIteam member can doCommandButton (for example barracks building China radar), not Whether given command itself is valid. After change doCommandButton for button %s becomes implemented.
Fixes AI using America ControlRods, China Mines, China Radar and GLA CamoNetting (likely more) in skirmish.


tomsons26 (changes requested):
The gotos in this function need further looking into but out of scope of this pr i guess, looks like this should be:
```
if (upgrade == nullptr) {
    break;
}

if (upgrade->Get_Type() == UPGRADE_TYPE_OBJECT
    && (Has_Upgrade(upgrade) || !Affected_By_Upgrade(upgrade))) {
    break;
}
```

tomsons26 (approved):

PR #1165, opened 2024-10-27
Fix map waypoint loading

### LINKED ISSUE #1053: Maps with custom scripts show no starting positions and player spawns as observer
Maps with rising water feature show no starting positions. Is Thyme specific bug.

Map:
[Dammed Cottages ZH v1.zip](https://github.com/TheAssemblyArmada/Thyme/files/13955933/Dammed.Cottages.ZH.v1.zip)

shot_20240116_222224_1

6emmes:
Fixes #1053


tomsons26 (changes requested):
either g_waypoints or s_waypoints doesn't seem should exist, looks like only g_waypoints should exist

6emmes:
@tomsons26 
Hi, thanks for comment.
Yes, those two variable are very confusing but it seems accurate to the source code (for example WaypointMap::Update).

tomsons26:
hu my comment edit didn't save..

remove s_waypoints and make
```
#ifndef GAME_DLL
WaypointMap *g_waypoints = nullptr;
MapCache *g_theMapCache = nullptr;
#endif
```
code should be only using g_waypoints, there is no duplicate WaypointMap * global

tomsons26 (approved):

